PR #21, opened 2012-07-17
JSON validation test and formatting corrections

bkng:
Addresses PLFM-1344: Typos in JSON schemas -- trailing commas


lib/lib-auto-generated/pom.xml
```diff
@@ -62,11 +62,22 @@
...
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-io</artifactId>
+		</dependency>
 
 		<dependency>
 			<groupId>junit</groupId>
 			<artifactId>junit</artifactId>
 			<scope>test</scope>
 		</dependency>
+
```
john-hill: Can you add a comment about why these are here.

lib/lib-auto-generated/pom.xml
```diff
@@ -62,11 +62,22 @@
 			<artifactId>commons-codec</artifactId>
 			<scope>provided</scope>
 		</dependency>
+
```
john-hill: Comment as well.

lib/lib-auto-generated/src/test/java/org/sagebionetworks/JSONParseTest.java
```diff
@@ -0,0 +1,58 @@
...
+		Iterator<?> itr =  FileUtils.iterateFiles(dir, new String[]{"json"}, true);
+
+		while (itr.hasNext()) {
+		    File file = (File) itr.next();		    
+		    try {
+		    	String jstr = readFileToString(file.getAbsolutePath());
+				assertTrue(JSONValidator.isValidJSON(jstr));
+		    } catch (JsonParseException jpe) {
+		    	fail("Parse failed on file '" + file.getName() + "'\n" + jpe.getMessage());
+			} catch (Exception ioe) {
```
john-hill: remove this catch, and let the test throw any exception needed.

PR #26, opened 2012-07-24
PLFM-1319: close security hole in registration process

jay-hodgson:
Issue a special registration token when creating a new user (sends in an email).  The registration token contains the encrypted session token. Caller can use the registration token to change the user's password, but that's all that can be done. After that, initiate a new session to get a real session token (can't avoid the ToU requirement)

### DIFF
--- a/lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/UserProfile.json
+++ b/lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/UserProfile.json
@@ -54,6 +54,11 @@
 				}
 			]
 		},
+		"email": {
+			"type": "string",
+			"title": "Email",
+			"description": "User's current email address"
+		},
 		"userName": {
 			"type": "string",
 			"title": "User Name",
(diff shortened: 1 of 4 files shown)


lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/UserProfile.json
```diff
@@ -54,6 +54,11 @@
 				}
 			]
 		},
```
jay-hodgson, line 56: added email to lay groundwork for PLFM-1060 and PLFM-267 (mirror crowd fields in user profile, and eventually give the user the ability to change their email address)

services/authutil/src/main/java/org/sagebionetworks/authutil/RegistrationInfo.java
```diff
@@ -0,0 +1,69 @@
+package org.sagebionetworks.authutil;
```
jay-hodgson, line 1: Remove, change to JSON

services/auth/src/main/java/org/sagebionetworks/auth/AuthenticationController.java
```diff
@@ -439,6 +474,25 @@ public void setPassword(@RequestBody User user,
...
+		
+		String sessionToken  = decryptedString(registrationToken.substring(REGISTRATION_TOKEN_PREFIX.length()));
+		String realUserId = CrowdAuthUtil.revalidate(sessionToken);
+		if (realUserId==null) 
+			throw new AuthenticationException(HttpStatus.BAD_REQUEST.value(), "Not authorized.", null);
+		User user = new User();
+		user.setEmail(realUserId);
+		user.setPassword(registrationInfo.getPassword());
+		CrowdAuthUtil.updatePassword(user);
+	}
```
jay-hodgson, line 494: log out session in crowd

PR #30, opened 2012-07-26
SWC-137: when requesting the anonymous user's profile, return a relatively empty profile (but with valid information for the anonymous user)

jay-hodgson:
also contains a change from the code review for PLFM-1319


services/repository/src/main/java/org/sagebionetworks/repo/manager/UserProfileManagerImpl.java
```diff
@@ -56,6 +57,17 @@ public UserProfileManagerImpl(UserProfileDAO userProfileDAO,
 	@Override
 	public UserProfile getUserProfile(UserInfo userInfo, String ownerId)
 			throws NotFoundException, DatastoreException, UnauthorizedException {
+		//if the user is set, and it's the anonymous user, then return an anonymous profile
+		if (userInfo != null && userInfo.getUser() != null && userInfo.getUser().getUserId() != null && 
+				AuthorizationConstants.ANONYMOUS_USER_ID.equals(userInfo.getUser().getUserId())){
+			//for anonymous, just return an empty profile that points to the correct principle id
+			UserProfile anonUserProfile = new UserProfile();
+			anonUserProfile.setOwnerId(userInfo.getIndividualGroup().getId());
+			anonUserProfile.setUserName(AuthorizationConstants.ANONYMOUS_USER_ID);
```
john-hill: Display should be "Anonymous"

PR #32, opened 2012-07-27
fix for PLFM-1341

john-hill:


services/repository/src/main/java/org/sagebionetworks/repo/web/DeadlockWatcher.java
```diff
@@ -0,0 +1,65 @@
...
+					// We might get another shot.
+				}
+				log.debug(String.format(FAILED_ATTEMPT, attempt, wait));
+				// make an exponential wait
+				wait *= wait;
+				attempt++;
+			}
+			// throw the error
+			log.debug(String.format(EXASTED_ALL_ATTEMPTS, attempt-1));
+			throw e;
```
dburdick, line 61: Maybe throw service unavailable instead of deadlock?

PR #38, opened 2012-08-03
PLFM-1348: Synapse Java Client login method needs to deal with Terms of Use

jay-hodgson:


client/synapseJavaClient/src/main/java/org/sagebionetworks/client/Synapse.java
```diff
@@ -302,18 +307,103 @@ public void login(String username, String password, boolean explicitlyAcceptsTer
...
 
-			JSONObject credentials = createAuthEntity("/session", loginRequest);
-
-			defaultGETDELETEHeaders.put(SESSION_TOKEN_HEADER, credentials
-					.getString(SESSION_TOKEN_HEADER));
-			defaultPOSTPUTHeaders.put(SESSION_TOKEN_HEADER, credentials
-					.getString(SESSION_TOKEN_HEADER));
-			requestProfile = reqPr;
+			try {
+				credentials = createAuthEntity("/session", loginRequest);
```
john-hill, line 311: The credentials returned from the auth service should include the first and last name and display name (can file a bug).

PR #39, opened 2012-08-03
PLFM-1326: Finished Controller -> Service refactor and added ServiceProvider. Updated EntityBundle and added integration test.

bkng:


services/repository/src/main/java/org/sagebionetworks/repo/web/service/ServiceProvider.java
```diff
@@ -0,0 +1,34 @@
...
+	public EntityBundleService entityBundleService;
+	@Autowired
+	public S3TokenService s3TokenService;
+	@Autowired
+	public SearchService searchService;
+	@Autowired
+	public UserGroupService userGroupService;
+	@Autowired
+	public UserProfileService userProfileService;
+
```
john-hill: add getter methods and make the members private

PR #41, opened 2012-08-06
PLFM-1348 and PLFM-1416

jay-hodgson:


client/synapseJavaClient/src/main/java/org/sagebionetworks/client/Synapse.java
```diff
@@ -302,18 +305,57 @@ public void login(String username, String password, boolean explicitlyAcceptsTer
...
+	public boolean revalidateSession() throws SynapseException {
+		JSONObject sessionInfo = new JSONObject();
+		try {
+			sessionInfo.put("sessionToken", getCurrentSessionToken());
+			try {
+				putAuthEntity("/session", sessionInfo);
+			} catch (SynapseForbiddenException e) {
+				//403 error
+				throw new SynapseTermsOfUseException(ServiceConstants.TERMS_OF_USE_ERROR_MESSAGE);
+			}
```
brucehoff, line 351: Can you get the 'terms of use error message' from e.getMessage(), rather than hard coding it as the message field in the newly formed exception?

lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/ChangeUserPassword.json
```diff
@@ -0,0 +1,14 @@
+{
+	"title": "Change User Password",
+	"description": "JSON schema for ChangeUserPassword POJO",
+	"properties": {
+		"sessionToken": {
+			"type": "string",
+			"description": "Session token"
+		},
```
brucehoff: Do you really need to put the session token in the request body?  It's already in the header.

services/auth/src/main/java/org/sagebionetworks/auth/AuthenticationController.java
```diff
@@ -459,18 +444,30 @@ public void sendChangePasswordEmail(@RequestBody User user) throws Exception {
 	@ResponseStatus(HttpStatus.NO_CONTENT)
 	@RequestMapping(value = "/apiPasswordEmail", method = RequestMethod.POST)
 	public void sendSetAPIPasswordEmail(@RequestBody User user) throws Exception {
-		sendUserPasswordEmail(user.getEmail(), PW_MODE.SET_API_PW);
+		String userId = null;
+		if (user.getEmail() != null && user.getEmail().length() > 0)
+			userId = user.getEmail();
+		else if (user.getSessionToken() != null && user.getSessionToken().length() > 0) {
+			userId = CrowdAuthUtil.revalidate(user.getSessionToken());
```
brucehoff: I don't think you need to call 'revalidate'.  If there's a session token in the original request, the authentication filter has already validated it and added the user name/id as a request parameter.

services/authutil/src/main/java/org/sagebionetworks/authutil/User.java
```diff
@@ -10,6 +10,7 @@
 	private String lastName;
 	private String displayName;
 	private Boolean acceptsTermsOfUse;
+	private String sessionToken;
```
brucehoff, line 14: remove sessionToken field

PR #56, opened 2012-08-10
Logging Library Addition

RadicalZephyr:
Refactor of ActivityLogger as well.

### DIFF
--- a/lib/logging/src/test/java/org/sagebionetworks/logging/LoggingUtilsTestHelper.java
+++ b/lib/logging/src/test/java/org/sagebionetworks/logging/LoggingUtilsTestHelper.java
@@ -1,0 +1,22 @@
+package org.sagebionetworks.logging;
+
+import org.junit.Test;
+
+public class LoggingUtilsTestHelper {
+
+	/**
+	 * Fake test to keep the test suite from failing because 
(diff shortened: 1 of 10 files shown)


lib/logging/src/test/java/org/sagebionetworks/logging/LoggingUtilsTestHelper.java
```diff
@@ -0,0 +1,22 @@
+package org.sagebionetworks.logging;
+
```
jay-hodgson, line 2: is this still necessary?

lib/logging/src/test/java/org/sagebionetworks/logging/SynapseLoggingUtilsTest.java
```diff
@@ -0,0 +1,179 @@
...
+
+import org.aspectj.lang.reflect.MethodSignature;
+import org.joda.time.DateTime;
+import org.joda.time.format.DateTimeFormatter;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+
+public class SynapseLoggingUtilsTest {
+	private static Method SIMPLE_METHOD = null;
```
jay-hodgson, line 23: constants?

PR #63, opened 2012-08-14
PLFM-1405 Etags need to be unique across stacks

eric-wu:
Basically rewriting the e-tag generation code to use collision-resistant, random numbers as e-tags.  More specifically I am creating an new interface ETagGenerator that takes a TaggableEntity as input and returns a String as the e-tag.  I am adding one implementation of ETagGenerator based on UUID.

I am changing the e-tag data type from Long to String as, to be collision-resistant, the tag needs to be at least 128-bit.  Both Java Long and MySql BigInt do not support that width directly.

Note that UUID only supports uniqueness.  It is not deterministic and thus cannot validate content.  Deterministic hashing on content can be used to achieve both.  However I am not adopting hashing yet as quite amount of design and coding will be needed.  I will defer in the future what fields to compute hash from (e.g. name + id + timestamp).

Besides being almost unique, UUID can be computed locally (in a distributed manner) and fast.  UUID initializes a static SecureRandom instance and the initialization takes 50 - 100 ms.  Once initialized, computing is fast at around 0.05 ms.

The only thing I am not sure about is the URL encoding/decoding for the e-tags in the dto's.  I don't think it is needed.  But to be safe, I am leaving them there.


eric-wu:
Bruce,

The memory issue with 'mvn install' is likely not from my change. I tested that the current upstream develop had the same memory constraint as my develop.  Both needed 2 GB to finish and both crashed with 1 GB Xmx.

The null check in NodeDaoImpl was removed by accident.  I have added it back (see my 2nd commit).  I also browsed through the files that there are no similar mistakes.

If you don't have other concerns, this is good to merge.

Thanks!

PR #66, opened 2012-08-15
PLFM-1417: Added service to fetch user/group headers, by name prefix.

bkng:


services/repository/src/test/java/org/sagebionetworks/repo/web/controller/UserProfileControllerTest.java
```diff
@@ -0,0 +1,77 @@
```
dburdick, line 0: Move this logic to a unit test on UserProfileService and mock the userManager to have test-defined users searched for.

services/repository/src/test/java/org/sagebionetworks/repo/web/util/PatriciaTrieTest.java
```diff
@@ -0,0 +1,318 @@
```
dburdick, line 0: You can just rely on their package to work properly as you're not wrapping it in an interface with potentially other impls

dburdick:
modifications to be done on develop

PR #69, opened 2012-08-16
Change the SynapseJavaClient to not output anything to the console

RadicalZephyr:
Instead, it uses it's logger, as it should.


jay-hodgson:
after a few hours of consideration, I've decided to approve this pull request.

PR #101, opened 2012-09-08
code review changes.  optionally set md5 for external links

jay-hodgson:
also added a set of additional tests around external locationables


john-hill:
Looks good!

PR #125, opened 2012-09-21
Add controller/manager/servie for PLFM-1499 (Allow users to query storage usage)

eric-wu:


services/repository/src/main/java/org/sagebionetworks/repo/web/service/StorageUsageServiceImpl.java
```diff
@@ -0,0 +1,76 @@
...
+
+	@Autowired
+	private StorageUsageManager storageUsageManager;
+
+	@Override
+	public StorageUsageSummaryList getStorageUsage(String currUserId, String userId, List<StorageUsageDimension> dList)
+			throws UnauthorizedException, DatastoreException {
+
+		// this will throw UnauthorizedException
+		validateUser(currUserId, userId);
```
brucehoff: name could be more descriptive like "checkUserAuthorization", or checkUserisSelfOrAdministrator

jay-hodgson:
Add a Synapse Java Client call?

PR #130, opened 2012-09-24
merged part of PLFM-1362

john-hill:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/DBOChangeDAO.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+public interface DBOChangeDAO {
+	
+	/**
+	 * If the objectId already exists, then replace it, else add a new row.
+	 * @param change
+	 * @return
+	 */
+	public DBOChange replaceChange(DBOChange change);
+
```
brucehoff, line 21: To be consistent with other DAOs, this interface should not know about DBO objects.

PR #154, opened 2012-10-04
PLFM-1527: create local ACLs by inheritance; PLFM-1438: batch UserGroupHeader REST service

bkng:


bkng:
These changes will support the rewrite of the portal ACLEditor.

john-hill:
Bennett,

If we add this method then the pressing the "Create local Sharing" button will not respect an "Apply", "Cancel" model and it must.

If the user presses the "Create Local Sharing" button, and then hits cancel or closes the dialog the Entity should remain unchanged.  The ACL editor should not attempt to Undo changes already applied.

With an apply/cancel model you need to create an ACL that has everything you want but has not yet been applied to the Entity.  We then show that ACL to the user and when they like how it looks, they can press apply.

Please call me if you have any questions.

bkng:
Hi John -

Ah, yes, I agree. The PLFM-1438 method was intended to be a safer repo-side
alternative to the existing createACL() method, insofar as creation of
local ACLs should reference benefactor ACLs as a general rule.

You bring up a good scenario that I hadn't yet considered. The editor does
download a copy of the benefactor ACL, so the "Create Local Sharing" button
could simply unlock the local copy of that ACL for portal-side editing,
without actually utilizing the new createACL() method.

That said, I placed the PLFM-1438 logic in the platform such that any
client could use that functionality. Please let me know if I should modify
it.

Thanks,
Bennett

PR #164, opened 2012-10-09
PLFM-1409: Fix getAllVersions service

RadicalZephyr:
Reorganized the commits a bit and cleaned up the history.


john-hill:
Looks good!

PR #172, opened 2012-10-15
added support for query filter on 'benefactorId' (PLFM-1542)

john-hill:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/query/jdo/SqlConstants.java
```diff
@@ -242,7 +242,7 @@
 	
 	static {
 		Field[] fields = Node.class.getDeclaredFields();
-		PRIMARY_FIELDS = new String[fields.length];
+		PRIMARY_FIELDS = new String[fields.length+1];
```
eric-wu, line 245: Revert the change

PR #180, opened 2012-10-19
Set up worker that updates from the dynamo queue and add parent id to ChangeMessage

eric-wu:


lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/message/TransactionalMessengerImplTest.java
```diff
@@ -131,6 +133,7 @@ public void testSendMessageSameIdDifferntType() throws JSONObjectAdapterExceptio
 		first.setTimestamp(new Date(System.currentTimeMillis()/1000*1000));
 		first.setObjectEtag("etag");
 		first.setObjectId("syn456");
+		// first.setParentId("syn789"); // parent id can be null
```
eric-wu, line 136: explicitly set to null

services/dynamo/pom.xml
```diff
@@ -24,7 +68,11 @@
 				<configuration>
 					<webResources>
 						<resource>
-							<directory>src/main/resources/webapp</directory>
+							<directory>src/main/webapp</directory>
+							<filtering>true</filtering>
+						</resource>
+						<resource>
+							<directory>src/main/resources</directory>
```
eric-wu, line 75: This may not be needed.

PR #189, opened 2012-10-26
Add DynamoDB initialization

eric-wu:


lib/lib-dynamo/src/main/java/org/sagebionetworks/dynamo/DynamoSetupImpl.java
```diff
@@ -0,0 +1,266 @@
...
+			return;
+		}
+
+		// Otherwise wait for the tables to be ready
+		long delay = timeoutInMillis / 60 + 1;
+		long startTime = System.currentTimeMillis();
+		long endTime = startTime + timeoutInMillis;
+		while (System.currentTimeMillis() < endTime) {
+
+			try {Thread.sleep(delay);} catch (InterruptedException e) {}
```
john-hill, line 104: throw a runtime

PR #208, opened 2012-11-14
Add code measurements and reports

eric-wu:


pom.xml
```diff
@@ -695,6 +721,41 @@
...
+			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-pmd-plugin</artifactId>
+				<version>2.7.1</version>
+				<configuration>
+					<linkXref>true</linkXref>
+					<targetJdk>1.6</targetJdk>
+					<excludes>
+						<exclude>**/generated/*.java</exclude>
```
eric-wu: This does not actually exclude the folder 'auto-generated-pojos'

pom.xml
```diff
@@ -695,6 +721,41 @@
...
+				<version>2.3</version>
+			</plugin>
+			<plugin>
+				<groupId>org.codehaus.mojo</groupId>
+				<artifactId>findbugs-maven-plugin</artifactId>
+				<version>${findbugs-maven-plugin.version}</version>
+			</plugin>
+			<plugin>
+				<groupId>com.atlassian.maven.plugins</groupId>
+				<artifactId>maven-clover2-plugin</artifactId>
```
eric-wu, line 756: Find out if we need to run the tests to get the coverage.

PR #212, opened 2012-11-15
PLFM-1550: fix permissions for creating/updating child entities

bkng:


john-hill:
Looks good.  Looking at NodeManagerImpl.java line 312, it seems we are missing a security check.  The user can only move to the new parent if they have create permission on the parent entity.

bkng:
Great, I'll look into that. Also, is there a good way to test moving an entity to be a child of another user's entity? Due to the current browser filtering in the Portal, it's impossible to select an entity which you do not own.

john-hill:
I filed an issue see: https://sagebionetworks.jira.com/browse/PLFM-1596.

There are some parentId change tests in NodeManagerImplAutoWiredTest.
Another option is to test it with a mocked unit NodeManagerImpleUnitTest.
I would prefer the unit test approach.

John

On Thu, Nov 15, 2012 at 3:28 PM, Bennett Ng notifications@github.comwrote:

> Great, I'll look into that. Also, is there a good way to test moving an
> entity to be a child of another user's entity? Due to the current browser
> filtering in the Portal, it's impossible to select an entity which you do
> not own.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/Sage-Bionetworks/Synapse-Repository-Services/pull/212#issuecomment-10430499.

## 

John Hill
Principal Software Engineer
Sage Bionetworks
1100 Fairview Ave N, M1-C117
Seattle, WA  98109

john.hill@sagebase.org
office: 206.667.1160
mobile: 206.335.1798
http://www.sagebase.org

PR #217, opened 2012-11-21
Implement DynamoDB DAO for hierarchical data

eric-wu:


lib/lib-dynamo/src/main/java/org/sagebionetworks/dynamo/dao/NodeTreeDaoNodeLineageImpl.java
```diff
@@ -0,0 +1,747 @@
...
+		if (parentLineage == null) {
+			// If not, we should create
+			return this.create(child, parent, timestamp);
+		}
+
+		// If the child is already pointing to the same parent, no update needed
+		String parentId = parentLineage.getAncestorOrDescendantId();
+		if (parent.equals(parentId)) {
+			// The same (child, parent) already exists
+			// We will skip the create and keep the old timestamp
```
eric-wu, line 167: Change to update

PR #226, opened 2012-11-27
(Re?)Implementation of Promote Version to top service

RadicalZephyr:


services/repository/src/main/java/org/sagebionetworks/repo/web/controller/EntityController.java
```diff
@@ -323,6 +323,32 @@ public void deleteEntityVersion(
...
+	 * @param id
+	 * @param userId
+	 * @param versionNumber
+	 * @param request
+	 * @throws NotFoundException
+	 * @throws DatastoreException
+	 * @throws UnauthorizedException
+	 * @throws ConflictingUpdateException
+	 */
+	@ResponseStatus(HttpStatus.NO_CONTENT)
```
john-hill: This is annotation is why you are not seeing any results returned.  You need it to be HttpStatus.CREATED.

RadicalZephyr: ... That would do it.  Let me fix that real quick.

integration-test/src/test/java/org/sagebionetworks/IT500SynapseJavaClient.java
```diff
@@ -1237,10 +1244,56 @@ public void testGetAllUserAndGroupIds() throws SynapseException{
...
+		toDelete.add(entity.getId());
+		return entity;
+	}
+
+	private <T extends Locationable> void versionEntity(T entity, int numVersions) throws SynapseException {
+		try {
+			for (int i = 0; i < numVersions; i++) {
+				entity.setMd5(md5Sum(entity.getMd5()));
+				entity = synapse.putEntity(entity);
+			}
```
john-hill, line 1286: why are you using the MD5 to create new versions?  Didn't we expose creating new versions explicitly like our users asked us to do?

PR #230, opened 2012-11-30
Addressed PLFM-1608

john-hill:


services/repository/src/main/java/profiler/org/sagebionetworks/cloudwatch/Consumer.java
```diff
@@ -23,16 +23,14 @@
...
-	// synchronized list to hold MetricDatums objects
-	private List<ProfileData> listProfileData = Collections
-			.synchronizedList(new ArrayList<ProfileData>());
-
-	// list of watchers who will be notified with success/failure of all "puts"
-	List<Watcher> listOfWatchers = Collections
-			.synchronizedList(new ArrayList<Watcher>());
+	// We us an atomic reference to the list instead of using synchronization.
+	private AtomicReference<List<ProfileData>> listProfileData = 
+			new AtomicReference<List<ProfileData>>(Collections.synchronizedList(new LinkedList<ProfileData>()));
```
john-hill, line 33: this should not be synchronized

PR #232, opened 2012-12-04
Patch dynamo on staging

eric-wu:
1) PLFM-1613 - Increase dynamo throughput and descrease worker frequency
2) PLFM-1614 - Add index.html for health check
3) Reduce dynamo tomcat logging

### DIFF
--- a/lib/lib-dynamo/src/main/java/org/sagebionetworks/dynamo/config/DynamoConfig.java
+++ b/lib/lib-dynamo/src/main/java/org/sagebionetworks/dynamo/config/DynamoConfig.java
@@ -4,6 +4,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
+import org.sagebionetworks.StackConfiguration;
 import org.sagebionetworks.dynamo.config.DynamoTableConfig.DynamoKey;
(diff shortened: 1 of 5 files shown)


john-hill:
Looks g

john-hill:
oops

lib/lib-dynamo/src/main/java/org/sagebionetworks/dynamo/config/DynamoConfig.java
```diff
@@ -26,6 +27,9 @@
 		DynamoKey rangeKey = new DynamoKey(DboNodeLineage.RANGE_KEY_NAME, ScalarAttributeType.S);
 		DynamoKeySchema keySchema = new DynamoKeySchema(hashKey, rangeKey);
 		DynamoThroughput throughput = new DynamoThroughput(18L, 10L);
+		if (this.isProdStack()) {
+			throughput = new DynamoThroughput(75L, 50L);
```
john-hill, line 31: Are we still planning to do this dynamically?

eric-wu, line 31: Yes, dynamically, I plan to do it in the near future.

PR #240, opened 2012-12-08
Fix the logging of arguments

RadicalZephyr:
Add a special case for dealing with HttpServletRequest and
also throw an exception if there isn't an argument of that
type.


RadicalZephyr:
Lib logging shouldn't be dependent on the HttpServletRequest.  Similiarly, it probably shouldn't be dependent on AspectJ style pointcut information (which it currently is).

RadicalZephyr:
[PLFM-1625](https://sagebionetworks.jira.com/browse/PLFM-1625) captures the dependency change.

PR #244, opened 2012-12-12
Add the two getDescendants() methods to the Java client

eric-wu:


integration-test/src/test/java/org/sagebionetworks/IT520SynapseJavaClientDynamo.java
```diff
@@ -0,0 +1,78 @@
...
+import org.sagebionetworks.client.Synapse;
+import org.sagebionetworks.client.exceptions.SynapseException;
+import org.sagebionetworks.repo.model.Entity;
+import org.sagebionetworks.repo.model.EntityIdList;
+import org.sagebionetworks.repo.model.Project;
+import org.sagebionetworks.repo.model.Study;
+
+// TODO: At this point, we have ~1000 messages to work through
+// We need a better way to go through the queue.
+@Ignore
```
john-hill, line 18: Try to enable if this test comes first

PR #247, opened 2012-12-13
Add LogEvent value class

RadicalZephyr:


lib/logging/src/main/java/org/sagebionetworks/logging/collate/LogEvent.java
```diff
@@ -0,0 +1,79 @@
...
+
+	public String getLine() {
+		return line;
+	}
+
+	@Override
+	public boolean equals(Object o) {
+		if (this == o) return true;
+
+		if (o instanceof LogEvent) {
```
eric-wu, line 42: A nitty-gritty comment.  Line 42-52, you could reduce the complexity of the if-else branches to one level like this:

if ( ! o instanceof LogEvent) {
    return false;
}

if (...) // check fields equals() {
    return true;
}

return false;

Or, you can let Eclipse generate the code for you.

lib/logging/src/main/java/org/sagebionetworks/logging/collate/LogEvent.java
```diff
@@ -0,0 +1,79 @@
...
+	public int hashCode() {
+		int result = 42;
+		result = 31 * result + timestamp.hashCode();
+		result = 31 * result + (line != null ? line.hashCode() : 0);
+		return result;
+	}
+
+	@Override
+	public int compareTo(LogEvent le) {
+		if (this.equals(le))
```
eric-wu, line 65: If I understand it right, the check for equals() does unnecessary extra work when the two are not equal.  It can be removed.

lib/logging/src/main/java/org/sagebionetworks/logging/collate/LogEvent.java
```diff
@@ -0,0 +1,79 @@
...
+			compareTo = line.compareTo(le.line);
+		else if (compareTo == 0 && le.line != null)
+			compareTo = -1;
+
+		return compareTo;
+	}
+
+
+}
\ No newline at end of file
```
eric-wu, line 79: Need a newline here

PR #248, opened 2012-12-13
Fixes for LogEvent and an interface for defining Reader's for specific log formats

RadicalZephyr:
As well as an implementation of a reader for the activity logs.

### DIFF
--- a/lib/logging/src/main/java/org/sagebionetworks/logging/collate/LogEvent.java
+++ b/lib/logging/src/main/java/org/sagebionetworks/logging/collate/LogEvent.java
@@ -6,7 +6,7 @@ public final class LogEvent implements Comparable<LogEvent> {
 
 	private final DateTime timestamp;
 	private final String line;
-	
+
 	public LogEvent(DateTime timestamp) {
 		if (timestamp == null) throw new IllegalArgumentException("Timestamp cannot be null.");
 		this.timestamp = timestamp;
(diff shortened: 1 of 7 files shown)


lib/logging/src/main/java/org/sagebionetworks/logging/collate/LogEvent.java
```diff
@@ -39,17 +39,18 @@ public String getLine() {
 	public boolean equals(Object o) {
 		if (this == o) return true;
```
brucehoff, line 40: equals should call compareTo rather than reimplementing the comparison logic

lib/logging/src/main/java/org/sagebionetworks/logging/reader/ActivityLogReader.java
```diff
@@ -0,0 +1,50 @@
...
+	public static class ActivityLogReaderFactory implements LogReader.LogReaderFactory<ActivityLogReader> {
+		@Override
+		public ActivityLogReader create(BufferedReader rdr) {
+			return new ActivityLogReader(rdr);
+		}
+	}
+
+	private BufferedReader reader;
+
+	public ActivityLogReader(BufferedReader rdr) {
```
brucehoff, line 29: consider making constructor private

lib/logging/src/test/java/org/sagebionetworks/logging/reader/ActivityLogReaderTest.java
```diff
@@ -0,0 +1,37 @@
...
+import org.sagebionetworks.logging.reader.ActivityLogReader.ActivityLogReaderFactory;
+
+public class ActivityLogReaderTest {
+
+	@Test
+	public void readLogEvent() throws IOException {
+		String pathname = "src/test/resources/ActivityLogReaderTestLogFile.log";
+		File logFile = new File(pathname);
+		if (!logFile.exists())
+			fail("Necessary resource: "+pathname+", is mising.");
```
brucehoff, line 21: mising->missing

PR #258, opened 2013-01-02
PLFM-1420 Repo service should not return current eTag for entity version

eric-wu:


brucehoff:
Philosophically, the solution is to make the 'version' type of the repo services to become 'read only' services.  I.e. if you get a specific version of an entity (even if it is the _latest_ version you are requesting), the returned object is read only because it lacks an etag.

lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/dao/NodeDAOImplTest.java
```diff
@@ -859,13 +860,19 @@ public void testNewVersionAnnotations() throws Exception {
 		NamedAnnotations namedCopyV1 = nodeDao.getAnnotationsForVersion(id, 1L);
 		Annotations v1Annos = namedCopyV1.getAdditionalAnnotations();
 		assertNotNull(v1Annos);
```
brucehoff, line 864: need one more test, i.e. to get the NamedAnnotations

brucehoff, line 864: may need test for getNodeForVersion.  Eric will double check.

brucehoff:
good job!!

PR #259, opened 2013-01-08
One more fix on the CHANGES table for PLFM-1627 

eric-wu:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/DBOChangeDAOImpl.java
```diff
@@ -112,4 +137,32 @@ public void deleteAllChanges() {
 		return ChangeMessageUtils.createDTOList(dboList);
 	}
 
+	// Retry with exponentially increasing delays
+	private <T> T retryForDeadlock(final Callable<T> task, int numOfRetries) {
+		// Implement with retries to work around deadlocks
+		long delay = 0L;
+		while (numOfRetries > 0) {
+			try {
+				Future<T> future = retryExecutor.schedule(task, delay, TimeUnit.SECONDS);
```
john-hill: do all tries on the same thread.

PR #264, opened 2013-01-09
Fix for PLFM-1662

john-hill:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/DBOChangeDAOImpl.java
```diff
@@ -81,10 +86,21 @@ public ChangeMessage replaceChange(ChangeMessage change) {
...
-		// First remove the row for this object.
-		simpleJdbcTemplate.update(SQL_DELETE_BY_OBJECT_ID, objectId);
+		if(type == null) throw new IllegalArgumentException("ObjectType cannot be null");
+		// To avoid gap locking we do not delete by the ID.  Rather we get the primary key, then delete by the primary key.
+		List<Long> list = simpleJdbcTemplate.query(SELECT_CHANGE_NUMBER_FOR_OBJECT_ID_AND_TYPE, new RowMapper<Long>(){
+			@Override
+			public Long mapRow(ResultSet rs, int rowNum) throws SQLException {
+				return rs.getLong(COL_CHANGES_CHANGE_NUM);
+			}},objectId, type.name());
+		if(list.size() > 1) throw new IllegalStateException("Found multiple rows with ObjectId: "+objectId+" and ObjectType type "+type);
```
john-hill: don't throw just delete all that are found.

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/message/TransactionalMessengerImpl.java
```diff
@@ -171,14 +155,14 @@ public void afterCommit() {
 		@Override
 		public void beforeCommit(boolean readOnly) {
 			// write the changes to the database
-			Map<String, ChangeMessage> currentMessages = getCurrentBoundMessages();
+			Map<ChangeMessageKey, ChangeMessage> currentMessages = getCurrentBoundMessages();
```
john-hill, line 158: put in a clean map instead of replace.

PR #268, opened 2013-01-12
Add a worker to sync RDS and DynamoDb periodically and other trivial improvements

eric-wu:


lib/lib-dynamo/src/main/java/org/sagebionetworks/dynamo/dao/NodeTreeDaoNodeLineageImpl.java
```diff
@@ -610,9 +610,15 @@ private NodeLineage getParentLineage(String child, DynamoDBMapper mapper) {
...
-			if (dist != (i + 1)) {
-				throw new IncompletePathException("Missing ancestor at depth " + (dboListSize - i - 1)
+			// The list is sorted by distance. The first ancestor at index 0 is the parent
+			// whose distance is 1. If we have a complete path, the distance of any ancestor
+			// in the list should be exactly i+1
+			if (dist > (i + 1)) {
+				throw new IncompletePathException("Missing ancestor at distance " + (i + 1)
 						+ " in the ancestor path for node " + node);
+			} else if (dist < (i + 1)) {
+				throw new MultipleInheritanceException("Node " + node
```
xschildw, line 620: Error msg is wrong

PR #284, opened 2013-01-25
Trashcan supporting changes

eric-wu:
I am stil working on the trash manager.  I'd like to check in some of the supporting changes first to avoid a large change list.

1) Add a field for the original parent
2) Add the TrashedEntity DTO
3) Rename DBOTrash to DBOTrashedEntity
4) Add a method to NodeInheritanceManager to also reset self-benefactors when moving to the trash can
5) Add a manager for node tree query service
6) Fix StackConfiguration rounding errors


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/DBOTrashCanDaoImpl.java
```diff
@@ -108,7 +114,24 @@ public void delete(Long userGroupId, Long nodeId)
...
+
+	private List<TrashedEntity> convertDboToDto(List<DBOTrashedEntity> dboList) {
+		List<TrashedEntity> trashList = new ArrayList<TrashedEntity>(dboList.size());
+		for (DBOTrashedEntity dbo : dboList) {
+			trashList.add(convertDboToDto(dbo));
 		}
+		return trashList;
+	}
+
+	private TrashedEntity convertDboToDto(DBOTrashedEntity dbo) {
```
john-hill, line 129: can be static

lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/TrashedEntity.json
```diff
@@ -0,0 +1,24 @@
...
+			"title": "Deleted By",
+			"description":"The ID of the user who deleted the entity."			
+		},
+		"deletedOn":{
+			"type": "string",
+			"format": "date-time",
+			"title": "Deleted On",
+			"description":"The date and time when the deletion occurred."			
+		},
+		"parentId": {
```
john-hill, line 19: original parent id?

PR #297, opened 2013-02-07
Rewrite the autowired test for trash manager and various changes per feedback

eric-wu:


services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/trash/TrashManagerImplAutowiredTest.java
```diff
@@ -58,5 +87,264 @@ public void after() throws Exception {
...
+		results = trashManager.viewTrash(testUserInfo, 0, 1000);
+		assertEquals(0L, results.getTotalNumberOfResults());
+		assertEquals(0, results.getResults().size());
+
+		try {
+			// This is a rather strange scenario
+			// Where the node is restored to the root (or any node that the
+			// user has the CREATE privilege, the node's benefactor is set
+			// the root. The user does not have READ access on the node however
+			nodeRetrieved = nodeManager.get(testUserInfo, nodeId);
```
john-hill, line 197: Currently create entity handles the case where something is added to root.  That needs to be generalized such that an move would setup the ACL the same way as a create.

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/trash/TrashManagerImplAutowiredTest.java
```diff
@@ -58,5 +87,264 @@ public void after() throws Exception {
...
+			fail();
+		} catch (UnauthorizedException e) {
+			assertTrue(true);
+		}
+	}
+
+	@Test
+	public void testMultipleNodeRoundTrip() throws Exception {
+
```
john-hill, line 206: will add other precondition tests as well.

PR #302, opened 2013-02-12
PLFM-1735: Move IdGenerator logic to Manager level for Competition, Submission

bkng:


bkng:
Will add integration test and move to 'develop' branch

PR #308, opened 2013-02-14
Add trash controller, more tests.

eric-wu:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/NodeInheritanceManagerImpl.java
```diff
@@ -129,7 +147,7 @@ private void changeAllChildrenTo(String currentlyInheritingFromId, String parent
 	private void addChildrenToChange(String currentBenefactorId, String parentId, Set<String> toChange) throws NotFoundException, DatastoreException{
 		// Find find the parent's benefactor
 		String parentCurrentBenefactorId = nodeInheritanceDao.getBenefactor(parentId);
-		if(parentCurrentBenefactorId.equals(currentBenefactorId)){
```
john-hill, line 132: pass along the boolean flag

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/NodeManagerImpl.java
```diff
@@ -364,14 +364,36 @@ private void updateNode(UserInfo userInfo, Node updatedNode, NamedAnnotations up
...
-			nodeDao.changeNodeParent(updatedNode.getId(), updatedNode.getParentId());
-			nodeInheritanceManager.nodeParentChanged(updatedNode.getId(), updatedNode.getParentId(), skipSelfBenefactor);
+			nodeDao.changeNodeParent(nodeInUpdate, parentInUpdate);
+			// Update the ACL accordingly
+			if (skipBenefactor) {
+				nodeInheritanceManager.nodeParentChanged(nodeInUpdate, parentInUpdate);
+			} else {
+				// If the node is being moved to right under the root, we need to create a new ACL
+				// The root cannot be the benefactor
+				boolean newAcl = nodeDao.isNodeRoot(parentInUpdate);
```
john-hill, line 381: we want to create an ACL for any node that moves into the root even if it is not related to the trash can.

PR #316, opened 2013-02-22
Java client and integration tests for trash can

eric-wu:
Added an integration test for migrating the trash can in IT100BackupRestoration.  But I set it disabled for now as the bootstrap trash folder is lost at this point.

### DIFF
--- a/integration-test/src/test/java/org/sagebionetworks/IT070SynapseJavaClientTrashCanTest.java
+++ b/integration-test/src/test/java/org/sagebionetworks/IT070SynapseJavaClientTrashCanTest.java
@@ -1,0 +1,109 @@
+package org.sagebionetworks;
+
+import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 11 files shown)


integration-test/src/test/java/org/sagebionetworks/IT070SynapseJavaClientTrashCanTest.java
```diff
@@ -0,0 +1,109 @@
...
+		String user = StackConfiguration.getIntegrationTestUserOneName();
+		String pw = StackConfiguration.getIntegrationTestUserOnePassword();
+
+		UserSessionData session = synapse.login(user, pw);
+		assertNotNull(session);
+		assertNotNull(session.getProfile().getUserName());
+		assertNotNull(session.getSessionToken());
+
+		parent = new Project();
+		parent.setName("IT530SynapseJavaClientTrashCanTest.parent");
```
john-hill, line 46: rename

PR #333, opened 2013-03-08
ability to create a single filehandle, and explicitly set it's content type

jay-hodgson:


client/synapseJavaClient/src/main/java/org/sagebionetworks/client/Synapse.java
```diff
@@ -1447,6 +1447,48 @@ public FileHandleResults createFileHandles(List<File> files) throws SynapseExcep
...
 	
 	/**
+	 * Upload a file to Synapse
+	 * 
+	 * @param file
+	 * @return
+	 * @throws IOException 
+	 * @throws ClientProtocolException 
+	 */
+	public FileHandle createFileHandle(File file, String contentType) throws SynapseException{
```
john-hill, line 1439: Looks like you did a copy/paste.  Instead can we refactor the file list method to use this method?

PR #347, opened 2013-03-20
Add a retryable HTTP client over EZID's REST APIs

eric-wu:


lib/lib-doi/src/test/java/org/sagebionetworks/doi/EzidClientTest.java
```diff
@@ -0,0 +1,157 @@
...
+		assertEquals("Synapse", httpClient.getParams().getParameter(CoreProtocolPNames.USER_AGENT));
+		AuthScope authScope = new AuthScope(AuthScope.ANY_HOST, AuthScope.ANY_PORT, "EZID", AuthPolicy.BASIC);
+		Credentials credentials = httpClient.getCredentialsProvider().getCredentials(authScope);
+		assertNotNull(credentials);
+		assertEquals(StackConfiguration.getEzidUsername(), credentials.getUserPrincipal().getName());
+		assertEquals(StackConfiguration.getEzidPassword(), credentials.getPassword());
+	}
+
+	@Test
+	public void testCreate() throws Exception {
```
eric-wu, line 57: Perhaps mocking it as the EZID service seems flaky and takes long to return

lib/lib-doi/src/test/java/org/sagebionetworks/doi/EzidClientTest.java
```diff
@@ -0,0 +1,157 @@
...
+import org.apache.http.params.CoreConnectionPNames;
+import org.apache.http.params.CoreProtocolPNames;
+import org.apache.http.protocol.HTTP;
+import org.junit.Test;
+import org.sagebionetworks.StackConfiguration;
+
+public class EzidClientTest {
+
+	@Test
+	public void testConstructor() throws Exception {
```
eric-wu, line 39: We might as well have a constructor that takes a http client for testing purpose to avoid the reflection in unit test.

PR #351, opened 2013-03-26
one through three now working

john-hill:


services/file/src/main/webapp/chunkedFileUpload.js
```diff
@@ -31,9 +31,11 @@ function getCookie(c_name) {
```
john-hill, line 30: delete this file

PR #369, opened 2013-04-05
PLFM-1648: save EntityBundle with Submission objects

bkng:


lib/jdomodels/src/main/java/org/sagebionetworks/evaluation/dao/SubmissionDAOImpl.java
```diff
@@ -90,10 +94,16 @@
 
 	@Override
 	@Transactional(readOnly = false, propagation = Propagation.REQUIRED)
-	public String create(Submission dto) throws DatastoreException {
+	public String create(Submission dto, EntityBundle bundle) throws DatastoreException, JSONObjectAdapterException {
```
brucehoff, line 97: consider serializing bundle at the mgr level and inserting into dto there.  With such a change, create simply takes a dto as a param.

brucehoff:
Great work, Bennett!!

PR #370, opened 2013-04-08
Implement get() and update() and config change

eric-wu:


lib/lib-doi/src/main/java/org/sagebionetworks/doi/EzidAsyncClient.java
```diff
@@ -14,21 +15,41 @@ public EzidAsyncClient(EzidAsyncCallback createCallback) {
 	}
 
 	@Override
+	public EzidDoi get(String doi, Doi dto) {
+		return ezidClient.get(doi, dto);
```
jay-hodgson, line 23: make async please

PR #404, opened 2013-04-30
PLFM-1808, PLFM-1856 (partial)

eric-wu:
1) Replace the DOI status READY with a new status CREATED (part of PLFM-1856)
2) Refactor RetryableHttpClient into its own class
3) Handle the error where the doi already exists (PLFM-1808)


eric-wu:
Mock RetryableHttpClient in EzidClientTest

PR #405, opened 2013-04-30
Plfm 267: ability to change user email address (do via updating the user profile)

jay-hodgson:


services/auth/src/main/java/org/sagebionetworks/auth/AuthenticationController.java
```diff
@@ -287,6 +237,32 @@ public void setPassword(@RequestBody ChangeUserPassword changeUserPassword,
...
+			@RequestParam(value = AuthorizationConstants.USER_ID_PARAM, required = false) String userId) throws Exception {
+		//user must be logged in to make this request
+		if (userId==null) 
+			throw new AuthenticationException(HttpStatus.BAD_REQUEST.value(), "Not authorized.", null);
+		String registrationToken = registrationInfo.getRegistrationToken();
+		if (registrationToken==null) 
+			throw new AuthenticationException(HttpStatus.BAD_REQUEST.value(), "Missing registration token.", null);
+		
+		String sessionToken  = registrationToken.substring(CrowdAuthUtil.CHANGE_EMAIL_TOKEN_PREFIX.length());
+		String realUserId = CrowdAuthUtil.revalidate(sessionToken);
```
jay-hodgson, line 252: CrowdAuthUtil should be refactored to be a autowired dependency (at the service layer) behind an interface so that we can test.  All this logic should be moved to the service layer (and make sure to test the new changeEmail(), if not everything!)

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/UserProfileManagerImpl.java
```diff
@@ -126,19 +131,30 @@ private UserProfile getAnonymousUserProfile(String principleId){
...
 		boolean canUpdate = UserProfileManagerUtils.isOwnerOrAdmin(userInfo, userProfile.getOwnerId());
 		if (!canUpdate) throw new UnauthorizedException("Only owner or administrator may update UserProfile.");
+		String oldEmail = userInfo.getIndividualGroup().getName();
 		attachmentManager.checkAttachmentsForPreviews(updated);
-		return userProfileDAO.update(updated, schema);
+		UserProfile returnProfile = userProfileDAO.update(updated, schema);
+		
+		//and update email if it is also set (and is different)
+		if (updated.getEmail() != null && !updated.getEmail().equals(oldEmail)) {
+			CrowdAuthUtil.copyUser(oldEmail, updated.getEmail(), rand);
```
jay-hodgson, line 154: make this throw a better exception if user already exists, and then change client to handle it

PR #409, opened 2013-05-03
PLFM-1856 Implement the READY status and PLFM-1834

eric-wu:


lib/lib-doi/src/main/java/org/sagebionetworks/doi/DxAsyncClient.java
```diff
@@ -0,0 +1,128 @@
...
+			throw new IllegalArgumentException("Callback cannot be null.");
+		}
+
+		// Validate the DOI string is in the correct format, e.g. 'doi:10.7303/syn1720822.1'
+		final String doiString = ezidDoi.getDoi();
+		if (doiString == null) {
+			throw new IllegalArgumentException("The DOI string cannot be null.");
+		}
+		final String doiPrefix = "doi:";
+		if (!doiString.startsWith(doiPrefix)) {
```
jay-hodgson, line 54: case insensitive?

PR #448, opened 2013-05-21
Clear a dynamo table via the admin controller and client (plus some refactoring)

eric-wu:
1) Remove the @author tag (changes in the lib-dynamo package)
2) Add DynamoAdminManager and wire it up to the admin controller and client
3) Move dynamo managers to a new 'dynamo' sub-package


integration-test/src/test/java/org/sagebionetworks/IT050SynapseJavaClientDynamo.java
```diff
@@ -32,6 +43,13 @@ public static void beforeClass() throws Exception {
 		synapse.login(StackConfiguration.getIntegrationTestUserOneName(),
 				StackConfiguration.getIntegrationTestUserOnePassword());
 
+		String tableName = DboNodeLineage.TABLE_NAME;
+//		String stackPrefix = StackConfiguration.getStack() + "-" + StackConfiguration.getStackInstance() + "-";
```
eric-wu: Remove the comments

PR #491, opened 2013-06-02
More StorageLocation cleanup and other misc changes

eric-wu:


lib/lib-dynamo/src/main/java/org/sagebionetworks/dynamo/config/DynamoConfig.java
```diff
@@ -26,7 +26,7 @@
 		DynamoKey hashKey = new DynamoKey(DboNodeLineage.HASH_KEY_NAME, ScalarAttributeType.S);
 		DynamoKey rangeKey = new DynamoKey(DboNodeLineage.RANGE_KEY_NAME, ScalarAttributeType.S);
 		DynamoKeySchema keySchema = new DynamoKeySchema(hashKey, rangeKey);
-		DynamoThroughput throughput = new DynamoThroughput(10L, 7L);
+		DynamoThroughput throughput = new DynamoThroughput(10L, 6L);
```
eric-wu: Should 5L.

PR #502, opened 2013-06-07
Add StorageQuota DTO

eric-wu:


lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/dao/DBOStorageQuotaDaoImplAutowiredTest.java
```diff
@@ -48,15 +50,66 @@ public void after() {
...
+		quota = new StorageQuota();
+		quota.setOwnerId(userId);
+		quota.setQuotaInMb(3L);
+		storageQuotaDao.setQuota(quota);
 		quota = storageQuotaDao.getQuota(userId);
-		assertEquals(3, quota.intValue());
-		// Should update the quota when it already exists in the db
-		storageQuotaDao.setQuota(userId, 1);
+		assertNotNull(quota);
+		assertNotNull(quota.getEtag());
```
john-hill, line 63: need a check that the etag is not only not null but also that the etag does not match the first value

lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/dao/DBOStorageQuotaDaoImplAutowiredTest.java
```diff
@@ -48,15 +50,66 @@ public void after() {
...
+	}
+
+	@Test(expected=ConflictingUpdateException.class)
+	public void testSetQuotaConflictingUpdateException() {
+		StorageQuota quota = new StorageQuota();
+		quota.setOwnerId(userId);
+		quota.setQuotaInMb(1L);
+		storageQuotaDao.setQuota(quota);
+		quota.setQuotaInMb(3L);
+		storageQuotaDao.setQuota(quota);
```
john-hill, line 111: need to test for conflict when the etag is not null

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/DBOStorageQuotaDaoImpl.java
```diff
@@ -36,20 +38,35 @@
...
+		if (quotaInMb == null) {
+			throw new IllegalArgumentException("Quota cannot be null.");
+		}
+		if (quotaInMb < 0) {
+			throw new IllegalArgumentException("Storage quota must be >= 0.");
+		}
 
-		DBOStorageQuota dboInUpdate = new DBOStorageQuota();
+		final DBOStorageQuota dboInUpdate = new DBOStorageQuota();
 		final Long userIdLong = KeyFactory.stringToKey(userId);
```
john-hill, line 59: The KeyFactory adds and removes the 'syn' prefix from entity ids and should not be used for anything but entity ids.  User IDs should not start with 'syn'.

PR #512, opened 2013-06-12
Remove dead competition code

eric-wu:


brucehoff:
Not sure where this came from in the first place but, yeah, it looks like dead code.

PR #530, opened 2013-06-25
ACL DAO level refactoring for evaluation ACLs

eric-wu:


eric-wu:
Need  to resend with new changes.

PR #537, opened 2013-06-27
ACL changes per code review feedback

eric-wu:


eric-wu:
1) Correct data migration dependencies
2) SELECT etag FOR UPDATE
3) Remove obsolete NodeBackupManager code
4) Make ACL's ID always a number

services/repository/src/main/java/org/sagebionetworks/repo/web/controller/EntityController.java
```diff
@@ -1144,4 +1138,20 @@ FileHandleResults getEntityFileHandlesForVersion(
 		// pass it!
 		return serviceProvider.getEntityService().getEntityFileHandlesForVersion(userId, id, versionNumber);
 	}
+
```
john-hill, line 1141: will better match the current behavior

PR #545, opened 2013-07-10
Swc 485 & SWC-491

dburdick:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/NodeDAOImpl.java
```diff
@@ -854,9 +854,10 @@ public EntityHeader getEntityHeader(String nodeId, Long versionNumber) throws Da
 		Long id = KeyFactory.stringToKey(nodeId);
 		ParentTypeName ptn = getParentTypeName(id);
 		String versionLabel = null;		
-		if(versionNumber != null) {
-			versionLabel = getVersionLabel(nodeId, versionNumber);
+		if(versionNumber == null) {
+			versionNumber = getCurrentRevisionNumber(nodeId);
```
jay-hodgson, line 858: Is it possible that this returns a null versionNumber (which would now lead to an Illegal Argument Exception in getVersionLabel)?

PR #557, opened 2013-07-17
PLFM-2018: Add message fields to Evaluation object

bkng:


brucehoff:
Great job on such short notice!!

PR #558, opened 2013-07-17
[PATCH] PLFM-2018: add message fields to Evaluation object

bkng:


bkng:
(previously reviewed as https://github.com/Sage-Bionetworks/Synapse-Repository-Services/pull/557)

PR #562, opened 2013-07-18
PLFM-2043: switch to flat columns for new Evaluation fields

bkng:
Also removed auto-generated DateAnnotation, since users could just use LongAnnotations for the same purpose.


eric-wu:
I am curious about the use of blob here.  If you used TEXT instead of BLOB, you could just avoid serialization and deserialization, right?

PR #567, opened 2013-07-22
Attempt to fix data migration failures

eric-wu:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/persistence/DBOAccessControlList.java
```diff
@@ -63,7 +63,11 @@ public String getDDLFileName() {
 	private Long id;
 	private String etag;
 	private Long creationDate;
+	private Long resource;
```
john-hill, line 66: We should fix this by adding a backup object for DBOAccessControlList.  The backup object would then have a long:Resource field that could be ignored by the translator.

eric-wu:
Correct.  Just curious if adding a set method here works vs. adding a brand new class just for a one time migration.  If it does not work, will add the backup class for this stack.

eric-wu:
This simple change fixed the problem.  Will leave the change on release-5 and will NOT merge it down to develop as it is a temporary thing just to make data migration to work on stack-5.

PR #569, opened 2013-07-24
PLFM-2037: update for automated User Profile Service documentation

jay-hodgson:


lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/UserProfile.json
```diff
@@ -154,7 +154,7 @@
 		"teamName": {
 			"type": "string",
 			"title": "Team Name",
-			"description": "Team Name",
+			"description": "This person's default DREAM8 challenge alias/team name",
 			"links": [
```
john-hill, line 158: Isn't profile used by all user's even those that do not participate in DREAM8?  Is this a more genetic field or is it only used for DREAM?

PR #571, opened 2013-07-24
PLFM-2055: (not) race condition in EvaluationManager.update()

bkng:


services/repository-managers/src/main/java/org/sagebionetworks/evaluation/manager/EvaluationManagerImpl.java
```diff
@@ -126,7 +127,9 @@ public Evaluation updateEvaluation(UserInfo userInfo, Evaluation eval)
 			throw new NotFoundException("No Evaluation found with id " + eval.getId());
 		}
 		if (!old.getEtag().equals(eval.getEtag())) {
-			throw new IllegalArgumentException("Your copy of Evaluation " + eval.getId() +
+			// NOTE: we have not yet locked the DB row for update; this will occur at the DAO layer.			
+			// This check is a performance optimization (in the interest of fail-fast behavior).
+			throw new ConflictingUpdateException("Your copy of Evaluation " + eval.getId() +
```
john-hill, line 132: This is not a "performance optimization" because you fetch the entire evaluation row from the DB to check one field.

bkng, line 132: Actually, we fetch the entire evaluation row to ensure that the user has not overwritten the immutable fields 'ownerId' and 'createdOn' (see the validateEvaluation() method)

PR #593, opened 2013-08-02
Remove the turnOffAcl boolean flag

eric-wu:


client/synapseJavaClient/src/main/java/org/sagebionetworks/client/Synapse.java
```diff
@@ -3643,6 +3643,8 @@ public Participant createParticipant(String evalId) throws SynapseException {
 	public Participant getParticipant(String evalId, String principalId) throws SynapseException {
 		if (evalId == null) throw new IllegalArgumentException("Evaluation id cannot be null");
 		if (principalId == null) throw new IllegalArgumentException("Principal ID cannot be null");
+		// Make sure we are passing in the ID, not the user name
+		Long.parseLong(principalId);
```
eric-wu: Rethrow as a better exception

PR #636, opened 2013-09-03
User profile upgrade

kaysoky:
- UserProfile stores a epoch timestamp for Terms Of Use signing time. 
- UserProfile no longer requires ObjectSchema to use


services/workers/src/main/java/org/sagebionetworks/crowd/workers/CrowdGroupSynchronizer.java
```diff
@@ -72,32 +71,48 @@ public void run() {
...
-					ObjectSchema schema = SchemaCache.getSchema(UserProfile.class);
+			}
+			
+			// Make sure a profile exists for each user
+			if (isIndividual) {
+				try {
+					UserProfile userProfile = userProfileDAO.get(principalId);
+					
+					// For profiles that already exist, migrate the boolean for the termsOfUse over
+					if (userProfile.getAgreesToTermsOfUse() == null) {
```
brucehoff: missing case: Crowd boolean changes between worker runs.

PR #646, opened 2013-09-05
Plfm 2130 1

xschildw:


lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/dao/migration/MigatableTableDAOImplAutowireTest.java
```diff
@@ -175,6 +176,135 @@ public void testMigrationRoundTrip() throws Exception {
 	}
 	
 	@Test
+	public void testMigrationRoundTripSome() throws Exception {
+		long startCount = fileHandleDao.getCount();
```
john-hill, line 180: Reduce the number of handles created for this test if possible.

PR #653, opened 2013-09-07
Turning the Crowd Synchronizer back on

kaysoky:
- Increased timeout for ~~two~~ one test
- Minor correction to worker


kaysoky:
This cannot be pulled into the same stack that the NamedToIdSynchronizer is in, since this worker requires the other one to run first.  

So this will probably wait until Stack-13.

kaysoky:
Since we're waiting on a change to _completely_ remove Crowd, this worker will need an overhaul to comply with that requirement.

PR #654, opened 2013-09-07
Plfm 2149

john-hill:


services/workers/src/main/resources/log-collate-spb.xml
```diff
@@ -0,0 +1,47 @@
...
+
+	<!-- This is the trigger used to fire this worker -->
+	<bean id="logCollateWorkerTrigger" class="org.springframework.scheduling.quartz.SimpleTriggerBean"
+		scope="singleton">
+		<property name="jobDetail">
+			<bean
+				class="org.springframework.scheduling.quartz.MethodInvokingJobDetailFactoryBean">
+				<property name="targetObject" ref="logCollateWorkerSemaphoreGatedRunner" />
+				<property name="targetMethod" value="attemptToRun" />
+			</bean>
```
john-hill, line 40: set concurrent to false

services/workers/src/test/java/org/sagebionetworks/log/worker/LogCollateIntegrationTest.java
```diff
@@ -0,0 +1,101 @@
...
+				singleKey = listing.getObjectSummaries().get(0).getKey();
+			}
+			if(singleKey == null){
+				log.debug("Waiting for logs to be collated.  There are currently : "+listing.getObjectSummaries().size()+" log files...");
+				Thread.sleep(1000);
+				long elpase = System.currentTimeMillis()-start;
+				assertTrue("Timed out waiting for the LogCollateWorker to collate all of the logs into a single file.",elpase < MAX_WAIT_MS);
+			}
+		}while(singleKey == null);
+
```
john-hill, line 93: need to use the find log containing key method or this will break as new log data gets written and swept to s3 while the test is running

PR #657, opened 2013-09-09
PLFM-2089:  add content size and md5 to FileVersion

jay-hodgson:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/NodeDAOImpl.java
```diff
@@ -732,6 +735,8 @@ public VersionInfo mapRow(ResultSet rs, int rowNum)
 				info.setVersionNumber(rs.getLong(COL_REVISION_NUMBER));
 				info.setVersionLabel(rs.getString(COL_REVISION_LABEL));
 				info.setVersionComment(rs.getString(COL_REVISION_COMMENT));
+				info.setContentMd5(rs.getString(COL_FILES_CONTENT_MD5));
+				info.setContentSize(rs.getString(COL_FILES_CONTENT_SIZE));
```
john-hill, line 739: Is there a test that checks for these new values?  There should already be a test that get the VersionInfo so we just need asserts that the new fields are set as expected.

PR #658, opened 2013-09-09
added jar signing for File upload Client

dburdick:


client/fileUploadClient/pom.xml
```diff
@@ -77,6 +77,26 @@
...
+							<goal>sign</goal>
+						</goals>
+					</execution>
+				</executions>
+				<configuration>
+					<keystore>SynapseKeystore</keystore>
+					<alias>SynapseKeystore</alias>
+					<storepass>${settings.org.sagebionetworks.stackEncryptionKey}</storepass>
+					<keypass>${settings.org.sagebionetworks.stackEncryptionKey}</keypass>
+				</configuration>
```
john-hill, line 98: Does this mean we are using the stack encryption key to sign the jars?

PR #659, opened 2013-09-09
Group switch

kaysoky:
- Removes Crowd and caching from UserManager, leading to a build-time decrease of a 2-4 minutes
- Removes UserDAOProxy, making all UserDAOs point to the DB
- Refactoring of TestUserDAO into TestGroupMembersDAO
- Lots of additional bootstrapping on non-production stacks.  This reduces dependency on Crowd.

~~Note: tests were run without the CrowdGroupSynchronizer enabled.~~


kaysoky:
Must be pulled into a stack after (#653).  So Stack-14

PR #664, opened 2013-09-10
List all evaluations of a project

kaysoky:
PLFM-2094

### DIFF
--- a/lib/jdomodels/src/main/java/org/sagebionetworks/evaluation/dao/EvaluationDAO.java
+++ b/lib/jdomodels/src/main/java/org/sagebionetworks/evaluation/dao/EvaluationDAO.java
@@ -13,88 +13,53 @@ public interface EvaluationDAO {
 	
 	/**
 	 * Lookup a Evaluation ID by name. Returns null if the name is not in use.
-	 * 
-	 * @param name
-	 * @return
-	 * @throws DatastoreException 
 	 */
 	public String lookupByName(String name) throws DatastoreException;
 
 	/**
 	 * Create a new Evaluation
 	 * 
-	 * @param dto     object to be created
+	 * @param dto object to be created
 	 * @return the id of the newly created object
-	 * @throws DatastoreException
-	 * @throws InvalidModelException
 	 */
 	public String create(Evaluation dto, Long ownerId) throws DatastoreException,
 			InvalidModelException;
 
 	/**
 	 * Creates a Evaluation from backup. The passed eTag is persisted.
-	 * 
-	 * @param dto
-	 * @param ownerId
-	 * @return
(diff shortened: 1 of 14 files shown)


lib/jdomodels/src/main/java/org/sagebionetworks/evaluation/dao/EvaluationDAO.java
```diff
@@ -13,114 +13,74 @@
...
-	 * @param id
-	 * @return
-	 * @throws DatastoreException
-	 * @throws NotFoundException
 	 */
 	public Evaluation get(String id) throws DatastoreException, NotFoundException;
 	
 	/**
+	 * Retrieves an evaluation via the project it draws its content from
+	 * May return null if no such evaluation exists
```
eric-wu: A minor issue here.  The documentation is incorrect.  It returns a list of evaluations not "an evaluation".  Hence the returned list might be empty but should never be null.

services/repository-managers/src/main/java/org/sagebionetworks/evaluation/manager/EvaluationManagerImpl.java
```diff
@@ -84,6 +84,20 @@ public Evaluation getEvaluation(UserInfo userInfo, String id)
 		}
 		return evaluationDAO.get(id);
 	}
+	
+	@Override
+	public QueryResults<Evaluation> getEvaluationByContentSource(UserInfo userInfo, String projectId, long limit, long offset)
+			throws DatastoreException, NotFoundException {
```
eric-wu, line 90: I would insert after here precondition checks for the parameters passed in.  For example, if they are null, throw IllegalArgumentException.

services/repository/src/test/java/org/sagebionetworks/repo/web/controller/EvaluationControllerAutowiredTest.java
```diff
@@ -451,6 +451,12 @@ public void testPaginated() throws DatastoreException, JSONObjectAdapterExceptio
 		for (Evaluation c : evals.getResults())
 			assertTrue("Unknown Evaluation returned: " + c.toString(), c.equals(eval1) || c.equals(eval2));
 		
+		// paginated evaluations by content source
+		evals = entityServletHelper.getEvaluationsByContentSourcePaginated(ownerName, KeyFactory.SYN_ROOT_ID, 10, 0);
+		assertEquals(2, evals.getTotalNumberOfResults());
+		for (Evaluation c : evals.getResults())
```
eric-wu: Please add braces for the for-loop block to improve readability and avoid hard-to-spot bugs.

PR #665, opened 2013-09-11
Add a trash worker that purges trash older than a month

eric-wu:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/trash/TrashManagerImpl.java
```diff
@@ -298,24 +304,12 @@ public void purgeTrash(UserInfo currentUser)
...
-		}
+		purgeTrash(trashList);
 	}
 
-	private void purge(List<TrashedEntity> trashList)
+	@Transactional(readOnly = false, propagation = Propagation.REQUIRED)
+	@Override
+	public void purgeTrash(List<TrashedEntity> trashList)
 			throws DatastoreException, NotFoundException {
 		// For subtrees moved entirely into the trash can, we want to find the roots
```
eric-wu, line 329: Sort the trashed items by ID to avoid deadlocks.

services/workers/src/main/java/org/sagebionetworks/trash/worker/TrashWorkerFactory.java
```diff
@@ -0,0 +1,24 @@
...
+	@Autowired
+	private TrashManager trashManager;
+
+	@Override
+	public void run() {
+		TrashWorker worker = new TrashWorker(trashManager);
+		try {
+			worker.purgeTrash();
+		} catch (DatastoreException e) {
+			throw new RuntimeException(e);
```
eric-wu: Log instead of exceptions.

PR #673, opened 2013-09-14
Removes the old way of serializing objects from a schema.

kaysoky:
Meant for Stack-14, when all UserProfiles have been migrated using the newer method.


kaysoky:
We'll need to activate a worker that updates all UserProfiles through the DTO level before this can go in.  The migrator only hits the DBO level.

PR #684, opened 2013-09-18
Replacement for the RepositoryMessagePublisher.timerFiredFindUnsentMessages() worker

kaysoky:
- Moves the functionality to Services-Workers
- Splits the work into smaller batches that can be scaled up if necessary
- Fix for an issue in MessageReceiverImpl where certain configurations will result in SQS complaints

### DIFF
--- a/services/workers/src/main/resources/message-popper-sqs-sns.xml
+++ b/services/workers/src/main/resources/message-popper-sqs-sns.xml
@@ -14,24 +14,13 @@
 	<import resource="classpath:aws-spb.xml" />
 	<import resource="classpath:cloudwatch-spb.xml" />
 	<import resource="classpath:dao-beans.spb.xml" />
-
-	<bean id="awsSQSClient"
-			class="com.amazonaws.services.sqs.AmazonSQSClient"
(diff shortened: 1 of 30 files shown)


services/workers/src/main/resources/message-popper-sqs-sns.xml
```diff
@@ -0,0 +1,66 @@
...
+		<property name="topicName" ref="stackConfiguration.repositoryChangeTopicName" />
+	</bean>
+		
+	<bean id="unsentMessagePopper"
+		class="org.sagebionetworks.asynchronous.workers.sqs.MessageReceiverImpl"
+		scope="singleton">
+		<property name="messageQueue" ref="unsentMessageQueue" />
+		<property name="workerFactory" ref="unsentMessagePopperFactory" />
+		<property name="maxNumberOfWorkerThreads" ref="stackConfiguration.semaphoreGatedMaxRunnersUnsentMessagePoppers" />
+		<property name="maxMessagePerWorker" value="10" />
```
john-hill: set to one

PR #691, opened 2013-09-24
Fix for UnsentMessageQueuer

kaysoky:
- ...  for sending more than 10 messages
- Configuration change to send smaller messages and hold them for sufficiently long

### DIFF
--- a/services/workers/src/main/java/org/sagebionetworks/message/workers/UnsentMessageQueuer.java
+++ b/services/workers/src/main/java/org/sagebionetworks/message/workers/UnsentMessageQueuer.java
@@ -6,6 +6,7 @@ import java.util.List;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.sagebionetworks.asynchronous.workers.sqs.MessageQueue;
+import org.sagebionetworks.asynchronous.workers.sqs.MessageReceiverImpl;
 import org.sagebionetworks.repo.model.dbo.dao.DBOChangeDAO;
 import org.sagebionetworks.repo.model.message.UnsentMessageRange;
 import org.sagebionetworks.schema.adapter.JSONObjectAdapterException;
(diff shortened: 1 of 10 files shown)


services/workers/src/main/java/org/sagebionetworks/message/workers/UnsentMessageQueuer.java
```diff
@@ -64,11 +66,19 @@ public void run() {
...
 		List<SendMessageBatchRequestEntry> batch = buildRangeBatch();
-		awsSQSClient.sendMessageBatch(new SendMessageBatchRequest(unsentMessageQueue.getQueueUrl(), batch));
+		for (int i = 0; i < batch.size(); i += MessageReceiverImpl.SQS_MAX_REQUEST_SIZE) {
+			List<SendMessageBatchRequestEntry> miniBatch = batch.subList(i, 
+					(i + MessageReceiverImpl.SQS_MAX_REQUEST_SIZE > batch.size() 
+							? batch.size() 
+							: i + MessageReceiverImpl.SQS_MAX_REQUEST_SIZE));
+			awsSQSClient.sendMessageBatch(new SendMessageBatchRequest(unsentMessageQueue.getQueueUrl(), miniBatch));
+		}
+		log.info("Just queued: " + batch.size() + " message(s)");
```
john-hill: There is a bug in this code.  Please write a test that catches the bug.

kaysoky: You mean the test "testQueueMoreThan10Messages" below?

john-hill: No the test below does not cover much.  In fact, the only thing it would catch is if Amazon throws an exception.
This code code needs three test case:
1. Batch size is less than max and all get sent.
2. Batch size is a multiple of max and all get sent.
3. Batch size is a multiple of max +1 and all get sent.
This type of testing is best done with Mocks.

PR #702, opened 2013-09-30
Authentication layers

kaysoky:
- Adds dangling implementations of the DAO, manager, controller, and filter necessary for the removal of Crowd
- Tests for the filter are disabled
- Adds password hashing utility
- Adds more functionality to the auth service
- Removes the ResourceAccess services


lib/models/src/main/java/org/sagebionetworks/repo/model/AuthenticationDAO.java
```diff
@@ -0,0 +1,58 @@
...
+	 */
+	public String getSecretKey(String id);
+	
+	/**
+	 * Generates a new secret key for the user
+	 */
+	public void changeSecretKey(String id);
+	
+	/**
+	 * Replaces the user's secret key with the specified one
```
john-hill, line 54: comment about the temp nature of this method. maybe deprecate it.

kaysoky:
To be broken into smaller pulls.

PR #710, opened 2013-10-08
Disconnection of Crowd from Authentication

kaysoky:
- Replaces authentication controller, service, and filter with a non-Crowd version
- Removes some components of Crowd (parts that are not needed by the migrator)
- Adds extra bootstrapping to account for the state normally maintained by Crowd-dev


kaysoky:
This pull is probably too large, I'll break it down as much as possible.

PR #713, opened 2013-10-08
Preparation to switch away from Crowd

kaysoky:
- Adds extra bootstrapping steps
- Adds a few extra methods to user-related managers
- Cleanup of a few interfaces
- Adds the (not yet connected) Crowd-less auth filter


kaysoky:
These files have undergone extensive changes, so I'll re-split the commit into smaller chunks.

PR #718, opened 2013-10-09
Pre Crowd removal - #1 Bootstrapping

kaysoky:
- Sets up the user state required to run the tests without Crowd as a state-store
- Includes the special limited Portal-user


kaysoky:
Design has changed...

PR #719, opened 2013-10-09
Pre Crowd removal - #2 Layer cleanup

kaysoky:
- A fix for the auth manager
- Addition of a createUser method for the UserManager
- Addition of a acceptTermsOfUse method for the UserProfileManager


kaysoky:
Design has changed...

PR #724, opened 2013-10-11
Processed Msgs table

xschildw:
Processed msgs table


lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/persistence/DBOProcessedMessageTest.java
```diff
@@ -0,0 +1,23 @@
+package org.sagebionetworks.repo.model.dbo.persistence;
+
+import org.junit.Test;
+
+public class DBOProcessedMessageTest {
+	@Test(expected=NullPointerException.class)
```
xschildw, line 6: Should be IllegalArgument and change corresponding in method

PR #725, opened 2013-10-11
Migrate crowd

xschildw:
Added call to migrate crowd


tools/tool-migration-utility/src/main/java/org/sagebionetworks/tool/migration/v3/MigrationClient.java
```diff
@@ -416,4 +422,33 @@ private void dumpDeferredExceptions() {
...
+		}
+		if (batchFailed) {
+			throw new RuntimeException("Failed during Crowd migration");
+		}
+	}
+	
+	public boolean containsFailure(List<CrowdMigrationResult> batch) {
+		boolean failed = false;
+		for (CrowdMigrationResult r: batch) {
+			if (r.getResultType() != CrowdMigrationResultType.SUCCESS) {
```
kaysoky: The other type of result (ABORT) is also considered a success.  It simply means the user does not exist in RDS, and therefore should not be migrated.

PR #726, opened 2013-10-11
Crowd disconnection

kaysoky:
Since Crowd is no longer used as a state-store for user information, the migration admin should not be able to delete itself via the MigrationManager.deleteAll() method or equivalent.  
- Takes Crowd out of the UserDAO
- Rewires the auth controller away from Crowd
- Removes the /login/v1 endpoint


kaysoky:
Probably will be unmergable after code-review fixes.

PR #764, opened 2013-10-24
Improvements to OpenID code

kaysoky:
Makes it easier to add more OpenID providers in the future.


kaysoky:
Note: still requires manual testing

PR #770, opened 2013-10-25
[PLFM-2256&7] V2 WikiPage work

heidiso:


client/synapseJavaClient/src/main/java/org/sagebionetworks/client/SynapseClient.java
```diff
@@ -462,6 +465,48 @@ public FileHandleResults getEntityFileHandlesForVersion(String entityId,
 			Long versionNumber) throws JSONObjectAdapterException,
 			SynapseException;
 
+	public V2WikiPage createV2WikiPage(String ownerId, ObjectType ownerType,
```
john-hill, line 468: I know many of the methods in this interface do not have comments but that is trend we must reverse.

john-hill:
Looks good.

PR #773, opened 2013-10-28
Re-enable group caching

kaysoky:
Adds a worker to periodically (12 minutes) check all users in the system and their group memberships.


kaysoky:
Nesting will be disabled, so this stuff is not needed :frowning:

PR #775, opened 2013-10-29
PLFM-2269

brucehoff:


brucehoff:
Jay,
This morning I had two pull requests.  I thought you pulled both in but this one was not pulled in:
https://github.com/Sage-Bionetworks/Synapse-Repository-Services/pull/775
I'm pulling it now, because I'm blocked unless I have this small change.  If we need to discuss further, please let me know.
Thanks,
Bruce

PR #779, opened 2013-10-31
Plfm 2253

john-hill:


eric-wu:
Build time reduced => productivity increased. Nice.

PR #788, opened 2013-11-01
409 on create existing user

kaysoky:
POST /user with an existing user should throw a 409, not a 401


services/repository/src/main/java/org/sagebionetworks/auth/services/AuthenticationServiceImpl.java
```diff
@@ -109,7 +110,7 @@ public void createUser(NewUser user) throws UnauthorizedException {
 		try {
 			userManager.createUser(user);
 		} catch (DatastoreException e) {
-			throw new UnauthorizedException("User '" + user.getEmail() + "' already exists", e);
+			throw new NameConflictException("User '" + user.getEmail() + "' already exists", e);
```
brucehoff, line 113: call userGroupDAO.doesPrincipalExist(name) (or whatever) and if so, throw the NameConfictException, else propagate the DataStoreException

kaysoky:
Combining this pull with a related one.

PR #789, opened 2013-11-01
Disable auto-create user via OpenID

kaysoky:
- Add parameter (default false) for creating a user if it doesn't exist already


services/repository/src/main/java/org/sagebionetworks/auth/AuthenticationController.java
```diff
@@ -220,7 +220,8 @@ public void invalidateSecretKey(
 	
 	/**
 	 * To authenticate via OpenID, this service takes all URL parameters returned by the OpenID provider (i.e. Google)
-	 * along with an optional parameter to explicitly accept the terms of use (org.sagebionetworks.acceptsTermsOfUse=true).
+	 * along with an optional parameter to explicitly accept the terms of use (org.sagebionetworks.acceptsTermsOfUse=true)
+	 * and an optional parameter to create a user account if the OpenID is not registered in Synapse (org.sagebionetworks.createUserIfNecessary=true).
```
brucehoff, line 235: If create...=false and user does not exist, then the service returns status 404

PR #804, opened 2013-11-06
Messaging - DAO changes

kaysoky:
Adds two join tables for messages to threads and threads to objects.
Each message can belong to one thread and each thread can belong to one object.


lib/jdomodels/src/main/resources/schema/Message-ddl.sql
```diff
@@ -1,14 +1,14 @@
 CREATE TABLE `MESSAGE` (
   `ID` bigint(20) NOT NULL,
-  `THREAD_ID` bigint(20) NOT NULL,
```
brucehoff: consider removing thread_id from Message and just having reply_to.  We'd then have a separate "reporting" table with two columns, (1) message_id (FK to this table's ID) and (2) THREAD, also a FK to this table's ID).  The ID of the first message in a thread is the ID of the thread.

In the reporting table, the Thread column is a foreign key to the message column.

PR #806, opened 2013-11-07
Split up synapse java client to support bridge client

marcel-blonk:


kaysoky:
This will conflict with https://github.com/Sage-Bionetworks/Synapse-Repository-Services/pull/803

Given the relative sizes of the pulls, this one should go in first if possible.

PR #811, opened 2013-11-11
Reduce num of wikis being migrated at a time

heidiso:


tools/tool-migration-utility/src/main/java/org/sagebionetworks/tool/migration/v3/MigrationClientMain.java
```diff
@@ -47,7 +47,12 @@ public static void main(String[] args) throws Exception {
 			System.exit(-1);
 		} else {
 			// Migrate wikis from V1 to V2 only if everything else migrated successfully
-			client.migrateWikisToV2();
+			try {
+				client.migrateWikisToV2();
+			} catch(Exception e) {
+				log.error("Migration of wikis to V2 failed.");
```
john-hill, line 53: Please include the full stack trace of the exception in the log.

PR #814, opened 2013-11-12
Etags for credentials

kaysoky:
As a secondary table of the UserGroup table, any change that changes credentials should touch the etag.


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/DBOAuthenticationDAOImpl.java
```diff
@@ -150,6 +150,8 @@ public Long checkEmailAndPassword(String email, String passHash) throws Unauthor
 	@Override
```
john-hill, line 150: The only thing missing is the tests that validate that etags change for each case.

PR #817, opened 2013-11-12
added size checking on all large requests

john-hill:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/table/ColumnConstants.java
```diff
@@ -0,0 +1,29 @@
...
+public class ColumnConstants {
+
+	/**
+	 * The maximum number of characters in a string column.
+	 */
+	public static final int MAX_CHARS_IN_STRING_COLUMN = 2000;
+	/**
+	 * The maximum number of bytes of a string.
+	 */
+	public static final int MAX_STRING_BYTES = new String(new char[MAX_CHARS_IN_STRING_COLUMN]).getBytes().length;
```
marcel-blonk: The correctness of this max number depends on the default encoding. '\0' is probably shortest possible character in most encodings.

john-hill: Good catch!

They should all be based on UTF-8 as that is what is used store the data.

john-hill: The max string now uses Character.MAX_VALUE converted to UTF-8 to calculate the max bytes per char in UTF-8.

client/synapseJavaClient/src/main/java/org/sagebionetworks/client/SynapseClientImpl.java
```diff
@@ -4192,6 +4195,14 @@ public UserEvaluationPermissions getUserEvaluationPermissions(String evalId) thr
 	}
 
 	@Override
+	public RowReferenceSet appendRowsToTable(RowSet toAppend) throws SynapseException {
+		if(toAppend == null) throw new IllegalArgumentException("RowSet cannot be null");
+		if(toAppend.getTableId() == null) throw new IllegalArgumentException("RowSet.tableId cannot be null");
+		String url = getRepoEndpoint()+ENTITY+"/"+toAppend.getTableId()+TABLE;
```
brucehoff, line 4201: ROW  rather than TABLE?

PR #819, opened 2013-11-12
Use iterator for safe modification of lists

heidiso:


tools/tool-migration-utility/src/main/java/org/sagebionetworks/tool/migration/v3/MigrationClient.java
```diff
@@ -143,9 +144,11 @@ public void migrateAllTypes(long batchSize, long timeoutMS, int retryDenominator
...
 		// Remove V2 Wiki Primary types to keep V2 data in place
-		for(MigrationType type: primaryTypes) {
+		ListIterator<MigrationType> iterator = primaryTypes.listIterator();
+		while(iterator.hasNext()) {
+			MigrationType type = iterator.next();
 			if(type.equals(MigrationType.V2_WIKI_PAGE) || type.equals(MigrationType.V2_WIKI_OWNERS)) {
-				primaryTypes.remove(type);
+				iterator.remove();
 			}
 		}
```
john-hill, line 152: A common solution to this problem is to build a new collection using the old.  Iterate over the old list, and add everything you want to keep to a new list. Then throw away the old and use the new.

PR #824, opened 2013-11-14
Messaging - Manager level

kaysoky:
Includes DAO-level code review changes

### DIFF
--- a/services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/MessageManagerImpl.java
+++ b/services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/MessageManagerImpl.java
@@ -1,0 +1,186 @@
+package org.sagebionetworks.repo.manager;
+
+import java.util.ArrayList;
+import java.util.List;
+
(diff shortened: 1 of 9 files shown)


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/MessageManagerImpl.java
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	@Override
+	public void markMessageStatus(UserInfo userInfo, String messageId, MessageStatusType status) {
+		messageDAO.updateMessageStatus(messageId, userInfo.getIndividualGroup().getId(), status);
+	}
+
+	@Override
+	@Transactional(readOnly = false, propagation = Propagation.REQUIRED)
+	public List<String> sendMessage(String messageId) throws NotFoundException {
```
brucehoff, line 106: please ensure there is a test which tries to send a message to AUTHENTICATED_USERS.  (It should not be possible to do so.)

PR #826, opened 2013-11-14
Address PLFM-2179 by inserting 'rolling' to the files that are still rolling

eric-wu:
...ng rolled over so that the files without the 'rolling' name can be picked up by dashboard for metrics.

### DIFF
--- a/services/workers/src/main/java/org/sagebionetworks/audit/worker/MergeWorker.java
+++ b/services/workers/src/main/java/org/sagebionetworks/audit/worker/MergeWorker.java
@@ -1,6 +1,7 @@
 package org.sagebionetworks.audit.worker;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.LinkedList;
 import java.util.List;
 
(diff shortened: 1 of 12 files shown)


services/workers/src/main/java/org/sagebionetworks/audit/worker/MergeWorker.java
```diff
@@ -119,7 +123,7 @@ private boolean mergeBatch(BatchData data) {
 					// Use the time stamp from the first record as the time stamp for the new batch.
 					long timestamp = mergedBatches.get(0).getTimestamp();
 					// Save the merged batches
-					String newfileKey = accessRecordDAO.saveBatch(mergedBatches, timestamp);
+					String newfileKey = accessRecordDAO.saveBatch(mergedBatches, timestamp, false);
```
john-hill, line 127: This will not work.  This new file could still be "rolling".  If we make this change we will end up with thousands of tiny files that will bring data processing to a crawl.

services/workers/src/test/resources/test-context.xml
```diff
@@ -62,4 +62,20 @@
...
+				<property name="targetMethod" value="attemptToRun" />
+				<property name="concurrent" value="false" />
+			</bean>
+		</property>
+		<!-- We stager the start delay of each trigger to spread out the timing -->
+		<property name="startDelay" value="31" />
+		<!-- This is a slow worker and does not need to be run more than once every 1 hour. -->
+		<property name="repeatInterval" value="3000" />
+	</bean>
+
```
john-hill, line 80: use stack config for only the repeat interval.

PR #841, opened 2013-11-20
[PLFM-2359] Add new V2 Wiki service call to get a specific version

heidiso:


heidiso:
These changes were combined with the other service additions to the V2 wiki service.
https://github.com/Sage-Bionetworks/Synapse-Repository-Services/pull/863

PR #844, opened 2013-11-21
MergeWorkerTest - Fix #1 (of 3 problems)

kaysoky:
Get a lock on the MergeWorker before running the test.  This prevents 1 of 3 possible ways to fail the test.

See https://sagebionetworks.jira.com/browse/PLFM-2360

### DIFF
--- a/services/workers/src/test/java/org/sagebionetworks/audit/worker/MergeWorkerTest.java
+++ b/services/workers/src/test/java/org/sagebionetworks/audit/worker/MergeWorkerTest.java
@@ -3,6 +3,7 @@ package org.sagebionetworks.audit.worker;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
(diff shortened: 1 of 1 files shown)


services/workers/src/test/java/org/sagebionetworks/audit/worker/MergeWorkerTest.java
```diff
@@ -23,18 +26,35 @@
...
+	private SemaphoreDao semaphoreDAO;
+	
+	private String lockToken;
+	
+	@Before
+	public void before() {
+		// Prevent the main scheduler from running another instance of the MergeWorker
+		lockToken = semaphoreDAO.attemptToAcquireLock(SEMAPHORE_KEY, MAX_WAIT);
+		if (lockToken == null) {
+			fail();
```
john-hill: The test cannot fail if it fails to get the lock.  It keep trying until a timeout or the lock is acquired.

kaysoky: Ok, changed.

PR #866, opened 2013-11-27
V2 Wiki Service Additions

heidiso:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/wiki/V2WikiManagerImpl.java
```diff
@@ -153,15 +153,15 @@ public V2WikiPage getRootWikiPage(UserInfo user, String objectId,	ObjectType obj
 		Long rootWikiId = wikiPageDao.getRootWiki(objectId, objectType);
 		WikiPageKey key = new WikiPageKey(objectId, objectType, rootWikiId.toString());
 		// The security check is done in the default method.
```
john-hill, line 155: Both create and update of a wiki page needs to validate the caller is the creator of new file handles introduced to the wiki.

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/wiki/V2WikiManagerTest.java
```diff
@@ -167,6 +182,22 @@ public void testFileHandleIdForFileNameAuthorized() throws DatastoreException, N
...
+	@Test
+	public void testMarkdownFileHandleIdForVersion() throws UnauthorizedException, NotFoundException {
+		WikiPageKey key = new WikiPageKey("123", ObjectType.EVALUATION, "345");
+		when(mockAuthManager.canAccess(any(UserInfo.class), any(String.class), any(ObjectType.class), any(ACCESS_TYPE.class))).thenReturn(true);
+		wikiManager.getMarkdownFileHandleId(new UserInfo(false), new WikiPageKey("123", ObjectType.EVALUATION, "345"), new Long(0));
+		verify(mockWikiDao, times(1)).getMarkdownHandleId(key, new Long(0));
+	}
+	
 	@Test (expected=UnauthorizedException.class)
 	public void testGetTreeUnauthorized() throws DatastoreException, NotFoundException{
```
john-hill, line 202: add a test where the user tries to update a wiki page using a filehandle id they not own (same for create).

Also include the case where one user updates a wiki makrdown with a file handle then a second user adds an attachment filehandle but does not change the markdown file handle of the first caller.  This should not fail.

PR #869, opened 2013-12-02
user name (email?) --> principalId for adding/removing admin methods on bridge client

alxdarksage:


integration-test/src/test/java/org/sagebionetworks/IT600BridgeCommunities.java
```diff
@@ -82,6 +87,13 @@ public static void beforeClass() throws Exception {
 
 		bridgeTwo = createBridgeClient(StackConfiguration.getIntegrationTestUserTwoName(),
 				StackConfiguration.getIntegrationTestUserTwoPassword());
+		
+		// Calling createSynapse(bridgeTwo).getUserSessionData().getProfile().getOwnerId();
+		// generates an "invalid session token" error. Do not know why.
+		SynapseClientImpl synapse = createSynapse();
```
marcel-blonk: That invalid session token needs to be figured out. The test here is partially to test the creation of a synapse client based on a bridge client. We can't just taking that test out. It'll come back to bite us.

alxdarksage: Going to ask Joseph about this tomorrow: I think it's a bug.

kaysoky: Let me replicate it and see what I find.

kaysoky: Not a bug in the auth services.  You're missing a line in the createSynapse(BridgeClient bridge) method

``` java
synapse.setAuthEndpoint(StackConfiguration.getAuthenticationServicePrivateEndpoint());
```

PR #873, opened 2013-12-03
Email via SES

kaysoky:
Replaces the EmailUtils and SendMail classes with the messaging framework.  Emailing is stubbed out for all testing below the integration test (java client) level.  

Note: will not work until the sagebase.org domain is verified with SES.  But the code is ready for review.


lib/stackConfiguration/src/main/java/org/sagebionetworks/StackConfiguration.java
```diff
@@ -202,6 +202,10 @@ public static String getFileServiceEndpoint() {
 	public static String getBridgeServiceEndpoint() {
 		return configuration.getBridgeServiceEndpoint();
 	}
+	
+	public static String getPortalEndpoint() {
+		return configuration.getPortalEndpoint();
```
brucehoff: what exactly is the value for a deployed stack?

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/MessageManagerImpl.java
```diff
@@ -225,25 +311,93 @@ public void markMessageStatus(UserInfo userInfo, MessageStatus status) throws No
...
-		
+	private List<String> processMessage(String messageId, boolean singleTransaction) throws NotFoundException {
 		MessageToUser dto = messageDAO.getMessage(messageId);
+		String messageBody = downloadEmailContent(dto.getFileHandleId());
+		return processMessage(dto, singleTransaction, messageBody);
+	}
+	
+	/**
+	 * See {@link #processMessage(String, boolean)}
+	 * Also used by {@link #sendEmail(org.sagebionetworks.repo.manager.MessageManager.EMAIL_TEMPLATE, String, String, Map, boolean)}
```
brucehoff: comment out of date

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/MessageManagerImpl.java
```diff
@@ -279,53 +435,87 @@ public void markMessageStatus(UserInfo userInfo, MessageStatus status) throws No
...
-		}
+		return recipients;
+	}
+	
+	/**
+	 * Helper for {@link #processMessage(String, boolean, String)}
+	 * 
+	 * Returns a string containing the body of a message
+	 */
+	private String downloadEmailContent(String fileHandleId) throws NotFoundException {
```
brucehoff, line 445: Security hole: check that the user can access the file before creating the file

lib/stackConfiguration/src/main/java/org/sagebionetworks/StackConfiguration.java
kaysoky: Looks like the bug I was experiencing the auth integration tests was related to this.
I'll remove this method since it seems to not be used anywhere else and doesn't return a useful value.

PR #877, opened 2013-12-03
Factor out TagMessenger and EtagGenerator and (deprecated) CRUD from backup

kaysoky:
The TagMessenger acts as a middle man that performs both etag changes and the sending of change messages.  That functionality is overloaded and should be separate.  
See: https://sagebionetworks.jira.com/browse/PLFM-2364

Since the EtagGenerator is tied closely with the TagMessenger, it is being removed along side this change
See: https://sagebionetworks.jira.com/browse/PLFM-2366

Also, many instances where these two classes are used lie within deprecated methods from V2 migration.  So those are being removed too.


john-hill:
Looks good!

PR #884, opened 2013-12-05
Addressed review comments, added tests and reworked a bit. Added support for blob and fixed support for datetime

marcel-blonk:


kaysoky:
I think we've moved away from the datetime type to Long, since SQL loses the millisecond increment when we use datetime.  

Also, is support for enums planned for future development?  When I tried to convert 5 messaging-related DBOs to use your annotations, I could not convert two of them due to the presence of an enum field.

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/DBOBuilder.java
```diff
@@ -71,34 +117,58 @@ public FieldColumn apply(Entry<Field> fieldEntry) {
 		return result.toArray(new FieldColumn[result.size()]);
 	}
 
-	private static final Object[][] TYPE_MAP = {
-		{ Long.class, "getLong" },
-		{ String.class, "getString" },
-		{ java.util.Date.class, "getDate" }
-	};
+	private static final Object[][] TYPE_MAP = { { Long.class, "getLong" }, { String.class, "getString" } };
```
kaysoky: If the field being mapped is nullable, you should use 

``` java
String res = getString(<fieldname>);
if (res != null) {
    return Long.parseLong(res);
}
```

i.e. If the type is a Long, your mapper returns a 0 when we expect a null.

marcel-blonk:
See updated pull request. I took out support for Date (can't support Date
as a type, because spring will have inconsistent expectations if the db
field is actually a long).
Enums are support with this pull request, stored as a char(32). If that is
not what you need, we should talk.

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/DBOBuilder.java
marcel-blonk: Good catch. See updated pull request

lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/DBOAnnotatedExample.java
```diff
@@ -0,0 +1,200 @@
...
+	private String comment;
+	@Field(name = "NAME", fixedchar = 16, defaultNull = true)
+	private String name;
+	@Field(name = "ETAG", etag = true, defaultNull = true)
+	private String etag;
+	@Field(name = "MODIFIED_BY", varchar = 256, nullable = false)
+	private String modifiedBy;
+	@Field(name = "MODIFIED_ON", nullable = false)
+	private Long modifiedOn;
+
```
kaysoky, line 43: Example of enum usage?

marcel-blonk, line 43: I see the issue. Spring jdbc only works one way on enums. Working on a fix for that.

PR #885, opened 2013-12-05
[PLFM-2405] Zip/Compressing markdown

heidiso:
Now:
- When mirroring CRUD of V1 Wikis, its markdown is zipped/compressed
- When doing the back-fill, the markdown of each wiki page will be zipped/compressed


kaysoky:
It should be possible to move the FileUtils to a different project so that both java client and services can depend on it, rather than duplicating the logic.  For example, the lib-securityUtilities or auth-util libraries can be used by both.

heidiso:
Thanks for pointing that out; I just moved them, and removed the repeated logic.

PR #890, opened 2013-12-06
moved to a new ID generator

john-hill:


lib/id-generator/src/main/java/org/sagebionetworks/ids/IdGenerator.java
```diff
@@ -16,7 +16,8 @@
 		FAVORITE_ID,
 		ACL_RES_ACC_ID,
 		COLUMN_MODEL_ID, 
-		MESSAGE_ID
+		MESSAGE_ID,
+		PRINCIPAL_ID
```
kaysoky, line 20: Looking at the older code, it looks like we used to use the DOMAIN_IDS type for UserGroups.  Should we remove that type now that it is no longer used?

kaysoky, line 20: Oops, nevermind.  That ID is still used.

PR #902, opened 2013-12-13
Plfm 2422

john-hill:


kaysoky:
I think I should do this merge.  Some parts of the diff are incorrect.

PR #919, opened 2013-12-17
Humongous Refactor - Part 2

kaysoky:
Changes the AuthFilter so that it returns a Long in the UserID param.

Cascades that change down to controllers -> services ( -> and sometimes managers) by changing the type from String to Long.  
Also enforces the name of the variable to be userId.


kaysoky:
I got a clean build before rebasing onto Heidi's code.  So this portion is ready for review while we wait for her to fix the build.

Closing for now though, since I'll need to rebase again after she fixes the tests.

PR #921, opened 2013-12-17
Fix Develop build

heidiso:
Edit test / create file handles before assigning to wiki pages


kaysoky:
Still broken.  See: https://github.com/john-hill/Synapse-Repository-Services/commit/cf43135a5b6465ff6326638638750dcb7d314115#diff-10

The file handle you create must contain a UUID.  That UUID is then included in the search query.  Right now, the test is searching for a UUID that is not in the file.  And the file is never uploaded to CloudSearch (since it doesn't point to an actual file).

PR #950, opened 2013-12-20
Soundex query

kaysoky:
Adds an additional required column (filled in by the DAO) in the PrincipalHeader table that can be used to match by soundex


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/DBOPrincipaHeaderDAOImpl.java
```diff
@@ -83,6 +100,16 @@ public Long mapRow(ResultSet rs, int rowNum)
...
+
+	/**
+	 * Called by Spring after bean initialization
+	 * 
+	 * Creates an Apache Soundex encoder 
+	 */
+	@Override
+	public void afterPropertiesSet(){
+		soundex = new Soundex();
+	}
```
john-hill: We believe a new Soundex() at runtime would be a better (and simpler) option here.

PR #954, opened 2013-12-20
Messaging - improvements to comments and error messages

kaysoky:
The message DAO already throws the correct exception (illegal argument) when invalid recipients are provided, but that error message is ugly.  The manager should check the recipients list and throw a prettier error.
- Also, improve javadocs for messaging


brucehoff:
Looks great!  I'm pulling it in.

PR #955, opened 2013-12-21
Fix cleanup issue

xschildw:


kaysoky:
Looks good.  I also changed the exception catching on the deleteFileHandle(...) to be more specific (error codes 404 or 500+).  But that makes no difference to the cleanup failing or not.

PR #957, opened 2014-01-01
Fix potential threading issue

marcel-blonk:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/message/RepositoryMessagePublisherImpl.java
```diff
@@ -129,7 +129,9 @@ public void fireChangeMessage(ChangeMessage message) {
 		if(message.getObjectType()  == null) throw new IllegalArgumentException("ChangeMessage.getObjectType() cannot be null");
 		if(message.getTimestamp()  == null) throw new IllegalArgumentException("ChangeMessage.getTimestamp() cannot be null");
 		// Add the message to a queue
-		messageQueue.get().add(message);
+		synchronized (this) {
```
john-hill, line 132: The point of the AtomicReference was to avoid locking (synchronized block).

Why is the use of AtomicReference in this case "incorrect"?

PR #963, opened 2014-01-06
If user has not created PD for type, return empty result set (do not thr...

alxdarksage:
...ow exception)


services/repository-managers/src/main/java/org/sagebionetworks/bridge/manager/participantdata/ParticipantDataManagerImpl.java
```diff
@@ -61,4 +68,14 @@ public PaginatedRowSet getData(UserInfo userInfo, String participantDataId, Inte
 		paginatedRowSet.setTotalNumberOfResults(totalNumberOfResults);
 		return paginatedRowSet;
 	}
+
+	private PaginatedRowSet getEmptyPaginatedRowSet() {
+		PaginatedRowSet paginatedRowSet = new PaginatedRowSet();
+		RowSet rowSet = new RowSet();
+		rowSet.setHeaders(new ArrayList<String>());
```
marcel-blonk, line 75: Collections.emptyList()

PR #1006, opened 2014-01-21
PLFM-2509

brucehoff:
This PR addresses two problems:
1) Fixed a bug in HttpClientHelperException constructor which was throwing and swallowing an exception;
2) In SharedClientConnection, user exceptions should wrap user messages but instead these exceptions were being passed bundles of information suitable for logging. For such user exceptions I stripped away all but the user messages.


client/synapseJavaClient/src/main/java/org/sagebionetworks/client/SynapseClient.java
```diff
@@ -159,9 +159,6 @@ public StackStatus getCurrentStackStatus()
 	 */
 	public String getFileEndpoint();
 
-	public AttachmentData uploadAttachmentToSynapse(String entityId, File temp, String fileName)
```
brucehoff, line 162: this is actually used in the portal

client/synapseJavaClient/src/main/java/org/sagebionetworks/client/SynapseClient.java
```diff
@@ -606,20 +595,6 @@ public Locationable updateExternalLocationableToSynapse(Locationable locationabl
 			String externalUrl, String md5) throws SynapseException;
 
 	@Deprecated
-	public AttachmentData uploadAttachmentToSynapse(String entityId, File dataFile)
-			throws JSONObjectAdapterException, SynapseException, IOException;
-
-	@Deprecated
-	public AttachmentData uploadUserProfileAttachmentToSynapse(String userId,
```
brucehoff, line 615: actually used in the portal

client/synapseJavaClient/src/main/java/org/sagebionetworks/client/SynapseClient.java
```diff
@@ -590,14 +587,6 @@ public File downloadFromSynapse(String path, String md5, File destinationFile)
 			throws SynapseException;
 
 	@Deprecated
-	public Locationable uploadLocationableToSynapse(Locationable locationable,
```
brucehoff, line 595: used by client

PR #1007, opened 2014-01-21
done

john-hill:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/migration/MigratableTableDAOImpl.java
```diff
@@ -186,6 +195,42 @@ private void registerObject(MigratableDatabaseObject dbo, boolean isRoot) {
...
+				});
+		if (names.isEmpty()) {
+			throw new IllegalArgumentException(
+					"BackupId columns must have a uniqueness constraint.  Could not find such a constraint for table: "
+							+ mapping.getTableName()
+							+ " column: "
+							+ backupColumnName);
+		}
+		if (log.isDebugEnabled()) {
+			log.debug("The following uniqueness constraint were found for table: "
```
brucehoff, line 228: constraint(s) were

brucehoff:
Nice job, addressing the root issue as well as fixing the problem!

PR #1011, opened 2014-01-22
Adding value types to the bridge data api

marcel-blonk:


lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/bridge/model/data/value/ParticipantDataLabValue.json
```diff
@@ -0,0 +1,30 @@
...
+    "properties":{
+        "enteredValue":{
+            "type":"string",
+            "description":"the entered value"
+        },
+        "units":{
+            "type":"string",
+            "description":"the units of the entered value"
+        },
+        "normalizedMin":{
```
alxdarksage, line 17: I would call this "normalMin", and "normalMax" or even "labMin" and "labMax" because these are not normalized, they are the medically normal range of values out of a given lab.

lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/bridge/model/data/value/ParticipantDataLabValue.json
```diff
@@ -0,0 +1,30 @@
...
+            "type":"number",
+            "description":"the normalized min desirable value"
+        },
+        "normalizedMax":{
+            "type":"number",
+            "description":"the normalized max desirable value"
+        },
+        "normalizedValue":{
```
alxdarksage, line 25: Normalized to what though? If we're normalizing we may also need "normalizedUnits" as well.

PR #1020, opened 2014-01-24
BridgeClient.deleteParticipantDataRows(...) added.

alxdarksage:


client/bridgeJavaClient/src/main/java/org/sagebionetworks/client/BridgeClientImpl.java
```diff
@@ -25,6 +26,8 @@
  */
 public class BridgeClientImpl extends BaseClientImpl implements BridgeClient {
 
+	private static final String PARTICIPANT_DATA_ROWS = "/rows";
```
marcel-blonk: Group with similar declarations a couple lines below

lib/lib-javaclient/src/main/java/org/sagebionetworks/client/SharedClientConnection.java
```diff
@@ -527,6 +527,26 @@ public JSONObject putJson(String endpoint, String uri, String jsonToPut, String
...
 	/**
+	 * Delete any JSONEntity (for IdList)
+	 * @param endpoint
+	 * @param uri
+	 * @param userAgent 
+	 * @param entity
+	 * @return
+	 * @throws SynapseException
+	 */
+	public void deleteJson(String endpoint, String uri, String jsonToDelete, String userAgent) throws SynapseException {
```
marcel-blonk: This seems a bit confusing. These are low level calls and the name suggests that this should do a DELETE http request. Maybe just call postJson in bridge client, since that is what you're doing at a low level

integration-test/src/test/java/org/sagebionetworks/IT610BridgeData.java
```diff
@@ -299,6 +300,36 @@ public void testGetCurrentParticipantData() throws Exception {
 		assertNull(currentRow.getCurrentData());
 		assertEquals("9", ((ParticipantDataStringValue) currentRow.getPreviousData().getData().get("level")).getValue());
 	}
+	
+	@Test
+	public void testDeleteParticipantDataRows() throws Exception {
```
marcel-blonk, line 305: This needs a test to show that you can delete one or more rows out of many. Delete all is the degenerate case

lib/jdomodels/src/main/java/org/sagebionetworks/bridge/model/dbo/dao/DBOParticipantDataDAOImpl.java
```diff
@@ -121,12 +122,36 @@ public void setS3Bucket(String s3bucket) {
...
+				participantDataDescriptorId).addValue(DBOParticipantData.PARTICIPANT_ID_FIELD, participantId);
+
+		if (rowIds == null || rowIds.getList() == null || rowIds.getList().isEmpty()) {
+			throw new IllegalArgumentException("1-* rows must be selected for deletion");
+		}
+		DBOParticipantData participantData = basicDao.getObjectByPrimaryKeyWithUpdateLock(DBOParticipantData.class, param);
+		DataTable dataTable = getDataFromBucket(participantData.getS3_bucket(), participantData.getS3_key());
+
+		for (Long rowId : rowIds.getList()) {
+			if (!dataTable.rows.containsKey(rowId)) {
```
marcel-blonk: I would be more forgiving here. If the row doesn't exist, consider it deleted.

lib/jdomodels/src/main/java/org/sagebionetworks/bridge/model/dbo/dao/DBOParticipantDataDAOImpl.java
```diff
@@ -121,12 +122,36 @@ public void setS3Bucket(String s3bucket) {
...
+				throw new NotFoundException("Row #" + rowId + " not found");
+			}
+			if (dataTable.rows.get(rowId) == null) {
```
marcel-blonk: Not a necessary check. If it is null (which it can't really be) we still want to remove it.

lib/jdomodels/src/test/java/org/sagebionetworks/bridge/model/dbo/dao/DBOParticipantDataDAOImplTest.java
```diff
@@ -380,4 +381,80 @@ public void testGetSetTypes() throws Exception {
...
+		idList.setList(rowIds);
+		participantDataDAO.deleteRows(participant.getParticipantId().toString(), model.getId().toString(), idList);
+
+		List<ParticipantDataRow> newRowSet = participantDataDAO.get(participant.getParticipantId().toString(), model
+				.getId().toString(), rows.columns);
+
+		assertEquals(1, newRowSet.size());
+		assertEquals("1", ((ParticipantDataStringValue)retainedRow.getData().get("a")).getValue());
+	}
+
```
marcel-blonk, line 499: See also IT test, this or IT test should test deleting a row in the middle.

client/bridgeJavaClient/src/main/java/org/sagebionetworks/client/BridgeClientImpl.java
marcel-blonk: Also, this probably should be /deleteRows, otherwise there is no indication that this REST url will delete things

PR #1021, opened 2014-01-27
Adding encryption for the mapping from user to data

marcel-blonk:


lib/stackConfiguration/src/main/java/org/sagebionetworks/StackConfiguration.java
```diff
@@ -1112,6 +1112,16 @@ public static String getJiraUserPassword() {
 				.getDecryptedProperty("org.sagebionetworks.repo.manager.jira.user.password");
 	}
 
+	public static String getBridgeDataMappingEncryptionKey() {
```
alxdarksage, line 1115: Are these keys supposed to be in memory but not on disk? Can they be loaded through StackConfiguration?

lib/stackConfiguration/src/main/resources/stack.properties
```diff
@@ -155,6 +155,9 @@ org.sagebionetworks.portal.api.getsatisfaction.secret=RmKXqRvGxYeN3yciA1P4TvCJue
 org.sagebionetworks.repo.manager.jira.user.name=synapse-jira-service
 org.sagebionetworks.repo.manager.jira.user.password=6PkUHq9yaYlIqlIZCUn0jmmR4IG0y8b+
 
+org.sagebionetworks.bridge.data.encryptionkey=todo
```
alxdarksage, line 158: Is this alright? I was thinking the keys had to be loaded from Amazon infrastructure and were always in memory.

lib/stackConfiguration/src/main/java/org/sagebionetworks/StackConfiguration.java
marcel-blonk, line 1115: Except for the todo part, this is the mechanism by which StackConfiguration will load the encrypted keys from S3.

services/repository-managers/src/main/java/org/sagebionetworks/bridge/manager/participantdata/ParticipantDataIdMappingManagerImpl.java
```diff
@@ -0,0 +1,70 @@
...
+		return participantIdsForUser;
+	}
+
+	@Override
+	@Transactional(readOnly = false, propagation = Propagation.REQUIRED)
+	public String createNewParticipantForUser(UserInfo alias) throws IOException, GeneralSecurityException {
+		// we need to find a unique id
+		long id;
+		int retry = 0;
+		for (;;) {
```
alxdarksage, line 51: A for loop would be terser. With the break statement, by definition it's only looping if an exception is thrown, each loop is a retry.

services/repository-managers/src/test/java/org/sagebionetworks/bridge/manager/participantdata/ParticipantDataIdMappingManagerImplTest.java
```diff
@@ -0,0 +1,75 @@
...
+import org.mockito.Mock;
+import org.sagebionetworks.bridge.manager.community.MockitoTestBase;
+import org.sagebionetworks.bridge.model.BridgeParticipantDAO;
+import org.sagebionetworks.bridge.model.BridgeUserParticipantMappingDAO;
+import org.sagebionetworks.repo.model.DatastoreException;
+import org.sagebionetworks.repo.model.UserInfo;
+
+import com.google.common.collect.Lists;
+
+public class ParticipantDataIdMappingManagerImplTest extends MockitoTestBase {
```
alxdarksage, line 23: The happy path is only tested a bit in the first method, where the collision leads to an add. Any other happy path tests, like adding the first, or adding several in succession for a given user?

services/repository-managers/src/main/java/org/sagebionetworks/bridge/manager/participantdata/ParticipantDataIdMappingManagerImpl.java
marcel-blonk, line 51: I want it to rethrow the exception it got from the create. Presumably, something besides an id collision is going wrong, so we want that to surface. With a for loop, the same would need to remember the exception and then do a post check to see that there was no successful call and an exception. Makes is as complex it seems?

services/repository-managers/src/test/java/org/sagebionetworks/bridge/manager/participantdata/ParticipantDataIdMappingManagerImplTest.java
marcel-blonk, line 23: That is tested in the IT test, where two datasets are added for the same user. It makes more sense there, as it tests the full stack. Collisions can only be tested with mocking.

services/repository-managers/src/main/java/org/sagebionetworks/bridge/manager/participantdata/ParticipantDataIdMappingManagerImpl.java
alxdarksage, line 51: The difference is slight then, though I find this more readable:

long id;
for (int i=0; i < 10; i++) {
    try {
        id = random.nextLong();
        participantDAO.create(id);
        break;
    } catch(DataStoreException dse) {
        if (i == 10) {
            throw dse;
        }
    }
}

alxdarksage, line 51: Or I would if github preserved the indentation. :-)

lib/stackConfiguration/src/main/resources/stack.properties
marcel-blonk, line 158: this is the key for development. In prod, the key comes from elsewhere.

PR #1031, opened 2014-01-29
Changes for editing/updating PD descriptors and columns

alxdarksage:


lib/jdomodels/src/main/java/org/sagebionetworks/bridge/model/dbo/dao/DBOParticipantDataDescriptorDAOImpl.java
```diff
@@ -129,16 +130,13 @@ public ParticipantDataDescriptor createParticipantDataDescriptor(ParticipantData
 	@Transactional(readOnly = false, propagation = Propagation.REQUIRED)
 	@Override
 	public void updateParticipantDataDescriptor(ParticipantDataDescriptor participantDataDescriptor) throws NotFoundException {
-		DBOParticipantDataDescriptor dboParticipantDataDescriptor = new DBOParticipantDataDescriptor();
```
alxdarksage, line 132: Revert to original implementation

PR #1035, opened 2014-01-30
PLFM-2230,PLFM-1673

brucehoff:


lib/jdomodels/src/main/resources/dao-beans.spb.xml
```diff
@@ -113,7 +113,7 @@
 			<property name="id" value="464532" />
 			<property name="groupAlias">
 			   	<bean class="org.sagebionetworks.repo.model.principal.BootstrapAlias">
-			   		<property name="aliasName" value="Synpase ACT Team" /><!-- This name is was already misspelled -->
+			   		<property name="aliasName" value="Synapse ACT Team" /><!-- This name is was already misspelled -->
```
jay-hodgson, line 116: Why did John write the associated comment, and not simply fix the misspelling when he added this?  Is there more to this story?

john-hill, line 116: We were looking up the team by the name so if it were to be change it would have been broken.

PR #1046, opened 2014-02-05
add explicit data type to distinguish internal data id

marcel-blonk:


lib/jdomodels/src/main/java/org/sagebionetworks/bridge/model/dbo/dao/DBOParticipantDataDAOImpl.java
```diff
@@ -117,9 +119,9 @@ public void setS3Bucket(String s3bucket) {
 		} catch (NotFoundException e) {
 			DBOParticipantData participantData = new DBOParticipantData();
 			participantData.setParticipantDataDescriptorId(Long.parseLong(participantDataDescriptorId));
-			participantData.setParticipantId(Long.parseLong(participantId));
```
alxdarksage, line 120: This seems to be changing from a participant's ID to the ID of a descriptor, is this intentional or was this naming ambiguity?

marcel-blonk, line 120: There was naming ambiguity here, that is now hopefully resolved

PR #1071, opened 2014-02-15
PLFM-2041

brucehoff:


services/repository-managers/src/main/java/org/sagebionetworks/evaluation/manager/EvaluationPermissionsManagerImpl.java
```diff
@@ -154,6 +154,24 @@ public boolean hasAccess(UserInfo userInfo, String evalId, ACCESS_TYPE accessTyp
 		return canAccess(userInfo, evalId, accessType);
 	}
 
+	/**
+	 * Whether the user has the access to the specified evaluation.
+	 * Has the same logic as 'hasAccess' but throws informative exception if the answer is false.
+	 */
+	public void validateHasAccess(UserInfo userInfo, String evalId, ACCESS_TYPE accessType)
```
brucehoff, line 158: rather than duplicating logic, call validateAccess, catch Unauth' and return false.

PR #1082, opened 2014-02-21
new calls for handling events with start and end dates

marcel-blonk:


client/bridgeJavaClient/src/main/java/org/sagebionetworks/client/BridgeClientImpl.java
```diff
@@ -425,4 +455,26 @@ private void delete(String uri) throws SynapseException {
...
+			Object key = keyValues[i];
+			Object value = keyValues[i + 1];
+			if (value != null) {
+				newUri.append(first ? '?' : '&');
+				first = false;
+				newUri.append(key.toString()).append('=').append(value.toString());
+			}
+		}
+		return newUri.toString();
+	}
```
alxdarksage: Library like URIBuilder (Apache http client) would make it simpler to manipulate the parameters (the Apache HTTP client version, not the Java WS version).

marcel-blonk: So done

PR #1092, opened 2014-02-24
Fixing 'no-arg constructor not present' error

alxdarksage:


alxdarksage:
... I hope. We'll see (the class should have a no-arg constructor by default, if I understand correctly, because it has no other constructor. Added it explicitly).

PR #1097, opened 2014-02-26
Plfm 2378

john-hill:


lib/lib-table-query/src/test/java/org/sagebionetworks/table/query/TableQueryParserTest.java
```diff
@@ -0,0 +1,341 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+import java.util.List;
+
+import org.junit.Test;
+import org.sagebionetworks.table.query.model.ColumnReference;
+import org.sagebionetworks.table.query.model.DerivedColumn;
```
john-hill, line 12: add test for key words within strings

PR #1100, opened 2014-02-27
Bridge always send Bridge as domain, and AuthFilter respects it

alxdarksage:


lib/lib-javaclient/src/main/java/org/sagebionetworks/client/SharedClientConnection.java
```diff
@@ -115,6 +116,23 @@ public SharedClientConnection(HttpClientProvider clientProvider) {
 		requestProfile = false;
 	}
 	
+	public SharedClientConnection(DomainType domain) {
+		this(new HttpClientProviderImpl(), domain);
+		this.domain = domain;
```
brucehoff, line 121: unneeded line

PR #1109, opened 2014-03-03
Plfm 2378 b

john-hill:


lib/lib-table-query/src/main/resources/table-query-parser.jj
```diff
@@ -114,27 +114,43 @@ SKIP : /* WHITE SPACE */
...
 | < interval: "INTERVAL">
+| < in: "IN">
+| < is: "IS">
+| < like: "LIKE" >
 | < max: "MAX" >
 | < min: "MIN" >
+| < not: "NOT" >
+| < null_specification: "NULL" >
+| < or: "OR" >
+| < overlaps: "OVERLAPS" >
```
dburdick, line 145: remove overlaps

lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/model/GeneralLiteral.java
```diff
@@ -0,0 +1,5 @@
+package org.sagebionetworks.table.query.model;
```
dburdick, line 1: remove

lib/lib-table-query/src/test/java/org/sagebionetworks/table/query/TableQueryParserTest.java
```diff
@@ -286,56 +532,68 @@ public void testSelectTableNameUnsigned() throws ParseException{
...
+	 * @param element
+	 * @return
+	 */
+	public static String toSQL(SQLElement element){
+		StringBuilder builder = new StringBuilder();
+		element.toSQL(builder);
+		return builder.toString();
+	}
+	
+//	@Test
```
dburdick, line 547: redo as round trip sql

PR #1114, opened 2014-03-05
Unit conversion utils.

alxdarksage:


lib/lib-shared-models/src/test/java/org/sagebionetworks/bridge/model/data/units/UnitsTest.java
```diff
@@ -0,0 +1,109 @@
...
+	}
+	
+	@Test
+	public void convertToNormalizedFromMeasure() {
+		Measure millis = new Measure(1000, Units.MILLILITER);
+		Measure normMillis = millis.convertToNormalized();
+
+		Measure oneLiter = new Measure(1, Units.LITER);
+		
+		assertFalse(millis.equals(normMillis));
```
cbare: If two equivalent amounts in different units are compared, should they be equal? Probably.

lib/lib-shared-models/src/test/java/org/sagebionetworks/bridge/model/data/units/UnitsTest.java
```diff
@@ -0,0 +1,109 @@
...
+	
+	@Test
+	public void conversionFromNormalizedUsingMoles() {
+		// Phosphate	2.5-4.5	mg/dL	0.32	mmol/L
+		// NOTE: This is messed up. Somehow the conversion factor is mixed up with the ratio.
+		// Can mg/dL be varied and how would this work?
+		Measure phosphorus = new Measure(1000, Units.PHOSPHORUS_MG_DL);
+		Measure normPhosphorus = phosphorus.convertToNormalized();
+
+		Measure mmolPhosphorus = new Measure(322.90000000000003, Units.MILLIMOLES_PER_LITER);
```
cbare: May want to compare doubles to within a tolerance? Round-tripping may introduce round-off errors such that converting a quantity x from unit A to unit B and back to A yields a result slightly different from x.

PR #1121, opened 2014-03-07
Plfm 2388

john-hill:


lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/SQLTranslatorUtils.java
```diff
@@ -0,0 +1,854 @@
...
+	
+	/**
+	 * Translate a ValueExpressionPrimary.
+	 * @param valueExpressionPrimary
+	 * @param builder
+	 * @param columnNameToIdMap
+	 * @return
+	 */
+	static boolean translate(ValueExpressionPrimary valueExpressionPrimary,
+			StringBuilder builder, Map<String, Long> columnNameToIdMap) {
```
xschildw, line 139: Should be translate_column (for when on left hand side of a predicate or select)

lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/SQLTranslatorUtils.java
```diff
@@ -0,0 +1,854 @@
...
+		if(betweenPredicate == null) throw new IllegalArgumentException("BetweenPredicate cannot be null");
+		// RHS
+		translate(betweenPredicate.getColumnReferenceLHS(), builder, columnNameToIdMap);
+		if(betweenPredicate.getNot() != null){
+			builder.append(" NOT");
+		}
+		builder.append(" BETWEEN ");
+		translate(betweenPredicate.getBetweenRowValueConstructor(), builder, parameters);
+		builder.append(" AND ");
+		translate(betweenPredicate.getAndRowValueConstructorRHS(), builder, parameters);
```
xschildw, line 700: Remove RHS from fct name

PR #1137, opened 2014-03-12
Plfm 2384 c

john-hill:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/table/TableRowManagerImpl.java
```diff
@@ -201,6 +202,10 @@ public static void validateQuerySize(SqlQuery query, List<ColumnModel> columnMod
 				ColumnModel cm = columIdToModelMap.get(id);
 				seletModels.add(cm);
 			}
+			// First make sure we have a limit
+			if(limit == null){
+				throw new IllegalArgumentException("Request exceed the maximum number of bytes per request because a LIMIT was not included in the query.");
```
john-hill, line 207: Change to when actual columns are selected there must be a limit.

PR #1158, opened 2014-03-21
Plfm 2640

john-hill:


lib/logging/pom.xml
```diff
@@ -34,21 +34,22 @@
...
+			<artifactId>spring-context</artifactId>
 		</dependency>
 		<dependency>
 			<groupId>org.springframework</groupId>
-			<artifactId>org.springframework.spring-library</artifactId>
-			<type>libd</type>
+			<artifactId>spring-context-support</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.aspectj</groupId>
```
eric-wu, line 51: spring-aop? or spring-aspects?

pom.xml
```diff
@@ -203,6 +208,16 @@
 				<version>${org.springframework.version}</version>
 			</dependency>
 			<dependency>
+				<groupId>org.springframework</groupId>
+				<artifactId>spring-context-support</artifactId>
+				<version>${org.springframework.version}</version>
```
eric-wu, line 213: Do we still need explicit dependency on quartz?

PR #1184, opened 2014-04-01
done

john-hill:


lib/lib-javadoc/src/main/resources/SQLSpecification2.csv
```diff
@@ -38,3 +38,5 @@ Pagination,"Pagination is used to limit the number of results returned in a sing
...
 Pagination,"The OFFSET element is optional.  In this example, LIMIT is used to limit the results to the first 5 rows.",select * from syn123 limit 5
 Pagination,Pagination parameters should always be at the end of the query,select * from syn123 where foo =1 group by bar limit 100 offset 0
+Reserved Columns,"Every table has at least two reserved columns ROW_ID and ROW_VERSION.  The values for these column are automatically managed and can not be directly modified.  However, these columns can be used in queries like any other column.  In this example, we are selecting all columns for a single row using its ROW_ID.",select * from syn123 where ROW_ID = 101
+Reserved Columns,"In this example we are listing all row that have a current version number greater than 12.  Note: while each row can have multiple versions, only the current version of each row will appear in the index used to support table queries.  That means it is not possible to list old version of a row or select a row using an old version number.",select * from syn123 where ROW_VERSION > 12
```
marcel-blonk: all rows (plural)

lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/SQLTranslatorUtils.java
```diff
@@ -58,6 +61,12 @@
  *
  */
 public class SQLTranslatorUtils {
+	
+	/**
+	 * The set of reserved column names includes things like ROW_ID and ROW_VERSION
+	 */
+	private static final Set<String> RESERVED_COLUMNS_NAMES = new HashSet<String>(Arrays.asList(ROW_ID, ROW_VERSION));
```
marcel-blonk: This means we must tell user they cannot use these reserved names

PR #1196, opened 2014-04-09
added max batch size for RowSets

john-hill:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/table/TableRowManagerImpl.java
```diff
@@ -76,18 +97,80 @@ public RowReferenceSet appendRows(UserInfo user, String tableId, List<ColumnMode
...
+			List<String> headers, List<Row> batch) throws IOException {
+		// Send the batch
+		RowSet delta = new RowSet();
+		delta.setEtag(etag);
+		delta.setHeaders(headers);
+		delta.setRows(batch);
+		delta.setTableId(tableId);
+		RowReferenceSet rrs = tableRowTruthDao.appendRowSetToTable(user.getId().toString(), tableId, models, delta);
+		if(results != null){
+			results.setEtag(rrs.getEtag());
```
marcel-blonk, line 165: Maybe take these settings and do the in the caller. Makes more sense for this method to just append row references

john-hill, line 165: This method is called in two places; when the batch size is exceeded and at the end if there is anything left in the last batch.  In both cases the logic is the exact same.  If I pull anything out then the caller will be forced to duplicate the moved logic.

lib/stackConfiguration/src/main/resources/stack.properties
```diff
@@ -202,6 +202,9 @@ org.sagebionetworks.doi.enabled=false
 # The maximum number of bytes per table request.  Currently set to 2 MB which matches
 # Tomcat's max bytes per POST
 org.sagebionetworks.table.max.bytes.per.request=2097152
+# The maximum number of rows in a single table change set file.
+org.sagebionetworks.table.max.rows.per.batch=25000
```
brucehoff: Please consider defining the batch size limit in terms of bytes rather than row count.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/table/TableRowManagerImpl.java
```diff
@@ -40,6 +46,8 @@
 
 public class TableRowManagerImpl implements TableRowManager {
 	
+	static private Log log = LogFactory.getLog(RepositoryMessagePublisherImpl.class);
```
brucehoff: wrong class name

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/table/TableRowManagerImpl.java
marcel-blonk, line 165: Not that important. But caller would have to do line 163-168 only once, instead of this method doing it over and over for each row.

john-hill, line 165: I see.

PR #1235, opened 2014-04-23
PLFM-2725

brucehoff:


services/repository/src/test/java/org/sagebionetworks/repo/web/controller/UserProfileControllerAutowiredTest.java
```diff
@@ -183,4 +187,33 @@ public void testFavoriteCRUD() throws Exception {
...
+		{
+			UserPreferenceBoolean pref = new UserPreferenceBoolean();
+			pref.setName("testPref");
+			pref.setValue(true);
+			preferences.add(pref);
+		}
+		ServletTestHelper.updateUserProfile(adminUserId, userProfile);
+		userProfile = ServletTestHelper.getUserProfile(dispatchServlet, adminUserId);
+		boolean foundIt = false;
+		for (UserPreference pref : userProfile.getPreferences()) {
```
jay-hodgson, line 209: ideally would prefer a map (to avoid iterating over any time I'm looking for a key)

PR #1272, opened 2014-05-09
use true and false as boolean values in index tables queries

marcel-blonk:


lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/TableIndexDAOImpl.java
```diff
@@ -187,36 +192,45 @@ public RowSet query(final SqlQuery query) {
...
+				if (!query.isAggregatedResult()) {
+					// Non-aggregate queries include two extra columns,
+					// row id and row version.
+					row.setRowId(rs.getLong(ROW_ID));
+					row.setVersionNumber(rs.getLong(ROW_VERSION));
+				}
+				// fill the value list
+				for (Integer index : nonMetadataColumnIndicies) {
+					String value = rs.getString(index);
+					ColumnModel columnModel = modeledColumns.get(index);
```
john-hill, line 219: move to a util that will generate the return value give a ColumnModel and string value.

PR #1281, opened 2014-05-14
Plfm 2771

brucehoff:


brucehoff:
This was reviewed when pulled into 'release-43'.  The current PR is to get it back into 'develop.'

PR #1292, opened 2014-05-20
Throttle the number of concurrent non-anonymous requests

marcel-blonk:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/semaphore/SemaphoreGatedRunnerImpl.java
```diff
@@ -83,11 +94,23 @@ public void setSemaphoreKey(String semaphoreKey) {
 	 * runners (inclusive) concurrently running across the entire cluster.  Set this to a number less than one to disable this runner.
 	 */
 	public void setMaxNumberRunners(int maxNumberRunners) {
+		this.maxNumberRunners = new StaticPropertyAccessor(maxNumberRunners);
```
marcel-blonk, line 97: Change to immutable

PR #1407, opened 2014-07-10
make index table updates batched

marcel-blonk:


lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/InstanceUtils.java
```diff
@@ -9,7 +9,7 @@
 
 	public static final String DATABASE_INSTANCE_NAME_TEMPALTE = "%1$s-%2$s-table-%3$d";
 	public static final String DATABASE_SCHEMA_NAME_TEMPALTE = "%1$s%2$s";
-	public static final String DATABASE_URL_NAME_TEMPALTE = "jdbc:mysql://%1$s/%2$s";
+	public static final String DATABASE_URL_NAME_TEMPALTE = "jdbc:mysql://%1$s/%2$s?rewriteBatchedStatements=true";
```
john-hill, line 13: What does this do?

marcel-blonk, line 13: It tells the mysql jdbc driver that it can rewrite batches of insert statements to insert multiple rows at a time (rewrite to "insert into x (c) values (:c)" with multiple rows to "insert into x (c) values (:c1) (:c2) (:c3) etc"). Not having that, means that insert statements submitted by batchUpdate will be executed one by one.

PR #1413, opened 2014-07-14
use long poll for table sqs messages

marcel-blonk:


lib/lib-worker/pom.xml
```diff
@@ -71,5 +88,22 @@
 			<artifactId>quartz</artifactId>
 		</dependency>
 
+		<dependency>
+						<groupId>com.google.guava</groupId>
+						<artifactId>guava</artifactId>
+		</dependency>
+		<dependency>
```
john-hill: only need one of these

lib/lib-worker/src/main/java/org/sagebionetworks/asynchronous/workers/sqs/MessagePollingReceiverImpl.java
```diff
@@ -0,0 +1,331 @@
...
+import org.springframework.beans.factory.annotation.Required;
+
+import com.amazonaws.AmazonClientException;
+import com.amazonaws.services.sqs.model.Message;
+import com.google.common.collect.Lists;
+
+/**
+ * A basic implementation of the MessageReceiver.
+ * 
+ * @author John
```
john-hill: derived from MessageReciever authored by mblonk

PR #1424, opened 2014-07-22
handle unavailable table better and easier creation of workers

marcel-blonk:


services/workers/src/test/java/org/sagebionetworks/table/worker/TableCSVDownloadWorkerIntegrationTest.java
```diff
@@ -117,41 +122,16 @@ public void after(){
 				s3Client.deleteObject(fileHandle.getBucketName(), fileHandle.getKey());
 				fileHandleDao.delete(fileHandle.getId());
 			}
+			if (fileHandle2 != null) {
```
john-hill, line 125: remove

PR #1427, opened 2014-07-28
PLFM-2901

toddis:
This was the fix to the issue of allowing an entity in the trash to be restored to a parent that is also in the trash.

Note: A bunch of the changes on "TrashManagerImplAutowiredTest.java" can be ignored, as they're just whitespace stuff that were the result of an accidental "replace all" call I made. Sorry about that.

### DIFF
--- a/services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/trash/TrashManagerImplAutowiredTest.java
+++ b/services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/trash/TrashManagerImplAutowiredTest.java
@@ -910,7 +910,7 @@ public class TrashManagerImplAutowiredTest {
 		assertFalse(nodeDAO.doesNodeExist(KeyFactory.stringToKey(nodeIdC1)));
 	}
 	
-	@Test(expected=EntityInTrashCanException.class)
+	@Test(expected=ParentInTrashCanException.class)
 	public void testRestoreToParentThatIsInTrashCan() throws Exception {
 		// A --> Parent
 		final Node nodeA = new Node();
--- a/services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/trash/TrashManagerImplAutowiredTest.java
+++ b/services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/trash/TrashManagerImplAutowiredTest.java
@@ -526,7 +526,7 @@ public class TrashManagerImplAutowiredTest {
 		//           /  \
 		//          A1  A2
 		//          |    |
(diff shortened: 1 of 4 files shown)


services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/trash/TrashManagerImplAutowiredTest.java
```diff
@@ -943,4 +971,4 @@ private void cleanUp() throws Exception {
 			} catch (NotFoundException e) {}
 		}
 	}
-}
+}
```
eric-wu: Needs a new line

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/trash/TrashManagerImplAutowiredTest.java
```diff
@@ -909,6 +909,34 @@ public void testAdmin() throws Exception {
 		assertTrue(nodeDAO.doesNodeExist(KeyFactory.stringToKey(nodeIdB2)));
 		assertFalse(nodeDAO.doesNodeExist(KeyFactory.stringToKey(nodeIdC1)));
 	}
+	
+	@Test(expected=ParentInTrashCanException.class)
+	public void testRestoreToParentThatIsInTrashCan() throws Exception {
+		// A --> Parent
+		final Node nodeA = new Node();
+		final String nodeNameA = "TrashManagerImplAutowiredTest.testPurge() A";
```
eric-wu: A typo.  Should use "TrashManagerImplAutowiredTest.testRestoreToParentThatIsInTrashCan() A" as the method name.  Node B's name below needs the same fix.

services/repository/src/main/java/org/sagebionetworks/repo/web/controller/BaseController.java
```diff
@@ -693,6 +694,24 @@ ErrorResponse handleEntityInTrashCanException(EntityInTrashCanException ex,
...
+	 *
+	 * @param ex
+	 *            the exception to be handled
+	 * @param request
+	 *            the client request
+	 * @return an ErrorResponse object containing the exception reason or some
+	 *         other human-readable response
+	 */
+	@ExceptionHandler(ParentInTrashCanException.class)
+	@ResponseStatus(HttpStatus.NOT_FOUND)
```
eric-wu: This clearly is a 4xx client error.  But I am not sure if 404 is the best choice here.  You may want to seek a second opinion here (from Jay?)

PR #1462, opened 2014-08-07
Plfm 2927

marcel-blonk:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/semaphore/CountingSemaphoreDaoImpl.java
```diff
@@ -123,33 +145,40 @@ public void releaseLock(String token, String extraKey) {
...
-					return null;
-				}
-
-				DBOCountingSemaphore shared = new DBOCountingSemaphore();
-				shared.setKey(keyName);
-				shared.setToken(UUID.randomUUID().toString());
-				shared.setExpires(System.currentTimeMillis() + lockTimeoutMS);
-				basicDao.createNew(shared);
-				return shared.getToken();
+				return this.transactionTemplate.execute(new TransactionCallback<String>() {
```
brucehoff, line 151: It's not obvious to my (perhaps untrained) eye how this fixes the problem.  Could you add a few words, or even just a reference to PLFM-2927, so that if someone supports this in the future they will have the background?  Thanks.

PR #1464, opened 2014-08-11
allow negative constant values

marcel-blonk:


lib/lib-table-query/src/test/java/org/sagebionetworks/table/query/SignedLiteralTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+	@Test
+	public void testDoubleToSQL() {
+		StringBuilder builder = new StringBuilder();
+		SignedLiteral ai = new SignedLiteral("-123.0", null);
+		ai.toSQL(builder);
+		assertEquals("-123.0", builder.toString());
+	}
+
+	@Test
```
john-hill, line 40: need a test for ".3"

PR #1465, opened 2014-08-11
add ENTITYID as a column type

marcel-blonk:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/persistence/table/ColumnModelUtlis.java
```diff
@@ -170,16 +175,35 @@ public static ColumnModel createNormalizedClone(ColumnModel toClone) {
 					// The max is beyond the allowed size
 					throw new IllegalArgumentException("ColumnModel.maxSize for a STRING must be greater than 0");
 				}
+				break;
+			case ENTITYID:
+			case FILEHANDLEID:
+				if (StringUtils.isEmpty(defaultValue)) {
+					defaultValue = null;
+				}
+				if (defaultValue != null) {
```
brucehoff, line 184: May be useful to document in the user-facing documentation that entity and file handle columns are not allowed default values.

PR #1495, opened 2014-09-09
PLFM-2979

brucehoff:


brucehoff:
Here I am merging a commit into 'develop' that was already merged into release-58.

PR #1501, opened 2014-09-16
emit etag when getting specific rows

marcel-blonk:


lib/lib-utils/src/main/java/org/sagebionetworks/util/ValidateArgument.java
```diff
@@ -0,0 +1,19 @@
...
+
+public class ValidateArgument {
+
+	public static void required(Object fieldValue, String fieldName) {
+		if (fieldValue == null) {
+			throw new IllegalArgumentException(fieldName + " is required.");
+		}
+	}
+
+	public static void notNull(Object fieldValue, String fieldName) {
```
john-hill: remove this one

PR #1562, opened 2014-10-14
Allow deleting in partial updates by passing in null for values

marcel-blonk:


brucehoff:
Please update the javadocs in the controller to explain that a null value means to delete the table row

PR #1571, opened 2014-10-17
put current row cache in the index database

marcel-blonk:


lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/SQLUtils.java
```diff
@@ -42,9 +43,28 @@
 	public static final String TABLE_PREFIX = "T";
 	private static final String COLUMN_PREFIX = "_C";
 	private static final String COLUMN_POSTFIX = "_";
-	public static final String TABLE_STATUS_POSTFIX = "S";
 
 	private static final Pattern COLUMN_NAME_PATTERN = Pattern.compile(COLUMN_PREFIX + "(\\d+)" + COLUMN_POSTFIX);
+
+	public enum TableType {
+		INDEX(""), STATUS("S"), CURRENT_ROW("CR");
```
john-hill, line 50: javadoc on what each are would be helpful

PR #1574, opened 2014-10-24
make count of total result work

marcel-blonk:


lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/model/SQLElement.java
```diff
@@ -7,14 +7,35 @@
  * @author John
  *
  */
-public interface SQLElement {
-	
-	
+public abstract class SQLElement {
+
+	public interface ColumnConvertor {
+		void convertTableName(String tableName, StringBuilder builder);
```
marcel-blonk, line 13: Add comments

lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/model/SetFunctionSpecification.java
```diff
@@ -50,7 +55,9 @@ public void toSQL(StringBuilder builder) {
 				builder.append(setQuantifier.name());
 				builder.append(" ");
 			}
-			this.valueExpression.toSQL(builder);
+			if (this.valueExpression != null) {
```
marcel-blonk, line 58: No longer needed

PR #1585, opened 2014-10-31
handle enum value checks

marcel-blonk:


lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/persistence/table/ColumnModelUtlisTest.java
```diff
@@ -159,60 +160,85 @@ public void testNormalizedDoubleEnum() {
...
 
-		ColumnModelUtlis.createNormalizedClone(original);
+		ColumnModelUtlis.createNormalizedClone(original, StackConfiguration.singleton().getTableMaxEnumValues());
+	}
+
+	@Test(expected = IllegalArgumentException.class)
+	public void testTooManyEnums() {
+		original.setName("name");
+		original.setColumnType(ColumnType.STRING);
+		original.setEnumValues(Collections.nCopies(101, "aaa"));
```
john-hill, line 207: max +1

lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/utils/TableModelUtils.java
```diff
@@ -282,6 +283,17 @@ public static String validateValue(String value, ColumnModel cm) {
 				throw new IllegalArgumentException("String '" + value + "' exceeds the maximum length of " + cm.getMaximumSize()
 						+ " characters. Consider using a FileHandle to store large strings.");
 			}
+			if (cm.getEnumValues() != null) {
+				if (!cm.getEnumValues().contains(value)) {
```
john-hill, line 287: comment as to why a list

PR #1587, opened 2014-11-03
Plfm 3080

brucehoff:


brucehoff:
Merging PLFM-3080 into staging ("release-66")

PR #1613, opened 2014-11-18
generate as clauses only where needed

marcel-blonk:


lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/SQLTranslatorUtils.java
```diff
@@ -71,7 +74,7 @@ public static boolean translate(QuerySpecification model, StringBuilder builder,
 			private ColumnModel columnModelLHS = null;
 			private Set<String> asColumns = Sets.newHashSet();
 			private String currentTableName = null;
-			private SQLClause currentClause = null;
+			private LinkedList<SQLClause> currentClauses = Lists.newLinkedList();
```
john-hill, line 77: easier to read in its own class

PR #1633, opened 2014-12-09
Plfm 3135

xschildw:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/bootstrap/EntityBootstrapperImpl.java
```diff
@@ -65,6 +75,31 @@
...
+					break;
+				}
+				// Sleep
+				TimeUnit.SECONDS.sleep(SLEEP_TIME_S);
+			}
+			
+			doBootstrap();
+
+		} catch (InterruptedException e) {
+			// Should not happen
```
john-hill: Please re-throw this warped in a RuntimeException

PR #1635, opened 2014-12-10
Plfm 1970

brucehoff:


jay-hodgson:
nice first step, stop duplicate approval creation.

PR #1648, opened 2014-12-17
try to use proper extension and content type for export

marcel-blonk:


services/workers/src/main/java/org/sagebionetworks/table/worker/CSVUtils.java
```diff
@@ -142,4 +142,32 @@ public static ColumnModel checkType(String value, ColumnModel currentType){
...
 	}
+
+	/**
+	 * make a rough guess as to what the extension for the file should be based on the separator.
+	 * 
+	 * @param separator
+	 * @return
+	 */
+	public static String guessExtension(String separator) {
+		String extension = "csv"; // by default, just use csv
```
john-hill, line 153: unit test?

PR #1656, opened 2014-12-18
handle deleted columns properly

marcel-blonk:


services/workers/src/test/java/org/sagebionetworks/table/worker/TableWorkerIntegrationTest.java
```diff
@@ -801,6 +801,74 @@ public void testPartialUpdateRoundTrip() throws Exception {
...
+		// wait for table to be available
+		sql = "select * from " + tableId;
+		queryResult = waitForConsistentQuery(adminUserInfo, sql, null, 100L);
+		assertEquals(1, queryResult.getQueryResults().getRows().size());
+		assertEquals(2, queryResult.getQueryResults().getRows().get(0).getValues().size());
+
+		// remove column a
+		schema.remove(0);
+		headers = TableModelUtils.getHeaders(schema);
+		columnManager.bindColumnToObject(adminUserInfo, headers, tableId, true);
```
brucehoff, line 853: "bindColumnToObject" is a misnomer as it can bind or remove multiple columns. Should be "setColumnsForObject" or something like that.

PR #1710, opened 2015-02-07
updating version of schema-to-pojo

brucehoff:


brucehoff:
Reviewed by Jay and previously discussed with John.

PR #1715, opened 2015-02-11
use table index as the latest row cache

marcel-blonk:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/table/TableRowCacheImpl.java
```diff
@@ -31,14 +25,14 @@ public boolean isCurrentVersionCacheEnabled() {
...
+	public boolean isCurrentRowCacheEnabled() {
+		return stackConfiguration.getTableEnabled();
 	}
 
 	@Override
 	public long getLatestCurrentVersionNumber(Long tableId) {
-		CurrentRowCacheDao currentRowCacheDao = connectionFactory.getCurrentRowCacheConnection(tableId);
-		if (!currentRowCacheDao.isEnabled()) {
+		CurrentVersionCacheDao currentRowCacheDao = connectionFactory.getCurrentVersionCacheConnection(tableId);
+		if (!isCurrentVersionCacheEnabled() || !currentRowCacheDao.isEnabled()) {
```
john-hill, line 35: remove both isCacheEnabled()

PR #1720, opened 2015-02-12
PLFM-3174-I

brucehoff:


client/synapseJavaClient/src/main/java/org/sagebionetworks/client/SynapseClient.java
```diff
@@ -1499,11 +1498,15 @@ public String downloadCsvFromTableAsyncStart(String sql, boolean writeHeader, bo
...
+	 * Note: Invalid IDs in the list are ignored:  The results list is simply
+	 * smaller than the set of IDs passed in.
+	 *
+	 * 
 	 * @param ids
 	 * @return
 	 * @throws SynapseException
 	 */
-	public ListWrapper<Team> listTeams(IdList ids) throws SynapseException;
+	public List<Team> listTeams(IdSet ids) throws SynapseException;
```
jay-hodgson, line 1509: Rather see these as standard `Set<Long>` parameters that we convert into an IdSet for the service call.

PR #1756, opened 2015-02-23
Split out ChangeMessage from other kinds of messages

marcel-blonk:


lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/message/ModificationMessage.json
```diff
@@ -0,0 +1,28 @@
...
+		{
+			"$ref": "org.sagebionetworks.repo.model.message.Message"
+		}
+	],
+	"properties": {
+		"userId": {
+			"type": "integer",
+			"description": "The user id for which the change applies"
+		},
+		"timestamp": {
```
john-hill, line 14: timestamp to message and no need for the other two.

lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/message/ChangeMessage.json
```diff
@@ -37,14 +42,6 @@
 				"UPDATE",
 				"DELETE"
 			]
-		},
```
john-hill, line 40: remove object ID and object type and timestamp.

lib/models/src/main/java/org/sagebionetworks/repo/model/message/TransactionalMessenger.java
```diff
@@ -34,9 +35,19 @@
 	public void sendMessageAfterCommit(ChangeMessage message);
```
john-hill, line 36: consider only taking modification and message in this class

PR #1799, opened 2015-03-24
PLFM-3260

kimyen:


client/synapseJavaClient/src/main/java/org/sagebionetworks/client/AsynchJobType.java
```diff
@@ -42,20 +45,58 @@
...
+	private String getEntityIdFromRequest(AsynchronousRequestBody request) {
+		if (request instanceof UploadToTableRequest && ((UploadToTableRequest) request).getTableId() != null) {
+			return ((UploadToTableRequest) request).getTableId();
+		} else if (request instanceof HasEntityId && ((HasEntityId) request).getEntityId() != null) {
+			return ((HasEntityId) request).getEntityId();
+		} else if ((request instanceof UploadToTableRequest && ((UploadToTableRequest) request).getTableId() == null) ||
+					(request instanceof HasEntityId && ((HasEntityId) request).getEntityId() == null)) {
+			throw new IllegalArgumentException("entityId cannot be null");
+		} else {
+			return null;
```
john-hill, line 73: If an IllegalArgumentException is thrown for null at line 71, why return null?

kimyen, line 73: At line 71, one of the following is true:
- the request is an instance of UploadToTableRequest and its tableId field is null; or
- the request is an instance of HasEntityId and its entityId field is null

At line 73, "else" means: the request is not an instance of UploadToTableRequest and also not an instance of HasEntityId. In this case, I do not care about getting the entityId field from this request. So the method returns null.

The way my mind works is using nested if-else. But I heard that nested if-else is bad, so I tried to write it this way. Does it make sense now? Does the comment above the method help at all? How should I write this better?

john-hill, line 73: I see what you are doing and it is fine.
Nested if-else is bad.

PR #1845, opened 2015-04-24
one missed file and make tests go faster

marcel-blonk:


lib/lib-test/src/test/java/org/sagebionetworks/junit/TestParallelizedSpringJUnit4ClassRunner.java
```diff
@@ -0,0 +1,91 @@
...
+	@AfterAll
+	public void afterAll() throws Exception {
+		System.err.println("afterAll" + Thread.currentThread().getId());
+		assertEquals(4 + 3 * 3, steps.incrementAndGet());
+	}
+
+	@AfterClass
+	public static void afterClass() {
+		System.err.println("afterClass" + Thread.currentThread().getId());
+		assertEquals(5 + 3 * 3, steps.incrementAndGet());
```
john-hill, line 54: comment would help

PR #1852, opened 2015-04-30
fixed

john-hill:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/table/TableRowTruthDAOImpl.java
```diff
@@ -67,7 +67,7 @@
  * @author John
  * 
  */
-public class TableRowTruthDAOImpl implements TableRowTruthDAO {
+public abstract class TableRowTruthDAOImpl implements TableRowTruthDAO {
```
marcel-blonk, line 70: I would say that instead of making this abstract and having TableRowTruthDAOImpl no longer be independent of CachingTableRowTruthDAOImpl, checkForRowLevelConflict should probably be pushed down from CachingTableRowTruthDAOImpl to TableRowTruthDAOImpl, it doesn't belong in the caching versions anymore.

john-hill, line 70: The CachingTableRowTruthDAOImpl  uses the cache to optimize the check.  The original version of the method in TableRowTruthDAOImpl would scan through all changes in S3 with a version number greater than the version number of the update.

PR #1873, opened 2015-05-12
needs a client call

john-hill:


services/repository/src/main/java/org/sagebionetworks/file/controller/UploadController.java
```diff
@@ -333,6 +334,34 @@ ExternalFileHandle createExternalFileHandle(
...
+	 * </p>
+	 * 
+	 * @param userId
+	 * @param fileHandle
+	 *            The S3FileHandle to create
+	 * @return
+	 * @throws DatastoreException
+	 * @throws NotFoundException
+	 */
+	@ResponseStatus(HttpStatus.OK)
```
brucehoff, line 353: HttpStatus.CREATED.  John notes that other, similar controllers currently return 200 OK.

PR #1883, opened 2015-05-19
PLFM-3390: update email templates

jay-hodgson:


jay-hodgson:
Do not pull.  It looks like major email clients do not support css link elements in the head...

jay-hodgson:
Converted all to inline css (very basic).  Can be reviewed now...

brucehoff:
Great job!!!!

PR #1920, opened 2015-06-15
allow the headers to be specified separately when uploading a csv

marcel-blonk:


lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/table/UploadToTableRequest.json
```diff
@@ -24,6 +24,13 @@
 		},
 		"csvTableDescriptor": {
 			"$ref":"org.sagebionetworks.repo.model.table.CsvTableDescriptor"
+		},
+		"columnIds": {
+		    "description":"The optional list of ColumnModel IDs that matches the columns in the CSV. If not specified, the CSV columns are taken from the csv (if a header row is given) or assumed to match the columns in the table entity. Order is important. If specified, the number of columns in the CSV must be at less or equal to the number of columns in this list",
+			"type": "array",
```
john-hill, line 30: The columns in the csv must match the order of the array.  Skipped columns must be indicated with a null.

Use ROW_ID and ROW_VERSION to identify them.

PR #1925, opened 2015-06-17
allow ignoring and missing columns

marcel-blonk:


client/synapseJavaClient/src/main/java/org/sagebionetworks/client/SynapseClientImpl.java
```diff
@@ -6243,15 +6243,22 @@ public DownloadFromTableResult downloadCsvFromTableAsyncGet(
...
 			String fileHandleId, String etag, Long linesToSkip,
-			CsvTableDescriptor csvDescriptor) throws SynapseException {
+			CsvTableDescriptor csvDescriptor, List<String> columnIds) throws SynapseException {
 		UploadToTableRequest uploadRequest = new UploadToTableRequest();
 		uploadRequest.setTableId(tableId);
 		uploadRequest.setUploadFileHandleId(fileHandleId);
 		uploadRequest.setUpdateEtag(etag);
 		uploadRequest.setLinesToSkip(linesToSkip);
 		uploadRequest.setCsvTableDescriptor(csvDescriptor);
+		uploadRequest.setColumnIds(null);
```
brucehoff, line 6261: null -> columnIds; need test that fails when this bug is there

PR #1986, opened 2015-07-02
Copy to S3 Snapshot

kimyen:


lib/lib-audit/src/main/java/org/sagebionetworks/audit/utils/BucketDaoProvider.java
```diff
@@ -20,12 +20,14 @@
 	private Map<String, BucketDao> bucketDaoMap;
```
john-hill: keep name map drop dao map. For simplicity.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/asynch/AsynchJobStatusManagerImpl.java
```diff
@@ -173,6 +182,11 @@ public String setComplete(String jobId, AsynchronousResponseBody body)
 			CacheableRequestBody request = (CacheableRequestBody) status.getRequestBody();
 			requestHash = jobHashProvider.getJobHash(request);
 		}
+		// capture the body of the response
+		if (body instanceof S3FileCopyResult || body instanceof S3FileCopyResults) {
```
john-hill: Add interface Snapshotable.
Remove AsynchBody from S3FileCopyResult (it should only be on Results). An S3FileCopyResults should implement Snapshotable.

PR #2006, opened 2015-07-17
better error message

marcel-blonk:


brucehoff:
Thank you for supporting users like me.  :^)

PR #2013, opened 2015-07-21
PLFM-3495

brucehoff:

PR #2014, opened 2015-07-21
Develop

brucehoff:


brucehoff:
This was CR'ed in a PR against relase-99.:#2013
This PR just merges the changes back into develop.

PR #2017, opened 2015-07-21
Plfm 3290 ii

brucehoff:


brucehoff:
This is simply the merge into develop of a fix for PLFM-3290 that was CR'ed against release-99.

PR #2018, opened 2015-07-21
PLFM-3492

kimyen:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/persistence/DBORevision.java
```diff
@@ -199,9 +210,10 @@ public MigrationType getMigratableTableType() {
 			public DBORevision createDatabaseObjectFromBackup(DBORevision backup) {
 				try {
 					// Retrieve the historical reference from a blob
-					Reference ref = DBORevisionUtils.convertBlobToReference(backup.getReference());
+					Reference ref = DBORevisionUtils.convertBlobToReference(backup.getReferences());
 					byte[] blob = JDOSecondaryPropertyUtils.compressReference(ref);
```
john-hill: While this works it means we must change this in release-100 or we will lose all references when we migrate from 99 to 100.  It would be safer to use references only if reference was null.

john-hill:
Looks good!

PR #2088, opened 2015-08-26
Added doi widget with a new mask, and supplemented tests

Njanderson:


lib/lib-shared-models/src/main/java/org/sagebionetworks/repo/model/EntityBundle.java
```diff
@@ -35,6 +36,7 @@
 	public static int TABLE_DATA				= 0x1000;
 	public static int ROOT_WIKI_ID				= 0x2000;
 	public static int BENEFACTOR_ACL			= 0x4000;
+	public static int DOI						= 0x8000;
```
jay-hodgson, line 39: Add to EntityBundleServiceImpl (to return doi in entity bundle if mask part is set), and add test for new mask in EntityBundleServiceImplTest

PR #2091, opened 2015-08-27
Node Record

kimyen:


services/workers/src/test/java/org/sagebionetworks/object/snapshot/worker/utils/NodeObjectRecordBuilderTest.java
```diff
@@ -47,10 +83,22 @@ public void invalidObjectType() {
...
 		ChangeMessage changeMessage = MessageUtils.extractMessageBody(message);
+
+		node.setIsPublic(false);
+		node.setIsControlled(false);
+		node.setIsRestricted(true);
 		ObjectRecord expected = ObjectRecordBuilderUtils.buildObjectRecord(node, timestamp);
+
 		ObjectRecord actual = builder.build(changeMessage);
 		Mockito.verify(mockNodeDAO).getNode(Mockito.eq("123"));
 		assertEquals(expected, actual);
```
john-hill, line 153: There were three new switches added but I only see a test for one of the three.  Expected a test for public true & false, controlled true and false, and restricted true and false.

services/workers/src/main/java/org/sagebionetworks/object/snapshot/worker/utils/NodeObjectRecordBuilder.java
```diff
@@ -32,11 +56,77 @@ public ObjectRecord build(ChangeMessage message) {
...
+		return record;
+	}
+
+	/**
+	 * Build a NodeRecord that wrap around Node object
+	 * 
+	 * @param node
+	 * @return
+	 */
+	public static NodeRecord buildNodeRecord(Node node) {
```
kimyen, line 116: need test

PR #2094, opened 2015-09-02
Plfm 2907

xschildw:


lib/lib-test/pom.xml
```diff
@@ -14,27 +14,34 @@
...
 			<groupId>org.sagebionetworks</groupId>
 			<artifactId>lib-models</artifactId>
 		</dependency>
-				<dependency>
+
+		<dependency>
 			<groupId>junit</groupId>
 			<artifactId>junit</artifactId>
+			<version>4.7</version>
+			<scope>compile</scope>
```
brucehoff, line 37: I think you want:
<scope>test</scope>
for junit

PR #2102, opened 2015-09-09
Plfm 3564

Njanderson:


john-hill:
What are we doing to do with all of the existing DOIs that have a null version number?

Njanderson:
I added another call such that these past null DOIs will still be get-able and create-able.

PR #2114, opened 2015-09-15
Removed a null check to let null replace existing filehandleIds

Njanderson:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/UserProfileUtils.java
```diff
@@ -31,8 +31,11 @@ public static void copyDtoToDbo(UserProfile dto, DBOUserProfile dbo) throws Data
 		} catch (IOException e) {
 			throw new RuntimeException(e);
 		}
-		if(dto.getProfilePicureFileHandleId()!= null){
-			dbo.setPictureId(Long.parseLong(dto.getProfilePicureFileHandleId()));
+		String profilePictureFileHandleId = dto.getProfilePicureFileHandleId();
```
jay-hodgson, line 34: Please add test (verifies that you can now set dbo picture id to null)

PR #2140, opened 2015-10-11
Plfm 3605

Njanderson:


lib/lib-search/src/main/java/org/sagebionetworks/search/SearchDaoImpl.java
```diff
@@ -130,18 +124,26 @@ public void createOrUpdateSearchDocument(List<Document> batch) throws ClientProt
 			for(Document document: documents){
 				prepareDocument(document);
 				serializedDocument = EntityFactory.createJSONStringForEntity(document);
-				// AwesomeSearch pukes on control characters. Some descriptions have
-				// control characters in them for some reason, in any case, just get rid
-				// of all control characters in the search document
-				String cleanedDocument = serializedDocument.replaceAll("\\p{Cc}", "");
```
jay-hodgson, line 164: Would you not want to clean up the AwesomeSearch control characters (and escaped control characters) before submitting to markdown processing?  Could this markdown processing and escaping happen at the same time that the wiki source is added to the search document text?

jay-hodgson, line 164: I suggest cleaning up the garbage before putting it in the index.  I think this would be done in  SearchDocumentDriverImpl.getAllWikiPageText(String nodeId), but you should probably double-check with John that this is the right place to do it.

Njanderson, line 164: I agree with this assessment after looking upstream, but I'll bounce this off of John. Thanks!

services/repository-managers/pom.xml
```diff
@@ -12,6 +12,20 @@
...
+			<!-- jsoup HTML parser library @ http://jsoup.org/ -->
+			<groupId>org.jsoup</groupId>
+			<artifactId>jsoup</artifactId>
+			<version>1.7.3</version>
+		</dependency>
+
+		<dependency>
+			<groupId>org.sagebionetworks</groupId>
+			<artifactId>markdown</artifactId>
+			<version>2015-06-22-6486f39-16</version>
```
john-hill, line 26: Will create a circular dependency?

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/search/SearchDocumentDriverImpl.java
```diff
@@ -423,6 +425,20 @@ public String getAllWikiPageText(String nodeId) throws DatastoreException, IOExc
 					builder.append(page.getTitle());
 				}
 				String markdownString = wikiPageDao.getMarkdown(key, null);
+				try {
+					markdownString = SynapseMarkdownProcessor.getInstance().markdown2Html(markdownString, false, null);
```
john-hill: There are at least three tests missing for this change.

services/repository-managers/pom.xml
Njanderson, line 26: Not sure how to check this unless my build fails when I try a final build before again asking for a code review.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/search/SearchDocumentDriverImpl.java
Njanderson: I believe I added the positive and negative tests in question.

services/repository-managers/pom.xml
Njanderson, line 26: My build completed successfully.

Njanderson, line 26: The dependencies of the Synapse Markdown dependency are:
  <dependencies>
    <dependency>
        <groupId>junit</groupId>
        <version>4.7</version>
        <scope>test</scope>
    </dependency>
    <dependency>
        
        <groupId>org.jsoup</groupId>
        <artifactId>jsoup</artifactId>
        <version>1.7.3</version>
    </dependency>
    <dependency>
        <groupId>commons-lang</groupId>
        <artifactId>commons-lang</artifactId>
        <version>2.6</version>
    </dependency>
  </dependencies>

PR #2148, opened 2015-10-21
Plfm 3482

john-hill:


lib/lib-search/src/main/resources/private/search-trigger.spb.xml
```diff
@@ -1,43 +0,0 @@
...
-
-	<!-- Turn on Spring's autoproxy using AspectJ's @Aspect annotations. -->
-	<aop:aspectj-autoproxy />
-
-	<!-- initializer worker for creating the cloud search -->
-	<bean id="cloudSearchInitializerRunner"
-		class="org.sagebionetworks.repo.init.DelayedInitializer" scope="singleton">
-		<property name="targetObject" ref="searchDao" />
-		<property name="targetMethod" value="postInitialize" />
-		<property name="semaphore">
```
xschildw, line 19: We need to initialize at least once.

PR #2221, opened 2015-12-02
Thread APIs

kimyen:


services/repository/src/main/java/org/sagebionetworks/repo/web/controller/DiscussionController.java
```diff
@@ -53,4 +63,107 @@
 			@PathVariable String projectId) throws DatastoreException, NotFoundException {
 		return serviceProvider.getDiscussionService().getForumMetadata(userId, projectId);
 	}
+
+	/**
+	 * Get limit number of threads starting at offset for a given forum
+	 * 
+	 * @param userId
```
kimyen: Who is allowed to use this API?
describe what each param means

PR #2242, opened 2015-12-16
Deprecated APIs

kimyen:


services/repository/src/main/java/org/sagebionetworks/repo/web/controller/EvaluationController.java
```diff
@@ -856,7 +738,7 @@ BatchUploadResponse updateSubmissionStatusBatch(
 	
 	/**	
 	 * Deletes a Submission and its accompanying SubmissionStatus.
-	 * Use of this service is discouraged, since Submissions should be immutable.
+	 * <b>This service is intended to only be used by ChallengesInfrastructure service account.</b>
```
brucehoff, line 741: "this service is only used for integration testing and should not be used in production."

PR #2248, opened 2015-12-18
added memory counting semaphore

john-hill:


lib/models/src/main/java/org/sagebionetworks/repo/model/semaphore/Token.java
```diff
@@ -0,0 +1,57 @@
+package org.sagebionetworks.repo.model.semaphore;
+
+/**
+ * Token used to track a semaphore lock.
+ *
+ */
+public class Token {
```
john-hill: Call this Lock.

brucehoff: maybe call this class "Lock".

PR #2304, opened 2016-02-03
added create ProxyFileHandle and integration test

john-hill:


services/repository/src/main/java/org/sagebionetworks/file/controller/UploadController.java
```diff
@@ -402,6 +403,33 @@ void uploadFiles(
 		// Pass it along
 		return fileService.createExternalS3FileHandle(userId, fileHandle);
 	}
+	
+	/**
+	 * Create an ProxyFileHandle to represent an File in a user's file
```
jay-hodgson, line 408: *a

PR #2314, opened 2016-02-06
Don't truncate the date.

xschildw:


lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/file/MultipartUploadDAOImplTest.java
```diff
@@ -311,4 +313,15 @@ public void testSetUploadComplete(){
...
 	}
 	
+	@Test
+	public void testPLFM3737() {
+		CompositeMultipartUploadStatus composite = multipartUplaodDAO.createUploadStatus(createRequest);
+		assertNotNull(composite);
+		Date d = composite.getMultipartUploadStatus().getStartedOn();
+		Calendar c = Calendar.getInstance();
+		c.setTime(d);
+		// Test will fail if run at 12:00:00am
```
xschildw: Ignore if test ran between 12:00am and 1:00am, wait if min or sec == 0.

PR #2324, opened 2016-02-12
SWC-2896

kimyen:
return an empty array instead of fill in the author


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/discussion/DBODiscussionThreadDAOImpl.java
```diff
@@ -84,7 +83,7 @@ public DiscussionThreadBundle mapRow(ResultSet rs, int rowNum)
 			}
 			String listString = rs.getString(COL_DISCUSSION_THREAD_STATS_ACTIVE_AUTHORS);
 			if (rs.wasNull()) {
-				dto.setActiveAuthors(Arrays.asList(dto.getCreatedBy()));
+				dto.setActiveAuthors(new ArrayList<String>());
 			} else {
```
john-hill, line 87: initialize array with zero or linked list

PR #2331, opened 2016-02-20
Plfm 3605 revisit

Njanderson:


jay-hodgson:
Deprecating the Synapse MD processor in favor of integrating with a client-side (js) extendable library (markdown-it).  Could look into including a node.js module in the backend, but I think Synapse search requires a bigger overhaul.

PR #2341, opened 2016-03-09
Auto subscribe

kimyen:
When a user creates a thread, this user is automatically subscribe to the thread.
When a user creates a reply, this user is automatically subscribe to the reply's thread.

Use the combination SUBSCRIBER_ID, OBJECT_ID, OBJECT_TYPE as primary key. This allows the idempotent behavior for subscribing to the same object multiple times. 

Refactor get list to only take one json object instead of taking json object and pass var through query param.


lib/jdomodels/src/main/resources/schema/Subscription-ddl.sql
```diff
@@ -4,7 +4,7 @@ CREATE TABLE IF NOT EXISTS `SUBSCRIPTION` (
   `OBJECT_ID` bigint(20) NOT NULL,
   `OBJECT_TYPE` ENUM('FORUM', 'DISCUSSION_THREAD') NOT NULL,
   `CREATED_ON` bigint(20) NOT NULL,
-  PRIMARY KEY (`ID`),
-  UNIQUE (`SUBSCRIBER_ID`, `OBJECT_ID`, `OBJECT_TYPE`),
+  PRIMARY KEY (`SUBSCRIBER_ID`, `OBJECT_ID`, `OBJECT_TYPE`),
+  UNIQUE (`ID`),
   CONSTRAINT `DISCUSSION_THREAD_SUBSCRIBER_ID_FK` FOREIGN KEY (`SUBSCRIBER_ID`) REFERENCES `JDOUSERGROUP` (`ID`)
 )
```
john-hill, line 10: ON DELETE CASCADE,

PR #2363, opened 2016-04-05
Check

xschildw:


lib/lib-search/src/main/java/org/sagebionetworks/search/SearchDaoImpl.java
```diff
@@ -301,8 +295,8 @@ public boolean isSearchEnabled() {
 
 	private CloudSearchClient validateSearchAvailable() throws ServiceUnavailableException {
 		validateSearchEnabled();
-		if (cloudHttpClient == null) {
-			throw new ServiceUnavailableException("Search service still initializing...");
+		if ((cloudHttpClient.getSearchServiceEndpoint() == null) || (cloudHttpClient.getDocumentServiceEndpoint() == null)) {
+			throw new ServiceUnavailableException("Search service not initialized...");
 		}
```
john-hill, line 310: I do not understand how this helps.  The method validateSearchAvailable() is called before each attempt to run a search.  So if the endpoints are null why not attempt to look them up and only throw ServiceUnavailableException only if the endpoints are unavailable.

PR #2375, opened 2016-04-07
PLFM-3812

kimyen:
filter out subscriptions in projects that user can no longer read


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/subscription/DBOSubscriptionDAOImpl.java
```diff
@@ -351,4 +375,25 @@ public void deleteList(String userId, List<String> idList, SubscriptionObjectTyp
...
+
+	public String getProjectQuery(SubscriptionObjectType objectType) {
+		ValidateArgument.required(objectType, "objectType");
+		switch (objectType) {
+			case FORUM:
+				return SQL_GET_PROJECT_FORUM_SUB;
+			case THREAD:
+				return SQL_GET_PROJECT_THREAD_SUB;
+			default:
+				throw new RuntimeException("Unsopported type "+objectType.name());
```
kimyen, line 396: Unsopported => Unsupported

PR #2380, opened 2016-04-12
PLFM-3818

kimyen:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/AuthenticationManagerImpl.java
```diff
@@ -18,33 +20,37 @@
...
+			// This will throw an UnauthorizedException if invalid
+			if (password != null) {
+				byte[] salt = authDAO.getPasswordSalt(principalId);
+				String passHash = PBKDF2Utils.hashPassword(password, salt);
+				authDAO.checkUserCredentials(principalId, passHash);
+			}
+			usernameThrottleGate.releaseLock(""+principalId, lockToken);
+			return getSessionToken(principalId, domain);
+		} else {
+			throw new RuntimeException(AuthorizationConstants.REASON_TOO_MANY_CONCURRENT_REQUESTS);
```
john-hill: Why throw a RuntimeException?  This error should map to 429 Too Many Requests.  It looks like 429 is currently mapped to TooManyRequestsException.

PR #2409, opened 2016-04-30
File view

john-hill:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/table/FileViewDaoImpl.java
```diff
@@ -0,0 +1,96 @@
...
+			+ ", '-',"
+			+ COL_NODE_ETAG
+			+ "))) FROM "
+			+ TABLE_NODE
+			+ " WHERE "
+			+ COL_NODE_TYPE
+			+ " = '"
+			+ EntityType.file
+			+ "' AND "
+			+ COL_NODE_PARENT_ID + " IN (:" + IDS_PARAM_NAME + ")";
```
kimyen, line 43: format

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/table/FileViewUtils.java
```diff
@@ -0,0 +1,290 @@
...
+ * Static utilities for FileViews.
+ *
+ */
+public class FileViewUtils {
+	
+	private static final String IDS_PARAM_NAME = "ids_param";
+
+	/**
+	 * Create the SQL need to build a view using the given column models.
+	 *
```
kimyen: How the SQl will look like with parts that are required and parts that are optional

SELECT N.ID, N.CURRENT_VERSION <, OPTIONAL PRIMARY COLUMNS><, OPTIONAL ANNOTATION COLUMN>
FROM NODE N <, REVISION R>
WHERE N.TYPE = 'file'
AND N.PARENT_ID IN (<a given set of containers>)
<AND N.ID = R.OWNER_ID AND N.CURRENT_VERSION = R.VERSION_NUMBER>;

PR #2410, opened 2016-04-30
PLFM-3841

kimyen:


lib/lib-markdown/src/main/java/org/sagebionetworks/markdown/MarkdownClient.java
```diff
@@ -0,0 +1,73 @@
...
+		if (markdownHttpClientProvider == null) {
+			throw new RuntimeException("MarkdownHttpClientProvider is null in MarkdownClient._init()");
+		}
+		this.httpClient = markdownHttpClientProvider.getHttpClient();
+	}
+
+	/**
+	 * Takes a json string requestContent (ex. {"markdown":"## a heading"})
+	 * Makes a call to the markdown server to convert the raw markdown to html
+	 * Return the json string representation of the response (ex. {"html":"<h2 toc=\"true\">a heading</h2>\n"})
```
kimyen: {"result":"<h2 toc=\"true\">a heading</h2>\n"}

lib/lib-markdown/src/main/java/org/sagebionetworks/markdown/MarkdownClient.java
```diff
@@ -0,0 +1,73 @@
...
+	 * @throws IOException
+	 */
+	public String requestMarkdownConversion(String requestContent) throws ParseException, IOException{
+		String url = markdownServiceEndpoint+MARKDOWN_TO_HTML;
+		HttpResponse response = HttpClientHelper.performRequest(httpClient, url, "POST", requestContent, DEFAULT_REQUEST_HEADERS);
+		int statusCode = response.getStatusLine().getStatusCode();
+		if (statusCode == 200) {
+			HttpEntity responseEntity = response.getEntity();
+			if (responseEntity != null) {
+				return EntityUtils.toString(responseEntity);
```
kimyen: EntityUtils.toString(responseEntity); should be called unconditionally before checking the status

lib/lib-markdown/src/main/java/org/sagebionetworks/markdown/MarkdownClient.java
```diff
@@ -0,0 +1,73 @@
...
+			}
+		} else {
+			log.info("Fail to request markdown conversion for request: "+requestContent+". Response: "+response.toString());
```
kimyen: throw an exception, let the client of this deal with the exception
exception include status code & message

PR #2488, opened 2016-07-12
make docLink and helpText visible to everyone

kimyen:


jay-hodgson:
Could add a test, like UserProfileManagerUtilsTest for user profile private fields.

PR #2489, opened 2016-07-13
PLFM-3909

kimyen:


services/repository/src/main/java/org/sagebionetworks/repo/web/filter/DockerFilter.java
```diff
@@ -0,0 +1,65 @@
...
+		if (up != null) {
+			try {
+				LoginCredentials credential = new LoginCredentials();
+				credential.setEmail(up.getUserName());
+				credential.setPassword(up.getPassword());
+				authenticationService.authenticate(credential, DomainType.SYNAPSE);
+				PrincipalAlias alias = authenticationService.lookupUserForAuthentication(up.getUserName());
+				userId = alias.getPrincipalId();
+			} catch (NotFoundException e) {
+			}
```
brucehoff, line 50: if returning 401 status works with the docker client, then please do so here.

services/repository/src/main/webapp/WEB-INF/web.xml
```diff
@@ -99,6 +99,16 @@
 	</filter-mapping>
 
 	<filter>
+		<filter-name>dockerFilter</filter-name>
+		<filter-class>org.sagebionetworks.repo.web.filter.DockerFilter</filter-class>
```
kimyen, line 103: need to use DelegatingFilterProxy and create bean for dockerFilter

PR #2516, opened 2016-07-26
Plfm 3708

zimingd:

PR #2517, opened 2016-07-26
PLFM-3708

zimingd:
fixed issues from merge request #2516 
Implemented purging trash leaves for java client


integration-test/src/test/java/org/sagebionetworks/IT070SynapseJavaClientTrashCanTest.java
```diff
@@ -201,4 +201,39 @@ public void testAdmin() throws SynapseException {
...
+		assertEquals(1, results.getResults().size());//only 1 item in trash
+		assertEquals(parent.getId(), results.getResults().get(0).getEntityId());//id of trash item matches parent node's id
+		
+		//purge leaves again to empty trash can
+		adminSynapse.purgeTrashLeaves(daysInTrash, limit);
+		results = adminSynapse.viewTrash(offset, limit);
+		assertNotNull(results);
+		assertEquals(0, results.getResults().size());//nothing in trash
+		
+		child = null;
```
kimyen, line 236: use try catch in after to clean up in case other dev forget to set the value to null

PR #2525, opened 2016-07-28
Plfm 3981

zimingd:


services/repository/src/main/java/org/sagebionetworks/repo/web/filter/UserThrottleFilter.java
```diff
@@ -30,17 +30,25 @@
...
 	
-	public static final long LOCK_TIMOUTE_SEC = 60*10; // 10 MINS
+	public static final long CONCURRENT_CONNECTIONS_LOCK_TIMEOUT_SEC = 60*10; // 10 MINS
 	// The maximum number of concurrent locks a user can have per machine.
 	public static final int MAX_CONCURRENT_LOCKS = 3;
+	
+	//limit users to 1 request per 1 second
+	public static final long REQUEST_FREQUENCY_LOCK_TIMEOUT_SEC =  1; //1 second
+	public static final int MAX_REQUEST_FREQUENCY_LOCKS = 1;
+
```
john-hill, line 46: I think we need to be more tolerant of bursts of activity.  For example, if we allow 30 locks each with a timeout of 30 seconds we still have an average rate of one request per second but also allow  a quick burst of requests without throttling.

PR #2531, opened 2016-08-08
PLFM-3830

kimyen:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/discussion/DiscussionUtils.java
```diff
@@ -1,15 +1,26 @@
...
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import org.sagebionetworks.repo.model.discussion.DiscussionThreadEntityReference;
+import org.sagebionetworks.repo.model.jdo.KeyFactory;
 import org.sagebionetworks.util.ValidateArgument;
 
 public class DiscussionUtils {
 	private static final Pattern USER_MENTION_PATTERN = Pattern.compile("@\\S+");
+	private static final Pattern ENTITY_REF_PATTERN = Pattern.compile("syn\\d+(\\s|$)");
```
john-hill, line 16: synch with Jay's, do we allow case?

PR #2564, opened 2016-08-16
use the enum that is defined in lib-autogenerated

kimyen:


brucehoff:
Looks great, thanks!!

PR #2577, opened 2016-08-25
Entity view b

john-hill:


lib/lib-table-cluster/src/test/java/org/sagebionetworks/table/cluster/TableIndexDAOImplTest.java
```diff
@@ -48,6 +52,7 @@
 
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
+import com.sun.tools.internal.ws.wsdl.framework.Entity;
```
john-hill, line 56: remove

services/workers/src/main/java/org/sagebionetworks/table/worker/EntityReplicationWorker.java
```diff
@@ -0,0 +1,103 @@
...
+
+	@Override
+	public void run(ProgressCallback<Void> progressCallback,
+			List<ChangeMessage> messages) throws RecoverableMessageException,
+			Exception {
+
+		// batch the create/update events and delete events
+		List<String> createOrUpdateIds = new LinkedList<>();
+		List<String> deleteIds = new LinkedList<String>();
+		groupByChangeType(messages, createOrUpdateIds, deleteIds);
```
john-hill, line 49: did not use deletedIds

PR #2578, opened 2016-08-27
Plfm-4005

zimingd:


lib/models/src/main/java/org/sagebionetworks/repo/model/throttle/ThrottleRule.java
```diff
@@ -0,0 +1,138 @@
...
+		this.maxCallsPerPeriod = maxCallsPerPeriod;
+	}
+
+	public void setPeriod(long periodInSeconds) {
+		ValidateArgument.requirement(periodInSeconds >= 0, "periodInSeconds must be a positive value");
+		this.periodInSeconds = periodInSeconds;
+	}
+
+    @Override
+    public int hashCode() {
```
john-hill, line 87: let eclipse build hasCode() equals() and toString()

PR #2593, opened 2016-09-09
Plfm 4031

john-hill:


services/workers/src/test/java/org/sagebionetworks/table/worker/TableViewIntegrationTest.java
```diff
@@ -270,6 +282,70 @@ public void testContentUpdate() throws Exception {
...
+		updates.add(request);
+		TableUpdateTransactionRequest transaction = new TableUpdateTransactionRequest();
+		transaction.setEntityId(fileViewId);
+		transaction.setChanges(updates);
+	
+		// wait for the change to complete
+		startAndWaitForJob(adminUserInfo, transaction, TableUpdateTransactionResponse.class);
+		// run the query again
+		String sql = "select etag from "+fileViewId+" where id = "+fileId;
+		QueryResultBundle results = waitForConsistentQuery(adminUserInfo, sql);
```
xschildw, line 311: Check that result does not include benefactor for select *.

PR #2622, opened 2016-09-23
Plfm 4059

brucehoff:


brucehoff:
This PR is a remerge of release-152 back in to develop

PR #2670, opened 2016-10-28
remove lib-concept and its dependency

kimyen:


kimyen:
This saves 7 minutes for my local build!

PR #2682, opened 2016-11-02
PLFM-4116 - update HttpClient version and AWS Java SDK version

kimyen:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/migration/ByteSizeUtils.java
```diff
@@ -39,6 +39,8 @@ public static int estimateSizeInBytes(BeanPropertySqlParameterSource object){
 					size += Integer.SIZE;
 				}else if(Types.INTEGER == type){
 					size += Integer.SIZE;
+				}else if(Types.BOOLEAN == type){
```
john-hill, line 42: In dead code.

services/repository-managers/pom.xml
```diff
@@ -151,7 +151,7 @@
 		<dependency>
 			<groupId>org.apache.httpcomponents</groupId>
 			<artifactId>httpcore</artifactId>
-			<version>4.2.3</version>
+			<version>4.4.4</version>
```
john-hill, line 154: all versions must be in the root pom.xml

PR #2706, opened 2016-11-15
PLFM-4150

kimyen:


services/repository/src/main/java/org/sagebionetworks/repo/web/controller/AccessApprovalController.java
```diff
@@ -159,10 +160,27 @@ AccessApproval createAccessApproval(
...
+	/**
+	 * Delete Access Approval. This service is only available to the ACT.
+	 * @param userId - The user who is making the request
+	 * @param accessRequirementId - The access requirement to look for
+	 * @param accessorId - The user whose access is being revoked
+	 * @throws UnauthorizedException
+	 * @throws NotFoundException
+	 */
+	@ResponseStatus(HttpStatus.OK)
+	@RequestMapping(value = UrlHelpers.ACCESS_APPROVAL_ACCESS_REQUIREMENT_ID_ACCESSOR_ID, method = RequestMethod.DELETE)
```
brucehoff: Do you feel this is more 'RESTful' than passing the requirement ID and accessor ID as query parameters:
DELETE /accessApproval?requirementId=xxxxx&accessorId=xxxx
?

If I see
/accessApproval/xxxx I would think that xxx is the ID of the access approval object.

PR #2719, opened 2016-11-19
convert markdownClient to use simpleHttpClient

kimyen:


lib/lib-markdown/src/test/java/org/sagebionetworks/markdown/MarkdownDaoImplTest.java
```diff
@@ -28,11 +28,11 @@ public void testConvertMarkdownWithNullMarkdown() throws Exception {
 		dao.convertMarkdown(null, null);
 	}
 
-	@Test (expected = HttpClientHelperException.class)
+	@Test (expected = HttpResponseException.class)
 	public void testConvertMarkdownWithNullResponse() throws Exception {
```
xschildw, line 31: MarkdownClientException

PR #2721, opened 2016-11-22
convert lib-doi to use SimpleHttpClient

kimyen:


lib/lib-doi/src/main/java/org/sagebionetworks/doi/DxAsyncClient.java
```diff
@@ -1,37 +1,38 @@
...
+	public static final String URL_PARAM = "?type=URL";
+	private static final String VALUES = "values";
+	private static final String DATA = "data";
+	private static final String VALUE = "value";
+
 	public DxAsyncClient() {
-		HttpClient httpClient = new DefaultHttpClient();
-		httpClient.getParams().setBooleanParameter(ClientPNames.HANDLE_REDIRECTS, false);
-		this.httpClient = new RetryableHttpClient(httpClient);
+		this.simpleHttpClient = new SimpleHttpClientImpl(null);
```
kimyen, line 29: add a new constructor to SimpleHttpClientImpl that does not take an argument

PR #2724, opened 2016-11-28
convert lib-search to use SimpleHttpClient

kimyen:


lib/lib-search/src/main/java/org/sagebionetworks/search/CloudSearchClient.java
```diff
@@ -83,46 +67,44 @@ public String getDocumentServiceEndpoint() {
...
-					// rethrow
-					logger.error("performSearch(): Exception rethrown (url="+(url==null?"null":url)+")");
-					throw(e1);
+			response = httpClient.get(request);
+			if (response.getStatusCode() == 200) {
+				return response.getContent();
+			}
+			if (response.getStatusCode() == 507) {
+				try {
+					Thread.sleep(backoffMs);
```
kimyen, line 90: do not silently catch an interrupted exception.

lib/lib-search/src/main/java/org/sagebionetworks/search/CloudSearchClient.java
```diff
@@ -83,46 +67,44 @@ public String getDocumentServiceEndpoint() {
...
+						+ " Reason: "+response.getStatusReason()+". Content: "+response.getContent()));
 			}
 			backoffMs *= 2;
-		} while ((s == null) && (backoffMs < MAX_BACKOFF_MS));
+		} while (backoffMs < MAX_BACKOFF_MS);
 		// If we're past the max backoff, throw the last 507 we got
 		if (backoffMs >= MAX_BACKOFF_MS) {
-			logger.error("performSearch(): Backoff exceeded (url="+(url==null?"null":url)+")");
-			throw(e);
+			logger.error("performSearch(): Backoff exceeded (url="+(uri==null?"null":uri)+")");
```
kimyen, line 104: (backoffMs >= MAX_BACKOFF_MS) is always true. remove if & return

services/repository/src/main/java/org/sagebionetworks/repo/web/controller/SearchController.java
```diff
@@ -50,7 +48,7 @@
 	 * @return search results from CloudSearch
 	 * @throws ClientProtocolException
 	 * @throws IOException
-	 * @throws HttpClientHelperException
+	 * @throws CloudSearchClientException
```
kimyen, line 51: in BaseController, map CloudSearchClientException to BAD_REQUEST

PR #2726, opened 2016-11-29
R159-PLFM-3938

xschildw:
Backport async migration to release-159


services/repository/src/test/java/org/sagebionetworks/repo/web/controller/ServletTestHelper.java
```diff
@@ -1432,7 +1432,38 @@ public AsynchronousJobStatus getAsynchJobStatus(DispatcherServlet instance, Long
 			return null;
 		}
 	}
+
```
xschildw, line 1435: Check if used.

PR #2759, opened 2016-12-08
R160 PLFM-4159

xschildw:
Cherry-picked from John's fix in release-161.


john-hill:
There are two parts to this fix.  This only includes the second.  I will setup a PR for this first.

PR #2767, opened 2016-12-10
pass the modifiable map as headers

kimyen:


kimyen:
Xa checked in the fix in a different PR.

PR #2775, opened 2016-12-13
refactor

kimyen:


client/synapseJavaClient/src/main/java/org/sagebionetworks/client/SynapseClient.java
```diff
@@ -820,10 +803,7 @@ public void updateMessageStatus(MessageStatus status)
 
 	public Long getChildCount(String entityId) throws SynapseException;
 
-	public SynapseVersionInfo getVersionInfo() throws SynapseException,
```
kimyen, line 823: Is this method already exist in BaseClient

client/synapseJavaClient/src/main/java/org/sagebionetworks/client/SynapseClientImpl.java
```diff
@@ -236,9 +229,6 @@
 
 	public static final String APPLICATION_OCTET_STREAM = "application/octet-stream";
 
-	private static final Logger log = LogManager
```
kimyen, line 239: Check to see if we can remove Logging dependency

client/synapseJavaClient/src/main/java/org/sagebionetworks/client/SynapseClient.java
kimyen: This wasn't been removed.

PR #2777, opened 2016-12-14
PLFM-4187

zimingd:


lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/util/TableSqlProcessor.java
```diff
@@ -180,5 +188,129 @@ private static OrderingSpecification getCurrentDirection(OrderByClause obc,
...
+		}
+		builder.append(")");
+		return builder.toString();
+	}
+	
+	/**
+	 * Appends a value to the string builder
+	 * and places single quotes (') around it if the string contains spaces
+	 */ 
+	private static void appendValueToStringBuilder(StringBuilder builder, String value){
```
john-hill: This will not enough.  Spaces are just one of many characters that would required quotes. If the ColumnType is STRING then the value needs to be in quotes.

lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/util/TableSqlProcessor.java
```diff
@@ -180,5 +188,129 @@ private static OrderingSpecification getCurrentDirection(OrderByClause obc,
...
+			return createEnumerationSearchCondition((FacetColumnValuesRequest) facetColumnRequest);
+		}else if (facetColumnRequest instanceof FacetColumnRangeRequest){
+			return createRangeSearchCondition((FacetColumnRangeRequest) facetColumnRequest);
+		}else{
+			throw new IllegalArgumentException("Unexpected instance of FacetColumnRequest");
+		}
+		
+	}
+	
+	private static String createRangeSearchCondition(FacetColumnRangeRequest facetRange){
```
john-hill: This method should be protected and have its own unit tests.

lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/util/TableSqlProcessor.java
```diff
@@ -180,5 +188,129 @@ private static OrderingSpecification getCurrentDirection(OrderByClause obc,
...
+			appendValueToStringBuilder(builder, min);
+			builder.append(" AND ");
+			appendValueToStringBuilder(builder, max);
+		}
+		
+		builder.append(")");
+		return builder.toString();
+	}
+	
+	private static String createEnumerationSearchCondition(FacetColumnValuesRequest facetValues){
```
john-hill: This method should be protected and have its own unit tests.

PR #2790, opened 2017-01-11
add admin service to update file

kimyen:


services/repository/src/main/java/org/sagebionetworks/repo/web/service/AdministrationServiceImpl.java
```diff
@@ -105,35 +119,14 @@
...
-			DBOChangeDAO changeDAO) {
-		super();
-		this.backupDaemonLauncher = backupDaemonLauncher;
-		this.objectTypeSerializer = objectTypeSerializer;
-		this.userManager = userManager;
-		this.stackStatusManager = stackStatusManager;
-		this.messageSyndication = messageSyndication;
-		this.changeDAO = changeDAO;
-	}
+	@Autowired
```
john-hill, line 122: Need to revert this pull request when we are done with this process.  Need to ensure these dependencies are removed from the services layer.

john-hill:
After the changes go live, we will need a check that throws 400 if the 'fileNameOverride' is not null on entity update or create.

PR #2791, opened 2017-01-11
Migration type names support.

xschildw:
Return list of type names instead of types.

### DIFF
--- a/lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/migration/MigrationTypeNames.json
+++ b/lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/migration/MigrationTypeNames.json
@@ -1,0 +1,12 @@
+{
+  "description": "List of MigrationType names",
+  "properties": {
+    "list": {
+      "type": "array",
+      "uniqueItems":false,
+      "items": {
+        "type":"string"
+      }
+    }
+  }
+}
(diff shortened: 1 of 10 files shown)


lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/migration/MigrationTypeNames.json
```diff
@@ -0,0 +1,12 @@
+{
+  "description": "List of MigrationType names",
+  "properties": {
+    "list": {
+      "type": "array",
+      "uniqueItems":false,
```
xschildw, line 6: That's the default.

services/repository/src/main/java/org/sagebionetworks/repo/web/controller/MigrationController.java
```diff
@@ -284,7 +300,22 @@ MigrationTypeList getMigrationTypes(
...
+	 * @throws NotFoundException
+	 */
+	@ResponseStatus(HttpStatus.OK)
+	@RequestMapping(value = { UrlHelpers.MIGRATION_TYPE_NAMES }, method = RequestMethod.GET)
+	public @ResponseBody
+	MigrationTypeNames getMigrationTypeNames(
+			@RequestParam(value = AuthorizationConstants.USER_ID_PARAM) Long userId) throws DatastoreException, NotFoundException {
+		return serviceProvider.getMigrationService().getMigrationTypeNames(userId);
+	}
+
```
xschildw, line 318: Add tests in adminClient to cover these 2 calls.

PR #2794, opened 2017-01-12
PLFM-4209

kimyen:
return a 400-level status when someone tries to create a message with no recipients.


brucehoff:
Looks great!

PR #2805, opened 2017-01-23
PLFM-4177

kimyen:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/persistence/DBOPrincipalHeader.java
```diff
@@ -36,7 +36,7 @@
 	@Field(name = SqlConstants.COL_PRINCIPAL_HEADER_PRINCIPAL_TYPE, nullable = false)
 	private PrincipalType principalType;
 
-	@Field(name = SqlConstants.COL_PRINCIPAL_HEADER_DOMAIN_TYPE, nullable = false)
+	@Field(name = SqlConstants.COL_PRINCIPAL_HEADER_DOMAIN_TYPE, nullable = true)
 	private DomainType domainType;
```
john-hill, line 40: Determine if we use this table at all.  If not we should remove the table (not just the column).

PR #2806, opened 2017-01-23
PLFM-4127

kimyen:
treating null as false for includeFileHandles and includePreSignedURLs


services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/file/FileHandleManagerImplTest.java
```diff
@@ -943,8 +943,8 @@ public void testGetFileHandleAndUrlBatch() throws Exception {
 	}
 	
 	@Test
-	public void testGetFileHandleAndUrlBatchUrlsOnly() throws Exception {	
-		batchRequest.setIncludeFileHandles(false);
+	public void testGetFileHandleAndUrlBatchUrlsOnly() throws Exception {
```
john-hill: need a null test that is the same as false, but still need the false test.

PR #2815, opened 2017-01-27
PLFM-4205

kimyen:
add API to retrieve subscriber count for a given topic


services/repository/src/main/java/org/sagebionetworks/repo/web/controller/SubscriptionController.java
```diff
@@ -173,4 +174,19 @@ public void deleteAll(
 			@RequestParam(value = "nextPageToken", required = false) String nextPageToken) {
 		return serviceProvider.getSubscriptionService().getSubscribers(userId, topic, nextPageToken);
 	}
+
+	/**
+	 * Retrieve number of subscribers for a given topic.
+	 *
```
john-hill, line 180: 'Retrieve the number'
Also
Note the permissions needed to make this call.

PR #2858, opened 2017-03-03
removed more references to project and benefactor

john-hill:


services/workers/src/main/java/org/sagebionetworks/object/snapshot/worker/utils/NodeObjectRecordWriter.java
```diff
@@ -126,7 +129,9 @@ public void buildAndWriteRecords(ProgressCallback<Void> progressCallback, List<C
 			} else {
 				try {
 					Node node = nodeDAO.getNode(message.getObjectId());
-					NodeRecord record = buildNodeRecord(node);
+					String benefactorId = nodeInheritanceDao.getBenefactor(message.getObjectId());
+					String projectId = nodeDAO.getProjectId(message.getObjectId());
+					NodeRecord record = buildNodeRecord(node, benefactorId, projectId);
```
john-hill, line 134: optimize to reduce the number of database calls.

PR #2898, opened 2017-03-28
add concrete type to access requirement table

kimyen:


lib/jdomodels/src/main/resources/schema/AccessRequirement-ddl.sql
```diff
@@ -6,8 +6,9 @@ CREATE TABLE IF NOT EXISTS `ACCESS_REQUIREMENT` (
   `MODIFIED_BY` bigint(20) NOT NULL,
   `MODIFIED_ON` bigint(20) NOT NULL,
   `ACCESS_TYPE` varchar(256) CHARACTER SET latin1 COLLATE latin1_bin NOT NULL,
+  `CONCRETE_TYPE` varchar(100) CHARACTER SET latin1 COLLATE latin1_bin,
```
kimyen, line 9: could use an enum

PR #2901, opened 2017-03-31
PLFM-4320

kimyen:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/DBOAccessApprovalDAOImpl.java
```diff
@@ -59,6 +60,12 @@
 		COL_ACCESS_APPROVAL_REQUIREMENT_ID+" IN (:"+COL_ACCESS_APPROVAL_REQUIREMENT_ID+
 		") AND "+COL_ACCESS_APPROVAL_ACCESSOR_ID+" IN (:"+COL_ACCESS_APPROVAL_ACCESSOR_ID+")";
 
+	private static final String SELECT_MET_ACCESS_REQUIREMENT_COUNT =
+			"SELECT COUNT(*)"
```
john-hill, line 64: count( distinct access_requirement_id)

PR #2933, opened 2017-04-18
PLFM-4336: handle double annotations

kimyen:


lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/SQLUtils.java
```diff
@@ -1264,12 +1264,41 @@ public static void buildSelect(StringBuilder builder,
 		builder.append(".");
 		builder.append(TableConstants.ENTITY_REPLICATION_COL_VERSION);
 		for(ColumnMetadata meta: metadata){
-			builder.append(", ");
-			builder.append(meta.getTableAlias());
-			builder.append(".");
-			builder.append(meta.getSelectColumnName());
-			builder.append(" AS ");
-			builder.append(meta.getColumnNameForId());
+			if (AnnotationType.DOUBLE.equals(meta.getAnnotationType())) {
```
john-hill, line 1267: I was looking for constants for 'NaN', 'Infinity' and '-Infinity' but I did not find any.  Instead, I found TableModelUtils.validateValue().  It seems valid values are case insensitive and include more options like '+inf' and '-inf'.  We need to try and support all of the cases.

john-hill:
Good Job!

PR #2934, opened 2017-04-18
Entity query refactor

john-hill:


lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/query/entity/TablesTest.java
```diff
@@ -0,0 +1,103 @@
...
+	@Test
+	public void testSortOnAnnotation(){
+		sortIndex = 0;
+		sortColumnName = "foo";
+		sortList = new SortList(sortIndex, sortColumnName, sortIsAscending);
+		Tables tables = new Tables(new ExpressionList(new LinkedList<Expression>()), sortList);
+		String sql = tables.toSql();
+		assertEquals(" FROM ENTITY_REPLICATION E LEFT JOIN ANNOTATION_REPLICATION A0"
+				+ " ON (R.ID = A0.ENTITY_ID AND A0.ANNO_KEY = :bJoinName0)", sql);
+	}
```
xschildw, line 76: Double-check mySQL behaviour, might add an unwanted filter on ANNO_KEY.

PR #2945, opened 2017-04-27
PLFM-1970 & PLFM-4356

kimyen:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/DBOAccessApprovalDAOImpl.java
```diff
@@ -122,24 +137,35 @@ public void delete(String id) throws DatastoreException, NotFoundException {
...
-		// don't create if it already exists
-		List<AccessApproval> existingApprovals = getForAccessRequirementsAndPrincipals(
-				Collections.singletonList(dto.getRequirementId().toString()),
-				Collections.singletonList(dto.getAccessorId()));
-		if (!existingApprovals.isEmpty()) {
-			return (T)existingApprovals.get(0);
-		}
-		
-		DBOAccessApproval dbo = new DBOAccessApproval();
+	public <T extends AccessApproval> T create(T dto) throws DatastoreException {
```
john-hill: There are two create methods: one for single values and another for batches.  The same logic is implemented twice.  This means if there is a bug we must fix it in both spots, just like we had to do for this pull-request.  The batch method can be used for all cases.  There is no need for duplicate code.

PR #2954, opened 2017-05-01
Commit status

kimyen:


before.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+# The required environment variables are:
+# user - e.g. 'pjmhill'
+
+# optional variables for github commit status API
+# github_token - the token that is used to push commit status
```
brucehoff, line 6: can you add guidance as to how to get a github token (perhaps link to the dev' bootstrap)?

after.sh
```diff
@@ -0,0 +1,30 @@
...
+# user - e.g. 'pjmhill'
+
+# optional variables for github commit status API
+# github_token - the token that is used to push commit status
+
+# require to check status
+# result - the result of the jenkins_build.sh -- needs to be captured in Jenkins' cell
+
+
+stack=dev
```
brucehoff: ideally, an environment variable like 'stack' would only be defined in one place.  You define it both in 'before' and 'after'

brucehoff:
Shouldn't there be an overarching script, e.g. 'travis_build.sh' that calls before.sh, docker_build.sh and after.sh?

before.sh
kimyen: already there

after.sh
kimyen: agree. I didn't see that you are reviewing. I'm also moving before.sh and after.sh to jenkins_build.sh. What do you think about moving stack to jenkins_build.sh since it's also defined in docker_build.sh

PR #3009, opened 2017-05-17
PLFM-4395

kimyen:


lib/jdomodels/src/main/resources/schema/AccessApproval-ddl.sql
```diff
@@ -6,10 +6,12 @@ CREATE TABLE IF NOT EXISTS `ACCESS_APPROVAL` (
   `MODIFIED_BY` bigint(20) NOT NULL,
   `MODIFIED_ON` bigint(20) NOT NULL,
   `REQUIREMENT_ID` bigint(20) NOT NULL,
-  `ACCESSOR_ID`  bigint(20) NOT NULL,
+  `REQUIREMENT_VERSION` bigint(20) DEFAULT 0,
+  `RESEARCH_PROJECT_ID` bigint(20) DEFAULT 0,
```
john-hill: change to submitter ID.

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/DBOAccessApprovalDAOImpl.java
```diff
@@ -276,8 +223,18 @@ public void setValues(PreparedStatement ps, int i) throws SQLException {
...
-				ps.setBytes(9, dbos.get(i).getSerializedEntity());
+				if (dbos.get(i).getRequirementVersion() != null) {
+					ps.setLong(8, dbos.get(i).getRequirementVersion());
+				} else {
+					ps.setLong(8, DBOAccessRequirementDAOImpl.DEFAULT_VERSION);
+				}
+				if (dbos.get(i).getResearchProjectId() != null) {
+					ps.setLong(9, dbos.get(i).getResearchProjectId());
+				} else {
+					ps.setNull(9, Types.INTEGER);
```
john-hill: This will be changed to submitterId and cannot be null.

PR #3016, opened 2017-05-24
Plfm 4385

john-hill:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/DBOAccessRequirementDAOImpl.java
```diff
@@ -159,19 +182,36 @@
...
+	 * @return
+	 */
+	private List<AccessRequirement> getAccessRequirements(List<Long> requirementIds){
+		if(requirementIds.isEmpty()){
+			return new LinkedList<>();
+		}
+		MapSqlParameterSource param = new MapSqlParameterSource();
+		param.addValue(COL_ACCESS_REQUIREMENT_ID.toLowerCase(), requirementIds);
+		List<AccessRequirement> dtos = namedJdbcTemplate.query(SELECT_CURRENT_REQUIREMENTS_BY_ID, param, requirmentMapper);
+		for (AccessRequirement dto : dtos) {
```
xschildw: Use subjects from the blob, don't run n queries.

PR #3022, opened 2017-05-30
when subjectIds is null return an empty list

john-hill:


jay-hodgson:
This fix doesn't make sense to me.  
AR ID 5612415 should return a valid subject list (not empty or null).

PR #3028, opened 2017-06-01
PLFM-4422

kimyen:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/persistence/DBOAccessApproval.java
```diff
@@ -39,7 +37,6 @@
 	private long expiredOn;
 	private Long requirementId;
 	private Long accessorId;
-	private byte[] serializedEntity;
```
kimyen, line 42: leave this field and it's methods
remove translator's logic about previous migration.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/dataaccess/SubmissionManagerImpl.java
```diff
@@ -265,20 +262,22 @@ public SubmissionPage listSubmission(UserInfo userInfo, SubmissionPageRequest re
 		return pageResult;
 	}
 
+	// TODO: this API will change
```
kimyen: remove this comment

PR #3038, opened 2017-06-14
Plfm 4442

john-hill:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/dataaccess/RequestManagerImpl.java
```diff
@@ -96,7 +82,7 @@ private RequestInterface createNewRequest(String accessRequirementId) {
 		return request;
 	}
 
-	public Renewal createRenewalFromRequest(RequestInterface current) {
+	public static Renewal createRenewalFromRequest(RequestInterface current) {
```
john-hill: rename createRenewalFromApprovedRequest()

PR #3058, opened 2017-06-30
PLFM-4463 - add SelfSignAccessRequirement

kimyen:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/AccessRequirementManagerImpl.java
```diff
@@ -263,32 +264,41 @@ public void deleteAccessRequirement(UserInfo userInfo,
...
+			if (req.getIsCertifiedUserRequired() == null) {
+				req.setIsCertifiedUserRequired(false);
+			}
+			if (req.getIsValidatedProfileRequired() == null) {
+				req.setIsValidatedProfileRequired(false);
+			}
+			return req;
 		}
-		return actAR;
+		return ar;
```
john-hill: Since there is no overlap between these types each type should have its own method.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/message/dataaccess/HasAccessorRequirementUtil.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.Set;
+
+import org.sagebionetworks.repo.model.AuthorizationConstants;
+import org.sagebionetworks.repo.model.GroupMembersDAO;
+import org.sagebionetworks.repo.model.HasAccessorRequirement;
+import org.sagebionetworks.repo.model.VerificationDAO;
+import org.sagebionetworks.util.ValidateArgument;
+
+public class HasAccessorRequirementUtil {
+
```
john-hill: The static utility pattern should be reserved for pure functional programming.  The core design principal of PLFM is inversion-of-control (IoC).  I should be able to ask if a set of users has met the requirements without knowledge of which DAOs are needed to answer the question.  This should be in the AuthorizationManager.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/AccessRequirementManagerImpl.java
```diff
@@ -263,9 +264,34 @@ public void deleteAccessRequirement(UserInfo userInfo,
...
+			return setDefaultValuesForSelfSignAccessRequirement(ar);
 		}
+		return ar;
+	}
+
+	/**
+	 * @param ar
+	 * @return
+	 */
+	public static AccessRequirement setDefaultValuesForSelfSignAccessRequirement(AccessRequirement ar) {
```
john-hill, line 279: why not use method overloading?

PR #3066, opened 2017-07-03
PLFM-2971 - Disallow canPublicJoin team invtns and requests

hhu94:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/team/MembershipInvitationManagerImpl.java
```diff
@@ -76,6 +77,10 @@ public MembershipInvtnSubmission create(UserInfo userInfo,
 		if (!authorizationManager.canAccess(
 				userInfo, mis.getTeamId(), ObjectType.TEAM, ACCESS_TYPE.TEAM_MEMBERSHIP_UPDATE).getAuthorized()) 
 			throw new UnauthorizedException("Cannot create membership invitation.");
+		Team team = teamDAO.get(mis.getTeamId());
+		if (team.getCanPublicJoin() != null && team.getCanPublicJoin()) {
+			throw new IllegalArgumentException("Cannot invite to public team.");
```
kimyen: This message will be shown to user. We need to provide more friendly and helpful message. For example, explaining that the team they are trying to invite users to is already open for public to join, therefore, a membership invitation is not needed here.

kimyen: Also, verify with @jay-hodgson, in case that a team is open for public, how would a user "invite" someone to join beside this method. Is there an alternative way to invite users without creating an MembershipInvitation?

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/team/MembershipRequestManagerImpl.java
```diff
@@ -103,6 +104,10 @@ public MembershipRqstSubmission create(UserInfo userInfo,
 			if (hasUnmetAccessRequirements(userInfo, mrs.getTeamId()))
 				throw new UnauthorizedException("Requested member has unmet access requirements which must be met before asking to join the Team.");
 		}
+		Team team = teamDAO.get(mrs.getTeamId());
+		if (team.getCanPublicJoin() != null && team.getCanPublicJoin()) {
+			throw new IllegalArgumentException("Cannot invite to public team.");
```
kimyen: Same here, we should explain that the team is open for public to join. Therefore, user doesn't have to create a request.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/team/MembershipRequestManagerImpl.java
```diff
@@ -103,6 +104,10 @@ public MembershipRqstSubmission create(UserInfo userInfo,
 			if (hasUnmetAccessRequirements(userInfo, mrs.getTeamId()))
 				throw new UnauthorizedException("Requested member has unmet access requirements which must be met before asking to join the Team.");
 		}
+		Team team = teamDAO.get(mrs.getTeamId());
```
kimyen, line 107: Can users get into this state from SWC?

kimyen: Verified with Jay that this change won't affect the web client.

PR #3071, opened 2017-07-07
PLFM-3718 - Add links to user profile and team pages in notification emails

hhu94:


services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/team/EmailParseUtil.java
```diff
@@ -44,4 +46,32 @@ public static String getTokenFromString(String s, String startString, String end
...
+				if (piece.charAt(0) == '#' && piece.charAt(piece.length() - 1) == '#') {
+					// Current piece is a variable
+					regex += ".*";
+				} else {
+					// Current piece is a literal string
+					regex += Pattern.quote(piece);
+				}
+			}
+		}
+		Pattern pattern = Pattern.compile(regex);
```
kimyen: This is expensive. Why do we need this method?

kimyen: The use of this method is for testing EmailUtils.readMailTemplate(String, Map<String, String>). 
The above static method should be fully tested. Its caller should not test its logic.
Please make sure that EmailUtils.readMailTemplate() is fully tested, and remove the redundant tests.

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/team/MembershipInvitationManagerImplTest.java
```diff
@@ -261,6 +263,17 @@ public void testGetOpenSubmissionsForTeamAndRequesterInRangeUnauthorized() throw
...
 	public void testCreateInvitationNotification() throws Exception {
 		MembershipInvtnSubmission mis = createMembershipInvtnSubmission(MIS_ID);
+		testCreateInvitationNotificationHelper(mis);
+	}
+	
+	@Test
+	public void testCreateInvitationNotificationWithNullCreatedOn() throws Exception {
+		MembershipInvtnSubmission mis = createMembershipInvtnSubmission(MIS_ID);
+		mis.setCreatedOn(null);
+		testCreateInvitationNotificationHelper(mis);
```
kimyen, line 273: testCreateInvitationNotification() is fully testing that membershipInvitationManagerImpl.createInvitationNotification() returns the correct values with createdOn field being set in mis. 

This test, testCreateInvitationNotificationWithNullCreatedOn(), should only test that the method doesn't fail for mis with createdOn set to null. The behavior we expect is that the method doesn't fail when createdOn is null. We should not expect that the link still be the same. This will make the test unstable when it's run on a slower machine.

hhu94: The helper method creates the expected string after calling createInvitationNotification, and createInvitationNotification should always make sure that createdOn is set. That's why the test should always succeed and be stable.

PR #3075, opened 2017-07-09
PLFM-4511

xschildw:
Worked for my private build.


brucehoff:
Looks good, I will merge the PR.  I'd still like to know where the volumes are coming from.  As you can see in the script, it creates a network and two containers, but no volumes.

PR #3077, opened 2017-07-10
PLFM-4476 - API to retrieve a page of subjects given an Access Requirement ID

kimyen:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/AccessRequirementManagerImpl.java
```diff
@@ -388,4 +390,16 @@ public static ManagedACTAccessRequirement convert(ACTAccessRequirement current,
 		toUpdate.setVersionNumber(current.getVersionNumber()+1);
 		return toUpdate;
 	}
+
+	@Override
+	public RestrictableObjectDescriptorResponse getSubjects(UserInfo userInfo, String accessRequirementId, String nextPageToken){
```
john-hill, line 395: Will check is authenticated users is needed to see an access requirement.

PR #3084, opened 2017-07-18
Plfm 4468

john-hill:


lib/models/src/main/java/org/sagebionetworks/repo/model/table/parser/DoubleTypeParser.java
```diff
@@ -0,0 +1,20 @@
+package org.sagebionetworks.repo.model.table.parser;
```
xschildw, line 1: Is this used?

lib/models/src/main/java/org/sagebionetworks/repo/model/table/parser/DoubleParser.java
```diff
@@ -1,15 +1,21 @@
...
+import org.sagebionetworks.repo.model.table.AbstractDouble;
 
-public class DoubleParser implements ValueParser {
+
+public class DoubleParser extends AbstractValueParser {
 
 	@Override
 	public Object parseValueForDatabaseWrite(String value) throws IllegalArgumentException {
+		if(value == null){
+			return null;
```
xschildw, line 11: Check if should throw exception

lib/models/src/test/java/org/sagebionetworks/repo/model/table/parser/DoubleTypeParserTest.java
```diff
@@ -0,0 +1,78 @@
+package org.sagebionetworks.repo.model.table.parser;
```
xschildw, line 1: Remove

PR #3093, opened 2017-07-24
PLFM-4439

xschildw:


services/workers/src/main/java/org/sagebionetworks/file/worker/PreviewWorker.java
```diff
@@ -91,10 +93,13 @@ public void run(ProgressCallback progressCallback, ChangeMessage changeMessage)
 					true);
 			throw new RecoverableMessageException();
 		} catch (Throwable e) {
-			// If we do not know what went wrong then we do no re-try
-			log.error("Failed to process message: " + changeMessage.toString(), e);
-			workerLogger.logWorkerFailure(this.getClass(), changeMessage, e,
-					false);
+			if (e.getCause() instanceof IIOException) {
```
xschildw, line 96: Check for null cause.

services/workers/src/test/java/org/sagebionetworks/file/worker/PreviewWorkerTest.java
```diff
@@ -194,7 +190,7 @@ public void testErrorReadingPNG() throws Exception{
 		when(mockPreveiwManager.generatePreview(meta)).thenThrow(expectedException);
 		// Fire!
 		worker.run(mockProgressCallback, change);
-		verify(mockWorkerLogger).logWorkerFailure(PreviewWorker.class, change, expectedException, false);
+		verifyZeroInteractions(mockWorkerLogger);
```
xschildw, line 193: Verify it's calling log.info().

services/workers/src/main/java/org/sagebionetworks/file/worker/PreviewWorker.java
xschildw: From http://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.20.2, no need to check for null, instanceof will be false in that case.

PR #3096, opened 2017-07-29
Throw NotFoundException if getBundle() returns no row.

xschildw:


brucehoff:
Very nice:  With this change a 500 status will become a 404, which is preferred.

PR #3107, opened 2017-09-01
list aliases

john-hill:


services/repository/src/main/java/org/sagebionetworks/repo/web/controller/UserProfileController.java
```diff
@@ -284,6 +286,26 @@ UserGroupHeaderResponsePage getUserGroupHeadersByPrefix(
...
+	 *         of the request is preserved in the response. If a requested alias
+	 *         does not match an existing user name or team name then no header
+	 *         will be returned.
+	 */
+	@ResponseStatus(HttpStatus.OK)
+	@RequestMapping(value = UrlHelpers.USER_GROUP_HEADERS_BY_ALIASES, method = RequestMethod.POST)
+	public @ResponseBody
+	UserGroupHeaderResponse getUserGroupHeadersByAliases(
+			@RequestBody AliasList request) {
+		return serviceProvider.getUserProfileService()
```
jay-hodgson, line 304: Do we need to set a maximum count on the number of aliases to resolve?

integration-test/src/test/java/org/sagebionetworks/IT500SynapseJavaClient.java
```diff
@@ -676,6 +676,27 @@ public void testGetUserGroupHeaders() throws Exception {
 		page = synapseOne.getUserGroupHeadersByPrefix(adminProfile.getUserName(), type, 5L, 0L);
 		assertTrue(page.getTotalNumberOfResults()>0);
 		
+		// Get the user group header using the user's name
+		List<String> aliases = Lists.newArrayList(adminProfile.getUserName());
+		List<UserGroupHeader> headers = synapseOne.getUserGroupHeadersByAliases(aliases);
+		assertNotNull(headers);
```
jay-hodgson, line 682: remove

services/repository/src/main/java/org/sagebionetworks/repo/web/controller/UserProfileController.java
jay-hodgson: nevermind, looks like it's 100!

PR #3113, opened 2017-09-15
Plfm 4566 b

john-hill:


lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/model/EscapeCharacter.java
```diff
@@ -1,30 +1,13 @@
...
-import java.util.List;
-
 /**
- * This matches &ltescape character&gt  in: <a href="https://github.com/ronsavage/SQL/blob/master/sql-92.bnf">SQL-92</a>
+ * EscapeCharacter ::= {@link CharacterValueExpression}
  */
-public class EscapeCharacter extends SQLElement {
+public class EscapeCharacter extends SimpleBranch {
 
 	CharacterValueExpression characterValueExpression;
```
john-hill, line 8: should not keep a reference to CharacterValueExpression

lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/model/Factor.java
```diff
@@ -2,11 +2,17 @@
 
 import java.util.List;
```
john-hill, line 3: Missing FactorTest

lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/model/Predicate.java
```diff
@@ -63,19 +63,19 @@ public BooleanFunctionPredicate getBooleanFunctionPredicate() {
 	}
 
 	@Override
-	public void toSql(StringBuilder builder) {
```
john-hill, line 66: Why not make this a SimpleBranch?

lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/model/SQLElement.java
```diff
@@ -14,24 +14,31 @@
 
 	private static final String EMPTY = "";
 	private static final String REGEX_QUOTES = "[',\"]";
```
john-hill, line 16: These regex are dead code

lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/model/ValueExpressionPrimary.java
```diff
@@ -3,16 +3,16 @@
 import java.util.List;
 
 /**
- * This matches &ltvalue expression primary&gt   in: <a href="https://github.com/ronsavage/SQL/blob/master/sql-92.bnf">SQL-92</a>
+ * ValueExpressionPrimary ::= {@link UnsignedValueSpecification} | {@link ColumnReference} | {@link SetFunctionSpecification}
  */
 public class ValueExpressionPrimary extends SQLElement implements HasReferencedColumn {
```
john-hill, line 9: Consider SimleBranch

PR #3115, opened 2017-09-18
PLFM-4614 (Subset of PLFM-2073)

hhu94:


hhu94:
Will break down into two separate pull requests for better separation of concerns. One PR to add the etag column and a second PR for the rest of the work.

PR #3116, opened 2017-09-20
PLFM-4619 - Verify that etag columns are non null in migration

hhu94:
Change all tables to have non null etag columns, both ddl and annotation
based.

Do not merge until we have made sure that every table in production has non nullable etag columns.


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/migration/MigratableTableDAOImpl.java
```diff
@@ -225,6 +229,25 @@ private void registerObject(MigratableDatabaseObject dbo, boolean isRoot) {
 		registeredMigrationTypes.add(dbo.getMigratableTableType());
 
 	}
+
+	private void validateEtagColumn(String tableName, String columnName) {
+		try (
```
xschildw: Please implement using jdbcTemplate and a query against INFORMATION_SCHEMA, as you showed me in mySQLWorkbench.

PR #3118, opened 2017-09-21
PLFM-2206 - Allow invitee to get and delete an invitation

hhu94:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/AuthorizationManagerImpl.java
```diff
@@ -128,6 +111,15 @@ public AuthorizationStatus canAccess(UserInfo userInfo, String objectId, ObjectT
 					return AuthorizationManagerUtil.AUTHORIZED;
 				}
 				return AuthorizationManagerUtil.accessDenied("Only ACT member can perform this action.");
+			case MEMBERSHIP_INVITATION:
+				MembershipInvtnSubmission mis = membershipInvtnSubmissionDAO.get(objectId);
```
brucehoff: instead of getting the entire object, add a new query (if not already in place) to get just the invitee ID for the invitation.

brucehoff: The logic for authorizing access to invitations is complex enough that it needs to be factored into it's own method (and tested there). The inputs are:
(1) the user's ID;
(2) the invitation ID;
(3) the access type (READ or DELETE);
 The logic is:
- if the user is the invitee and the access type is READ or DELETE, then they are approved.
- if the user has admin' access (represented as TEAM_MEMBERSHIP_UPDATE access) to the team for which membership is to be granted and the access type is READ or DELETE, then they are approved.
- if the user is a Synapse administrator then access is approved.
- if the user is anonymous then access is denied.

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/AuthorizationManagerImplTest.java
```diff
@@ -632,4 +631,19 @@ public void testCanSubscribeAnonymous() {
...
+		String misId = "1";
+		mis.setId(misId);
+		String teamId = "123";
+		mis.setTeamId(teamId);
+		mis.setInviteeId(userInfo.getId().toString());
+		mis.setMessage("Please join our team.");
+		MembershipInvtnSubmissionDAO mockMembershipInvtnSubmissionDAO = Mockito.mock(MembershipInvtnSubmissionDAO.class);
+		when(mockMembershipInvtnSubmissionDAO.get(misId)).thenReturn(mis);
+		ReflectionTestUtils.setField(authorizationManager, "membershipInvtnSubmissionDAO", mockMembershipInvtnSubmissionDAO);
+		assertTrue(authorizationManager.canAccess(userInfo, misId, ObjectType.MEMBERSHIP_INVITATION, ACCESS_TYPE.TEAM_MEMBERSHIP_UPDATE).getAuthorized());
```
brucehoff: add  a comment above "// method under test"

brucehoff: Tests to add:
- user is invitee, user is team mgr, user is neither
- various access types (READ, DELETE, other)

brucehoff:
Good work!

PR #3119, opened 2017-09-21
PLFM-4614 - Add etag to MembershipInvtnSubmission table

hhu94:


lib/jdomodels/src/main/resources/schema/MembershipInvitationSubmission-ddl.sql
```diff
@@ -1,5 +1,6 @@
 CREATE TABLE IF NOT EXISTS `MEMBERSHIP_INVITATION_SUBMISSION` (
   `ID` bigint(20) NOT NULL,
+  `ETAG` char(36) NOT NULL DEFAULT 'defaultEtag',
```
hhu94: Thoughts on this default?

PR #3133, opened 2017-10-04
PLFM-2073 Phase 1 - Remaining small tweaks

hhu94:
- [x] Rename verifyInvitee to getInviteeVerificationSignedToken
- [x] Rename /membershipInvitation/id/verification to /membershipInvitation/id/inviteeVerificationSignedToken
- [x] Remove parseMembershipInvtnSignedToken and getInviteeEmail (unused code)
- [x] Verify that expiresOn is not null in updateInviteeId


services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/team/MembershipInvitationManagerImplTest.java
```diff
@@ -441,6 +441,18 @@ public void testUpdateId() {
 		}
 		assertTrue(caughtException);
 
+		// Token with null expiresOn should fail
+		token.setExpiresOn(null);
+		caughtException = false;
+		try {
+			membershipInvitationManagerImpl.updateInviteeId(userId, MIS_ID, token);
+		} catch (IllegalArgumentException e) {
```
brucehoff, line 449: an alternative is to just put "fail();" after line 448.  The test will only reach this line if it fails to throw an exception.

try {
	membershipInvitationManagerImpl.updateInviteeId(userId, MIS_ID, token);
	fail("IllegalArgumentException expected");
} catch (IllegalArgumentException e) {
	// as expected
}

PR #3141, opened 2017-10-13
PLFM-4647: Only output msg from JDOSTACKSTATUS.

xschildw:


services/repository/src/main/java/org/sagebionetworks/repo/web/StackStatusInterceptor.java
```diff
@@ -49,7 +49,11 @@ public boolean preHandle(HttpServletRequest request,
 		}
 		if (StatusEnum.READ_ONLY == status || StatusEnum.DOWN == status) {
 			StackStatus full = stackStatusDao.getFullCurrentStatus();
-			throw new ServiceUnavailableException("Synapse is down for maintenance. Message: "+full.getCurrentMessage());
+			String msg = "Synapse is down for maintenance.";
+			if ((full.getCurrentMessage() != null) && (full.getCurrentMessage() != "")) {
```
jay-hodgson, line 53: != is comparing the object reference, not the value.  Please change to !full.getCurrentMessage().isEmpty() for the second condition.
Are there unit tests for this method?

PR #3143, opened 2017-10-13
PLFM-2073 - Change to encoded signed token

hhu94:
Leaving the MembershipInvtnSignedToken encoded and passing it through in that form reduces the work the web client needs to do.


hhu94:
Reviewed with Bruce

PR #3144, opened 2017-10-16
Include httpMethod in throttle rules matching.

xschildw:
Depends on PR in common utils.


john-hill:
Closing since this is over a year old.

PR #3150, opened 2017-10-24
update with new etags results

john-hill:


lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/model/SparseChangeSet.java
```diff
@@ -436,6 +454,10 @@ public String toString() {
 					+ ", rowVersionNumber=" + versionNumber + ", valueMap="
 					+ valueMap + "]";
 		}
+
```
xschildw: Remove

lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/model/SparseChangeSet.java
```diff
@@ -436,6 +454,10 @@ public String toString() {
 					+ ", rowVersionNumber=" + versionNumber + ", valueMap="
 					+ valueMap + "]";
 		}
+
+		private SparseChangeSet getOuterType() {
```
xschildw, line 458: Remove

PR #3165, opened 2017-11-03
used correct date type

john-hill:


pom.xml
```diff
@@ -590,9 +590,7 @@
 			<dependency>
 				<groupId>com.sun</groupId>
 				<artifactId>tools</artifactId>
-				<version>1.7</version>
-				<scope>system</scope>
-				<systemPath>${java.home}/../lib/tools.jar</systemPath>
+				<version>1.8.20171103</version>
```
brucehoff: please add a comment emphasizing that this explicitly references Java 8 and would have to be updated if we update the JDK used to build Synapse.

PR #3229, opened 2018-01-24
(PLFM-909) fixed conversion of SearchQuery objects into SearchRequest objects. R…

zimingd:
…eplaced some uses of StringBuilder with StringJoiner


zimingd:
already reviewed but merged to develop instead of release-213 in #3228

PR #3237, opened 2018-02-02
Added "consortium" search index. and modified all other annotation-based indicies to document only 1 annotation value (PLFM-4754)

zimingd:

PR #3243, opened 2018-02-08
Removed primary annotations from search document generation. Annotation keys are case insensitive for document generation(PLFM-4754)

zimingd:


zimingd:
follow up to #3237

PR #3283, opened 2018-03-17
merge release-219 into develop

zimingd:
Should be no diff from develop but I messed up the merge order of 218, 219, and develop


zimingd:
closing because there is merge conflict and if I resolve merge conflict in Github, develop will end up merged into release-219 first

PR #3304, opened 2018-05-14
increased max number of field values added to a cloudsearch document from 100 to 1000 (PLFM-4950)

zimingd:


zimingd:
We need to investigate possible performance impacts before deciding to make this change.

PR #3316, opened 2018-06-01
Hyperlink DOIs against preferred resolver

katrinleinweber:
Same as https://github.com/Sage-Bionetworks/Synapse-Markdown/pull/21 ;-)


nickgros:
We no longer poll the DOI resolver to verify that a DOI is created, so this isn't an issue anymore. We now just make requests to DataCite, our DOI provider ([using HTTPS here](https://github.com/Sage-Bionetworks/Synapse-Repository-Services/blob/23315c3ddccf6aab3a9076db4e2e03c485388420/lib/lib-doi/src/main/java/org/sagebionetworks/doi/datacite/DataciteClientImpl.java#L45)).

Thank you for your contribution!

PR #3318, opened 2018-06-06
Added way to specify sorting and count limiting of facets of search results (PLFM-4870)

zimingd:


lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/search/query/SearchFacetOption.json
```diff
@@ -0,0 +1,29 @@
+{
+	"description": "Specify which field should be returned as facets and the format of the results for each field",
+	"properties": {
+		"name": {
```
john-hill, line 4: make and enum and coordinate with Jay on the values (with an abstraction if possible).

zimingd:
Closing for now because changes are incomplete and further refactoring is needed. I will come back to this after I finish a different issue that has higher priority.

PR #3329, opened 2018-07-04
set string max as max of all with the same name

john-hill:


john-hill:
This does not take pagination into consideration.

PR #3331, opened 2018-07-06
Plfm 5034

john-hill:


lib/lib-table-cluster/src/test/java/org/sagebionetworks/table/cluster/TableIndexDAOImplTest.java
```diff
@@ -1469,6 +1470,121 @@ public void testGetPossibleAnnotationsForContainers(){
...
+		file1.getAnnotations().add(annoDto);
+
+		tableIndexDAO.addEntityData(mockProgressCallback, Lists.newArrayList(file1, file2));
+		
+		Set<Long> containerIds = Sets.newHashSet(222L, 333L);
+		long limit = 5;
+		long offset = 0;
+		List<ColumnModel> columns = tableIndexDAO.getPossibleColumnModelsForContainers(containerIds, ViewTypeMask.File.getMask(), limit, offset);
+		assertNotNull(columns);
+		assertEquals(2, columns.size());
```
john-hill, line 1514: use a set of ColumnModels instead of the if

PR #3358, opened 2018-08-13
PLFM-4817

nickgros-sage:
use a new team search sql query to search for only usergroups that have an associated team (currently teamsearch retrieves usergroups, which causes a notfounderror to be thrown when finding a matching usergroup that does not have a team)


brucehoff:
Another home run!

PR #3363, opened 2018-08-22
PLFM-5064 and PLFM-5065

nickgros-sage:
This should prepare the DBO and DAO levels for the DOI API changes

### DIFF
--- a/lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/persistence/DBODoi.java
+++ b/lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/persistence/DBODoi.java
@@ -177,7 +177,9 @@ public class DBODoi implements MigratableDatabaseObject<DBODoi, DBODoi> {
 			@Override
 			public DBODoi createDatabaseObjectFromBackup(DBODoi backup) {
 				// Initialize updated fields to match the created fields
-				backup.setUpdatedBy(createdBy);
+				if (updatedBy == null) {
+					backup.setUpdatedBy(createdBy);
+				}
 				return backup;
 			}
 		};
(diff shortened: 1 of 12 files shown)


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/persistence/DBODoi.java
```diff
@@ -166,10 +176,8 @@ public MigrationType getMigratableTableType() {
 		return new BasicMigratableTableTranslation<DBODoi>() {
 			@Override
 			public DBODoi createDatabaseObjectFromBackup(DBODoi backup) {
-				// Convert null object version to -1
-				if (backup.objectVersion == null) {
-					backup.setObjectVersion(NULL_OBJECT_VERSION);
-				}
+				// Initialize updated fields to match the created fields
+				backup.setUpdatedBy(createdBy);
```
john-hill: only set if updated by is null

lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/dao/DoiUtilsTest.java
```diff
@@ -240,9 +309,103 @@ public void testConvertToDtoNullDbo() {
...
+	@Test(expected = IllegalArgumentException.class)
+	public void testConvertToDtoV2NullDbo() {
+		DBODoi dbo = null;
+		// Note DBO is null, so it should not be converted to a DTO.
+		// Call under test.
+		DoiUtils.convertToDtoV2(dbo);
+	}
+	
+	@Test
+	public void testIllegalArgExceptionOnMissingValue() {
```
john-hill, line 321: each in their own test

PR #3375, opened 2018-08-31
support for sessionId in Synapse Java Client (PLFM-4586)

zimingd:


services/repository/src/main/java/org/sagebionetworks/repo/web/filter/SessionIdCookieSetterFilter.java
```diff
@@ -29,7 +29,6 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
 			//add cookie to the response
 			Cookie sessionIdCookie = new Cookie(HttpRequestIdentifierUtils.SESSION_ID_COOKIE_NAME, UUID.randomUUID().toString());
 			sessionIdCookie.setHttpOnly(true);
-			sessionIdCookie.setSecure(true);
```
zimingd, line 32: This is supposed to make the client only remember the cookie if they are contacting via HTTPS. It is not necessary for our use case of the sessionID and makes local testing impossible since that's done over HTTP.

zimingd (commented):
I think I'll need get changes to the SimpleHttpClient pulled in before I can try to get a clean build:

https://github.com/Sage-Bionetworks/SimpleHttpClient/pull/10

client/synapseJavaClient/src/main/java/org/sagebionetworks/client/BaseClientImpl.java
```diff
@@ -247,6 +248,27 @@ public void setUserIpAddress(String ipAddress){
 		defaultPOSTPUTHeaders.put(X_FORWARDED_FOR_HEADER, ipAddress);
 	}
 
+	@Override
+	public void setSessionId(String sessionId){
+		try {
+			URL url = new URL(getRepoEndpoint());
```
zimingd: cache endpoint domain on set repo endpoint

PR #3387, opened 2018-09-10
PLFM-3574

nickgros-sage:
this should eliminate 500 errors when clients call a known URI with an unsupported verb


brucehoff:
Nice!

PR #3391, opened 2018-09-12
PLFM-5064 PLFM-5115 assorted, mostly DOI-related changes

nickgros-sage:
- use stack instance to change the demo DOI prefix when not on prod (helps avoid DOI minting collisions and clarifies which builds are minting which DOIs)
- set DOI locate controller to use the manager for path params (so the generated URL matches the request managed by the controller)
- fix DOI DAO update call to get the old object before making a database call (this way the ID, associatedBy, etc. fields do not change on update)
- only make one database call to update a DOI
- enable DoiAssociation in entity bundle
- REST doc enhancements for DOI and entitybundle
- new transaction annotation MandatoryWriteReadCommitted
- changes in DOI DAO and manager to apply new transaction anno


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/DBODoiAssociationDaoImpl.java
```diff
@@ -98,6 +98,11 @@ public DoiAssociation createDoiAssociation(DoiAssociation dto) throws DuplicateK
 	@WriteTransaction
 	@Override
 	public DoiAssociation updateDoiAssociation(DoiAssociation dto) {
+		// Fill in the fields from the existing object that required for the DBO that the client may not pass in
+		DoiAssociation existing = getDoiAssociation(dto.getObjectId(), dto.getObjectType(), dto.getObjectVersion());
```
nickgros-sage: Here we're making the second call to retrieve this DOI. Prior to this, the manager attempted to retrieve an eTag for the object.

I initially see two logical ways to avoid making this second call, and both involve avoiding the use of getEtagForUpdate:
- Remove this getDoiAssociation call and retrieve the entire association in the manager, instead of just the eTag. In this case, the DAO must assume the manager is setting all of the fields that should be immutable (like the ID of the association)
- Keep this getDoiAssociation call and compare the eTags in the DAO. The getDoiAssociation will throw a NotFoundException if it doesn't exist (thrown to the manager via this method) and this method can throw a ConflictingUpdateException if the eTags do not match. The manager then simply handles the exceptions as it does now, with the only change being the delegation the requirements of getEtagForUpdate to this method.

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/DBODoiAssociationDaoImpl.java
```diff
@@ -95,9 +96,25 @@ public DoiAssociation createDoiAssociation(DoiAssociation dto) throws DuplicateK
 		return getDoiAssociation(newId.toString());
 	}
 
-	@WriteTransaction
+	@RequiresNewReadCommitted
```
nickgros-sage: This fixes it but I don't know why?

Am I violating anything?

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/DBODoiAssociationDaoImpl.java
```diff
@@ -145,7 +162,6 @@ public DoiAssociation getDoiAssociation(String objectId, ObjectType objectType,
 		return DoiUtils.convertToDtoV2(dbo);
 	}
 
-	@NewWriteTransaction
 	@Override
 	public String getEtagForUpdate(String objectId, ObjectType objectType, Long versionNumber) throws NotFoundException {
```
nickgros-sage: getDoiForUpdate

lib/stackConfiguration/src/main/resources/stack.properties
```diff
@@ -149,6 +149,7 @@ org.sagebionetworks.repo.manage.shouldMessagesBePublishedToTopic=true
 
 #--- DOI Minting Service Credentials and Parameters ---
 org.sagebionetworks.doi.prefix=10.5072
+org.sagebionetworks.doi.demo.prefix=10.5072
```
nickgros-sage: remove demo

lib/stackConfiguration/src/main/java/org/sagebionetworks/StackConfigurationImpl.java
```diff
@@ -788,7 +788,11 @@ public String getEzidUrl() {
 	 * Prefix under which DOIs should be registered.
 	 */
 	public String getDoiPrefix() {
-		return configuration.getProperty("org.sagebionetworks.doi.prefix");
+		if (isProductionStack()) {
+			return configuration.getProperty("org.sagebionetworks.doi.prefix");
+		} else {
+			return configuration.getProperty("org.sagebionetworks.doi.demo.prefix") + "/" + getStackInstance();
```
nickgros-sage: // we do this because .....

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/DBODoiAssociationDaoImpl.java
```diff
@@ -67,20 +62,11 @@
 	 * DOI client creating the DOI is an asynchronous call and must happen outside the transaction to
 	 * avoid race conditions.
 	 */
-	@NewWriteTransaction
+	@MandatoryWriteReadCommittedTransaction
 	@Override
```
john-hill: `@WriteTransactionReadCommitted`

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/DBODoiAssociationDaoImpl.java
```diff
@@ -91,25 +77,14 @@ public DoiAssociation createDoiAssociation(DoiAssociation dto) throws DuplicateK
 				throw e;
 			}
 		}
-
 		return getDoiAssociation(newId.toString());
 	}
 
-	@WriteTransaction
+	@MandatoryWriteReadCommittedTransaction
```
john-hill: `@WriteTransactionReadCommitted`

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/DBODoiAssociationDaoImpl.java
```diff
@@ -128,26 +103,21 @@ public DoiAssociation getDoiAssociation(String id) throws NotFoundException {
...
-		MapSqlParameterSource paramMap = new MapSqlParameterSource();
-		paramMap.addValue(COL_DOI_OBJECT_ID, KeyFactory.stringToKey(objectId));
-		paramMap.addValue(COL_DOI_OBJECT_TYPE, objectType.name());
-		if (versionNumber == null) {
-			paramMap.addValue(COL_DOI_OBJECT_VERSION, DBODoi.NULL_OBJECT_VERSION);
-		} else {
-			paramMap.addValue(COL_DOI_OBJECT_VERSION, versionNumber);
-		}
-		DBODoi dbo = null;
+		return getDoiAssociation(objectId, objectType, versionNumber, false);
```
john-hill: use a parameter for 'forUpdate'

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/DBODoiAssociationDaoImpl.java
```diff
@@ -128,26 +103,21 @@ public DoiAssociation getDoiAssociation(String id) throws NotFoundException {
...
+	}
+
+	@MandatoryWriteReadCommittedTransaction
+	@Override
+	public DoiAssociation getDoiAssociationForUpdate(String objectId, ObjectType objectType, Long versionNumber) throws NotFoundException {
 		try {
-			dbo = namedJdbcTemplate.queryForObject(SELECT_DOI_BY_ASSOCIATED_OBJECT, paramMap, rowMapper);
-		} catch (IncorrectResultSizeDataAccessException e) {
-			handleIncorrectResultSizeException(e);
+			return getDoiAssociation(objectId, objectType, versionNumber, false);
```
john-hill: user parameter 'forUpdate' and set the value to true.

PR #3393, opened 2018-09-14
Plfm 5004

brucehoff:


client/synapseJavaClient/src/main/java/org/sagebionetworks/client/SynapseClientImpl.java
```diff
@@ -2514,7 +2514,7 @@ public SearchResults search(SearchQuery searchQuery) throws SynapseException {
 
 	@Override
 	public String getSynapseTermsOfUse() throws SynapseException {
```
jay-hodgson: SWC does not reference, safe to remove.

PR #3418, opened 2018-10-05
Plfm 4590

brucehoff:
This is just a clean up of the code base, no modification of functionality.


lib/jdomodels/src/main/java/org/sagebionetworks/evaluation/dao/EvaluationDAOImpl.java
```diff
@@ -154,6 +165,15 @@ public Evaluation get(String id) throws DatastoreException, NotFoundException {
 		}
 	}
 	
+	private static String authorizationSQLWhere() {
```
brucehoff: make this a static string

zimingd (approved):

PR #3423, opened 2018-10-09
integration test changed to use higher level

john-hill:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/table/TableQueryManagerImpl.java
```diff
@@ -402,6 +402,10 @@ public QueryResultBundle queryBundle(ProgressCallback progressCallback, UserInfo
 		if (options.returnMaxRowsPerPage()) {
 			bundle.setMaxRowsPerPage(queryResult.getMaxRowsPerPage());
 		}
+		if(options.runSumFileSizes()) {
```
john-hill: This option condition was missing and was the cause of the bug.

zimingd (approved):

PR #3446, opened 2018-11-09
PLFM-4727 remove POST and DELETE ACL for evaluations

nickgros:
https://sagebionetworks.jira.com/browse/PLFM-4727


brucehoff (approved):
Good job!

PR #3453, opened 2018-11-21
PLFM-5268 add GET /entity/{id}/dockerTag

nickgros:
This PR deprecates (but maintains functionality of) GET /entity/{id}/dockerCommit in favor of the identical-in-functionality but more-appropriately-named GET /entity/{id}/dockerTag

The old dockerCommit endpoint should be removed after the web client is configured to use dockerTag (the web client is the only user of the endpoint, per warehouse records)


brucehoff (approved):
Looks good!

PR #3456, opened 2018-11-28
PLFM-5288 remove deprecated dockerCommit URI and corresponding java client method

nickgros:


brucehoff:
@nickgros Has the web portal switched to the new Java client method (listDockerTags) instead of listDockerCommits?

nickgros:
@brucehoff Yes. See https://github.com/Sage-Bionetworks/SynapseWebClient/pull/3788

brucehoff (approved):
Looks great!

PR #3457, opened 2018-11-29
PLFM-5293

nickgros:
better username matching in discussions

### DIFF
--- a/services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/discussion/DiscussionUtilsTest.java
+++ b/services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/discussion/DiscussionUtilsTest.java
@@ -46,15 +46,9 @@ public class DiscussionUtilsTest {
 	@Test
 	public void testGetMentionedUsernameCaseDoubleAlpha(){
 		Set<String> expected = new HashSet<String>();
-		expected.add("anonymous");
+		expected.add("anon");
+		expected.add("ymous");
 		assertEquals(expected, DiscussionUtils.getMentionedUsername("@anon@ymous"));
-	}
-
-	@Test
-	public void testGetMentionedUsernameCaseContainSpecialChar(){
-		Set<String> expected = new HashSet<String>();
-		expected.add("anonymous?#$%!^&*)(+-_=`~/'\"|.,<>");
-		assertEquals(expected, DiscussionUtils.getMentionedUsername("@anonymous?#$%!^&*)(+-_=`~/'\"|.,<>"));
 	}
 
 	@Test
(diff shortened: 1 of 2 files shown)


services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/discussion/DiscussionUtilsTest.java
```diff
@@ -46,17 +46,11 @@ public void testGetMentionedUsernameCaseWhiteSpacePostfix(){
 	@Test
 	public void testGetMentionedUsernameCaseDoubleAlpha(){
 		Set<String> expected = new HashSet<String>();
-		expected.add("anonymous");
+		expected.add("anon");
+		expected.add("ymous");
 		assertEquals(expected, DiscussionUtils.getMentionedUsername("@anon@ymous"));
```
nickgros, line 51: The fix in this PR breaks this test, so I changed the expected result.

I don't quite understand why this test existed in it's previous state. Is there a reason to preserve the old behavior?

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/discussion/DiscussionUtilsTest.java
```diff
@@ -46,17 +46,11 @@ public void testGetMentionedUsernameCaseWhiteSpacePostfix(){
...
 	}
 
-	@Test
-	public void testGetMentionedUsernameCaseContainSpecialChar(){
-		Set<String> expected = new HashSet<String>();
-		expected.add("anonymous?#$%!^&*)(+-_=`~/'\"|.,<>");
-		assertEquals(expected, DiscussionUtils.getMentionedUsername("@anonymous?#$%!^&*)(+-_=`~/'\"|.,<>"));
```
nickgros, line 57: I am also unsure if there is a compelling reason to preserve this behavior.

nickgros: Replaced with getUsernameWithValidSpecialChars. See below.

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/discussion/DiscussionUtilsTest.java
brucehoff: I think it works better now.  Good job!

brucehoff (approved):
Nice!

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/discussion/DiscussionUtilsTest.java
kimyen: I do have much time to look at this now, but please do check how we use the parsed username to look up users. We may have done this because we are using services that list prefix of username. So if the username is not exact, we may get both user "kim" and "kimyen" for "@kim".

nickgros: @kimyen Thanks for bringing this up--I took a closer look. The method under test simply extracts full usernames, which it passes to the UserManager as-is to determine the user. The new regular expression is more restrictive and only looks for legal username characters. The old one allowed '@' in a username, which would never match to a real user. There is no prefix matching happening here.

kimyen: Thanks @nickgros for taking a look! I appreciate it. I should have looked at the changes before making any comments.

PR #3466, opened 2018-12-14
PLFM-5314 initial work

nickgros:
add column to entity replication table "isInSynapseStorage"


lib/models/src/main/java/org/sagebionetworks/repo/model/table/TableConstants.java
```diff
@@ -317,5 +320,10 @@ public static Long getReservedColumnId(String columnName) {
 					+" WHERE "
 					+REPLICATION_SYNC_EXP_COL_EXPIRES+" > :"+EXPIRES_PARAM
 					+" AND "+REPLICATION_SYNC_EXP_COL_ID+" IN (:"+ID_PARAMETER_NAME+")";
-	
+
+	public static final String SELECT_PROJECTS_BY_SIZE =
+			"SELECT "
```
john-hill, line 325: fix group by

john-hill (approved):

PR #3468, opened 2019-01-10
PLFM-5314

nickgros:
https://sagebionetworks.jira.com/browse/PLFM-5314

PR #3475, opened 2019-01-25
PLFM-5314

nickgros:
code review changes from #3468


integration-test/src/test/java/org/sagebionetworks/ITDownloadStorageReportTest.java
```diff
@@ -52,32 +41,18 @@ public static void afterClass() throws Exception {
...
 			try {
-				response = synapse.generateStorageReportAsyncGet(jobToken);
-				if (response.getResultsFileHandleId() != null) {
-					csvCreated = true;
-				}
-			} catch (SynapseResultNotReadyException e) {
-				assertNotEquals(e.getJobStatus().getJobState(), AsynchJobState.FAILED);
+				synapse.generateStorageReportAsyncGet(jobToken);
+				fail("Expected exception");
+			} catch (SynapseForbiddenException e) {
```
nickgros, line 52: A forbidden exception tells us the call reached the manager, which is sufficient for testing the controller

lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/TableIndexDAOImpl.java
```diff
@@ -960,17 +961,15 @@ public long getSumOfFileSizes(List<Long> rowIds) {
...
-				SQL_SELECT_PROJECTS_BY_SIZE,
-				(rs, rowNum) -> {
-					SynapseStorageProjectStats stats = new SynapseStorageProjectStats();
-					stats.setId(rs.getString(1));
-					stats.setProjectName(rs.getString(2));
-					stats.setSizeInBytes(rs.getLong(3));
-					return stats;
-				});
-
+	public void streamSynapseStorageStats(Callback<SynapseStorageProjectStats> callback) {
```
nickgros, line 964: Stream one row at a time to the callback, so we don't load the entire call into memory

lib/models/src/main/java/org/sagebionetworks/repo/model/report/SynapseStorageProjectStats.java
```diff
@@ -0,0 +1,70 @@
+package org.sagebionetworks.repo.model.report;
+
+public class SynapseStorageProjectStats {
```
nickgros, line 3: Convert the JSON->POJO to just a POJO, since it is never exposed client-side

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/report/StorageReportManagerImpl.java
```diff
@@ -24,32 +23,36 @@
...
+		if (request.getReportType() == null) {
+			request.setReportType(StorageReportType.ALL_PROJECTS);
+		}
+
+		switch (request.getReportType()) {
+			case ALL_PROJECTS:
+				String[] header = {"projectId", "projectName", "sizeInBytes"};
+				writer.writeNext(header);
+
+				Callback<SynapseStorageProjectStats> callback = value -> {
```
nickgros, line 44: Write the object to the CSV when the callback is invoked. The invocation should happen once per row in the DAO.

services/workers/src/main/java/org/sagebionetworks/report/worker/StorageReportCSVDownloadWorker.java
```diff
@@ -57,20 +55,16 @@ public void run(ProgressCallback progressCallback, Message message) throws Excep
 		AsynchronousJobStatus status = asynchJobStatusManager.lookupJobStatus(message.getBody());
 		String fileName = "Job-"+status.getJobId();
 		File temp = null;
-		CSVWriter writer = null;
-		try{
+		try {
 			UserInfo user = userManager.getUserInfo(status.getStartedByUserId());
 			DownloadStorageReportRequest request = AsynchJobUtils.extractRequestBody(status, DownloadStorageReportRequest.class);
 			// The CSV data will first be written to this file.
 			temp = File.createTempFile(fileName, ".csv");
```
nickgros, line 62: Should I use FileProvider? Originally considered to use it to mock a temp file and verify deletion. I did not switch to it here because mocking the temp file cannot be done. The CSVWriter requires a valid file, which cannot be mocked (it relies on setting private methods).

services/workers/src/main/java/org/sagebionetworks/report/worker/StorageReportCSVDownloadWorker.java
```diff
@@ -57,20 +55,16 @@ public void run(ProgressCallback progressCallback, Message message) throws Excep
 		AsynchronousJobStatus status = asynchJobStatusManager.lookupJobStatus(message.getBody());
 		String fileName = "Job-"+status.getJobId();
 		File temp = null;
```
nickgros, line 57: Look at other file provider methods. Perhaps need to add a method that adds a Writer

john-hill (approved):

PR #3499, opened 2019-02-26
PLFM-5404

nickgros:
https://sagebionetworks.jira.com/browse/PLFM-5404

DataCite seems to occasionally pass either blank XML (unsure of the cause) or use namespaces. This will let us tolerate parsing these.


nickgros:
closing to create PR based on current release branch

PR #3532, opened 2019-03-27
use new worker-utilities (PLFM-5451)

zimingd:


lib/stackConfiguration/src/main/resources/stack.properties
```diff
@@ -202,6 +202,9 @@ org.sagebionetworks.semaphore.gated.max.runners.annotations=1
 # The maximum number of workers in the cluster that will pop Table update messages from SQS
 org.sagebionetworks.semaphore.gated.max.runners.table.cluster=10
 
+#The maximum number of reader locks allowed at once for WriteReadSemaphoreRunner
+org.sagebionetworks.write.read.semaphore.runner.max.readers=8
```
john-hill, line 206: see the tableQueryQueueMessageReveiverTrigger.semaphoreMaxLockCount is the same number.

PR #3543, opened 2019-04-02
Plfm 5040

brucehoff:


lib/stackConfiguration/pom.xml
```diff
@@ -23,6 +24,13 @@
 			<groupId>org.apache.logging.log4j</groupId>
 			<artifactId>log4j-core</artifactId>
 		</dependency>
+		<dependency>
```
zimingd: the root pom.xml should have the dependency version version

lib/stackConfiguration/src/test/java/org/sagebionetworks/SynapseS3ClientImplTest.java
```diff
@@ -0,0 +1,59 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.junit.MockitoJUnitRunner;
+import org.sagebionetworks.aws.AwsClientFactory;
+import org.sagebionetworks.aws.CannotDetermineBucketLocationException;
+import org.sagebionetworks.aws.SynapseS3ClientImpl;
+
+import com.amazonaws.services.s3.model.Region;
+
+@RunWith(MockitoJUnitRunner.class)
```
zimingd: Remove. No mocks were used.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/ProjectSettingsManagerImpl.java
```diff
@@ -194,11 +193,8 @@ public void deleteProjectSetting(UserInfo userInfo, String id) throws DatastoreE
 		if (storageLocationSetting instanceof ExternalS3StorageLocationSetting) {
 			UserProfile userProfile = userProfileManager.getUserProfile(userInfo.getId().toString());
 			ExternalS3StorageLocationSetting externalS3StorageLocationSetting = (ExternalS3StorageLocationSetting) storageLocationSetting;
-			if (!StringUtils.isEmpty(externalS3StorageLocationSetting.getEndpointUrl())
-					&& !externalS3StorageLocationSetting.getEndpointUrl().equals("https://" + Constants.S3_HOSTNAME)) {
-				throw new NotImplementedException("Synapse does not yet support external S3 buckets in non-us-east-1 locations");
-			}
 			validateOwnership(externalS3StorageLocationSetting, userProfile);
+			validateBucketAccess(externalS3StorageLocationSetting);
```
zimingd, line 196: this should occur before `validateOwnership()` otherwise, `validateOwnership()` will call `s3client.getObject()`, which assumes US_EAST_1 when region not present, and get one of the errors at line 325 instead of a error telling users to check their bucket's permission.

lib/stackConfiguration/src/main/java/org/sagebionetworks/aws/SynapseS3ClientImpl.java
```diff
@@ -0,0 +1,220 @@
...
+	private Map<Region, AmazonS3> regionSpecificClients;
+
+	private Map<String, Region> bucketLocation;
+
+	public SynapseS3ClientImpl(Map<Region, AmazonS3> regionSpecificClients) {
+		this.regionSpecificClients=regionSpecificClients;
+		bucketLocation = Collections.synchronizedMap(new PassiveExpiringMap<>(1, TimeUnit.HOURS));
+	}
+	
+	public Region getRegionForBucket(String bucketName) {
```
john-hill, line 56: unit tests for all three test cases.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/ProjectSettingsManagerImpl.java
john-hill, line 196: also need a test

services/repository/src/test/java/org/sagebionetworks/file/controller/UploadControllerTest.java
```diff
@@ -38,6 +40,8 @@
 	private S3FileHandle handleOne;
 	private PreviewFileHandle handleTwo;
 	private List<String> toDelete;
+	
+	private static final String BUCKET_NAME = StackConfigurationSingleton.singleton().getS3Bucket();
```
john-hill: is this change still needed?

lib/stackConfiguration/pom.xml
brucehoff: good catch, thanks!

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/ProjectSettingsManagerImpl.java
brucehoff: That's a logical assumption, @zimingd but it's actually not true.  There is a bucket, "synodos.eu.frankfurt", which is in a European region but from which Synapse can list and read objects (files) using a us-east-1 S3 client.  However, Synapse cannot detect the region and will generate broken presigned URLs when using the us-east-1 S3 client.  So the checks about downloading file and about detecting the region are independent/orthogonal.

lib/stackConfiguration/pom.xml
brucehoff: By the way, the same problem was in the repository-managers pom.xml.  I fixed it there too.

services/repository/src/test/java/org/sagebionetworks/file/controller/UploadControllerTest.java
brucehoff: No, and this applies to three other tests which I've reverted as well.

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/ProjectSettingsManagerImplUnitlTest.java
```diff
@@ -0,0 +1,84 @@
...
+	private SynapseS3Client synapseS3Client;
+
+	@Mock
+	private StorageLocationDAO storageLocationDAO;
+
+	@Before
+	public void before() {
+		projectSettingsManagerImpl = new ProjectSettingsManagerImpl();
+	
+		ReflectionTestUtils.setField(projectSettingsManagerImpl, "userProfileManager", userProfileManager);
```
john-hill, line 46: If you use @InjectMocks you do not need to inject the mocks manually.

john-hill (approved):

PR #3575, opened 2019-04-26
added linear entity id parser with version support

john-hill:


lib/models/src/main/java/org/sagebionetworks/repo/model/entity/EntityId.java
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Immutable representation of an Entity's ID including an optional version
+ * number.
+ * Use {@linkplain EntityIdBuilder} to create new instances of this class.
+ * To parse an EntityId from a string use {@linkplain EntityIdParser}
+ */
+public class EntityId {
+
+	Long id;
```
john-hill: both fields can be final.

lib/models/src/main/java/org/sagebionetworks/repo/model/entity/EntityIdParser.java
```diff
@@ -0,0 +1,178 @@
...
+ */
+public class EntityIdParser {
+
+	// Null char is used to indicate parser termination.
+	private static final int NULL_CHAR = 0x0;
+
+	private int index;
+	private char[] chars;
+	private char currentChar;
+	EntityIdBuilder builder;
```
john-hill, line 16: did not use builder

lib/models/src/main/java/org/sagebionetworks/repo/model/entity/EntityIdParser.java
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	/**
+	 * Consume a single Long from the character array
+	 * 
+	 * @return The Long read from the array.
+	 */
+	private Long consumeLong() {
+		int start = index;
+		// consume all digits
```
john-hill, line 106: make this calculate the number
test cases syn.123 and syn.

zimingd (approved):

PR #3606, opened 2019-05-22
Restrict file names that are allowed for uploaded S3 filehandles to match restrictions on Entity names (PLFM-5510)

zimingd:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/jdo/EntityNameValidation.java
```diff
@@ -17,7 +17,7 @@
 
 	// match one or more whitespace characters
 	private static final Pattern ALLOWABLE_CHARS = Pattern
-			.compile(ModelConstants.VALID_ENTITY_NAME_REGEX);
```
brucehoff, line 20: put back in ModelConstants, otherwise the change will break the GWT portal

brucehoff: Comment the ModelConstants class, explaining that the purpose is to define constants used both in the back end and in the GWT portal.

lib/lib-shared-models/src/main/java/org/sagebionetworks/repo/model/util/ModelConstants.java
```diff
@@ -23,9 +23,6 @@
 import org.sagebionetworks.repo.model.ACCESS_TYPE;
 
 public class ModelConstants {
-	
-	public static final String VALID_ENTITY_NAME_REGEX = "^[a-z,A-Z,0-9,_,., ,\\-,\\+,(,)]+";
-
```
brucehoff: put back!

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/jdo/EntityNameValidation.java
```diff
@@ -32,13 +32,13 @@ public static String valdiateName(String key) {
 		key = key.trim();
 		if ("".equals(key))
 			throw new IllegalArgumentException(
-					"Entity names cannot be empty strings");
+					"Names cannot be empty strings");
```
brucehoff: Please see if we can treat nulls the same as empty and white space strings.

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/jdo/EntityNameValidation.java
```diff
@@ -32,13 +32,13 @@ public static String valdiateName(String key) {
 		key = key.trim();
 		if ("".equals(key))
 			throw new IllegalArgumentException(
-					"Entity names cannot be empty strings");
```
brucehoff, line 35: valdiateName -> validateName

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/file/FileHandleManagerImpl.java
```diff
@@ -1040,6 +1041,8 @@ public S3FileHandle createS3FileHandleCopy(UserInfo userInfo, String handleIdToC
 		ValidateArgument.requirement(StringUtils.isNotEmpty(fileName) || StringUtils.isNotEmpty(contentType),
 				"Either the fileName or the contentType needs to be set");
 
+		EntityNameValidation.valdiateName(fileName);
```
brucehoff: would it be less confusing to not use the name 'entity validation' if the check also applies to file names, which are not entity names?

brucehoff (approved):

PR #3618, opened 2019-05-29
PLFM-5265

nickgros:
Allow setting newVersion parameter when using SynapseClient.putEntity + docs changes/deleting unused files


lib/lib-javadoc/src/main/resources/bannerHeader.html
```diff
@@ -1,4 +1,4 @@
-<header style="background: none repeat scroll 0% 0% rgb(238, 238, 238);">
+<header style="background: none repeat scroll 0% 0% rgb(43, 104, 143);">
```
nickgros, line 1: Change color of rest docs header background to Synapse's blue

john-hill (approved):

PR #3625, opened 2019-06-04
Revert "Plfm 5570"

xschildw:
I think we discussed getting the reconciler out.


xschildw:
If OK, I can merge develop into 268 before getting out of RO mode.

zimingd:
#3628 already did this in 268 I think. we just need to merge 268 to develop

PR #3640, opened 2019-06-10
limit search document upload to once every 10 seconds (PLFM-5570)

zimingd:


lib/lib-search/src/main/java/org/sagebionetworks/search/CloudsSearchDomainClientAdapter.java
```diff
@@ -69,14 +72,19 @@ public void sendDocuments(Iterator<Document> documents){
...
 			} catch (DocumentServiceException e) {
 				logger.error("The following documents failed to upload: " +  documentIds);
 				documentIds = null;
 				updateAndSendKinesisLogRecords(e.getStatus());
 				throw handleCloudSearchExceptions(e);
 			} catch (IOException e){
 				throw new TemporarilyUnavailableException(e);
-			}finally{
+			} catch (InterruptedException e) {
+				logger.error("sleep was interrupted");
```
john-hill: Recommend terminating on IterrupteedException.  It can occur with the graceful shutdown of an process.

john-hill (approved):

PR #3641, opened 2019-06-12
Plfm 4890 i

john-hill:


lib/lib-table-query/src/main/resources/table-query-parser.jj
```diff
@@ -859,9 +871,11 @@ BooleanFunction booleanFunction():
 ValueExpression valueExpression():
 {
   	NumericValueExpression numericValueExpression = null;
+  	EntityId entityId = null;
 }
 {
 	numericValueExpression = numericValueExpression() { return new ValueExpression(numericValueExpression); }
+	| entityId = entityId() { return new ValueExpression(entityId); }
```
john-hill: This might be the wrong spot? Currently  a column contains "syn123".  We found at least one case where a column name contains 'syn<ID>'

lib/lib-table-query/src/test/java/org/sagebionetworks/table/query/EntityIdTest.java
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	@Test
+	public void testEntityIdWithVersion() throws ParseException {
+		EntityId value = new TableQueryParser("syn123.456").entityId();
+		assertEquals("syn123.456", value.toSql());
+	}
+
+	@Test
+	public void testEntityIdTooManyDots() throws ParseException {
```
john-hill: move to the query parser to ensure this fails.

lib/lib-table-query/src/test/java/org/sagebionetworks/table/query/EntityIdTest.java
```diff
@@ -0,0 +1,55 @@
...
+		EntityId value = new TableQueryParser("syn123.456.222").entityId();
+		assertEquals("syn123.456", value.toSql());
+	}
+
+	@Test
+	public void testIdAndVersionDotOnly() throws ParseException {
+		EntityId value = new TableQueryParser("syn123.").entityId();
```
john-hill, line 41: same for this test

zimingd (approved):

PR #3650, opened 2019-06-20
PLFM-4899 Deprecated Query Services

kimyen:


xschildw (approved):

integration-test/src/test/java/org/sagebionetworks/IT500SynapseJavaClient.java
```diff
@@ -1099,4 +919,39 @@ public void testGetEntityChildrenProjects() throws SynapseException{
...
 		assertEquals(dataset.getId(), synapseOne.lookupChild(project.getId(), dataset.getName()));
 	}
+	
+	
+	/*
+	 * Test deprecated services
+	 * These test can be removed when the services are removed
+	 */
+	
+	@Test
```
jay-hodgson, line 929: Note that you could use the annotation instead in this case.
Like:
@Test (expected= SynapseDeprecatedServiceException.class)

PR #3653, opened 2019-06-27
PLFM-5613

nickgros:


lib/communicationUtilities/src/main/java/org/sagebionetworks/utils/MD5ChecksumHelper.java
```diff
@@ -19,6 +20,8 @@
  */
 public class MD5ChecksumHelper {
 
+	private static final Pattern VALID_MD5_REGEX = Pattern.compile("^[0-9a-fA-F]{32}$");
+
```
nickgros, line 24: Try to get rid of this class

john-hill:
Please have this reviewed or close.

PR #3652, opened 2019-06-25
PLFM-5604 google cloud storage client code

nickgros:

PR #3667, opened 2019-07-09
PLFM-5666 rest docs fix 

nickgros:
fixes an error slipped in with rest docs change in #3652 where a custom CSS rule was applied to a class that referred to more elements than intended


lib/lib-javadoc/src/main/resources/controllersHtmlTemplate.html
```diff
@@ -7,7 +7,7 @@
 		<div id="main">
 			<div id="mainWrapper" class="ym-wrapper">
 				<div class="ym-wbox">
-					<div id="rootPanel" class="ym-grid linearize-level-1">
+					<div id="rootPanel" class="ym-grid linearize-level-1 allServicesList">
```
nickgros, line 10: styling that was applied to #rootPanel (which incorrectly applied to elements on individual service pages) is now applied to div.allServicesList, which only occurs here

zimingd (approved):

PR #3668, opened 2019-07-10
override cloudmainin character encoding to utf-8 (PLFM-5664)

zimingd:


services/repository/src/main/java/org/sagebionetworks/auth/CloudMailInCharacterEncodingFilter.java
```diff
@@ -0,0 +1,39 @@
...
+	public void init(FilterConfig filterConfig) throws ServletException {
+		//nothing to do
+	}
+
+	@Override
+	public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws IOException, ServletException {
+		chain.doFilter(new HttpServletRequestWrapper((HttpServletRequest) request) {
+			@Override
+			public String getCharacterEncoding() {
+				return "utf-8";
```
brucehoff, line 30: Do you want to override the character set in every request or just when it's not specified by the client?

john-hill (approved):

services/repository/src/main/java/org/sagebionetworks/auth/CloudMailInCharacterEncodingFilter.java
zimingd: We have no idea whether CloudMailIn is specifying a charset or not when they call our API. To be safe and avoid having to wait another 2 weeks for the changes to go in, I always set the charset as utf-8. Since UTF-8 is a superset of both ASCII and ISO-8859-15, there is no risk in having Spring read characters from those charsets as UTF-8. This filter is also restricted to the single CloudMailIn API call so impact is limited.

PR #3676, opened 2019-07-15
ignore types that have been removed

john-hill:


xschildw (approved):
Looks good to me. John's build passed so merging to get going...

PR #3717, opened 2019-08-23
5604 review changes b

nickgros:


integration-test/src/test/java/org/sagebionetworks/IT049FileHandleTest.java
```diff
@@ -75,8 +84,15 @@
 	private File imageFile;
 	private Project project;
 
+	// Hard-coded dev test bucket
+	private String googleCloudBucket = "dev.test.gcp-storage.sagebase.org";
```
nickgros, line 88: Best/correct way to do this? Used spring in an autowired test but seems to not be the pattern in integration tests

integration-test/src/test/java/org/sagebionetworks/ITV2WikiPageTest.java
```diff
@@ -111,9 +111,9 @@ public void before() throws SynapseException, IOException {
 		list.add(imageFile);
 		list.add(imageFileTwo);
 		list.add(markdownFile);
-		fileHandle = synapse.multipartUpload(imageFile, null, true, false);
-		fileHandleTwo = synapse.multipartUpload(imageFileTwo, null, true, false);
-		markdownHandle = synapse.multipartUpload(markdownFile, null, false, false);
+		fileHandle = (S3FileHandle) synapse.multipartUpload(imageFile, null, true, false);
+		fileHandleTwo = (S3FileHandle) synapse.multipartUpload(imageFileTwo, null, true, false);
+		markdownHandle = (S3FileHandle) synapse.multipartUpload(markdownFile, null, false, false);
```
john-hill, line 116: Can these be changed to cloudprovider?

john-hill (approved):

PR #3726, opened 2019-09-05
PLFM-4590 updates following code review

brucehoff:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/oauth/OIDCTokenHelperImpl.java
```diff
@@ -144,52 +121,45 @@ public String createOIDCaccessToken(
...
+		if (pieces.length!=3) throw new IllegalArgumentException("Expected three sections of the token but found "+pieces.length);
 		String unsignedToken = pieces[0]+"."+pieces[1]+".";
 		JsonWebKey matchingKey=null;
 		{
 			Jwt<Header,Claims> unsignedJwt = Jwts.parser().parseClaimsJwt(unsignedToken);
+			String keyId = (String)unsignedJwt.getHeader().get(JwsHeader.KEY_ID);
 			for (JsonWebKey jwk : jsonWebKeySet.getKeys()) {
-				if (jwk.getKid().equals(unsignedJwt.getHeader().get("kid"))) {
+				if (jwk.getKid().equals(keyId)) {
 					matchingKey = jwk;
```
marcomarasca, line 138: Missing break? An alternative is to replace the for loop with a stream filter expression, e.g:
jsonWebKeySet.getKeys().stream().filter(key -> key.getKid().equals(keyId)).findFirst(). Note that it returns an Optional, which can be used later on with ifPresent().

marcomarasca (approved):
Looks great, thanks! Only comment about the missing break in the OIDCTokenHelperImpl, after line 138.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/oauth/OIDCTokenHelperImpl.java
brucehoff: I added in the 'break'

PR #3727, opened 2019-09-05
follow up to PLFM-4590 DAO code review

brucehoff:


brucehoff:
Requested changes from previous code review:
- shorten table column names
- avoid retrieving DBO in update method
- no need to get CMS key ID for key alias, since alias can be passed to encryption service

john-hill (approved):

PR #3736, opened 2019-09-12
PLFM-5767: Rename athena query output location

marcomarasca:
..in order to avoid interference with the log collator. This is a cherry picked commit.


marcomarasca:
Will merge the other PR

PR #3752, opened 2019-09-17
From stream to athena integration test

marcomarasca:
I added a complete integration test from sending events to the stream, to querying S3 with athena and saving the results. Note this should looked at after PR #3751 is merged.


marcomarasca:
Cherry picked in another branch for clarity

PR #3785, opened 2019-10-04
PLFM-5842: Follow up, throws instead of logging

marcomarasca:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/principal/PrincipalManagerImpl.java
```diff
@@ -241,7 +231,14 @@ public PrincipalAliasResponse lookupPrincipalId(PrincipalAliasRequest request) {
 		return response;
 	}
 	
-	private EmailQuarantineStatus getQuarantineStatus(QuarantinedEmail quarantinedEmail) {
+	private void assertNotQuarantinedEmail(String email, String messageType) {
```
john-hill, line 234: validate

john-hill (approved):

PR #3793, opened 2019-10-14
PLFM-5882: Added id to quarantined email table

marcomarasca:
..so that we can migrate the table


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/migration/MigratableTableDAOImpl.java
```diff
@@ -338,8 +340,20 @@ public String mapRow(ResultSet rs, int rowNum)
 					+ mapping.getTableName() + ":");
 			log.debug("\t" + names.toString());
 		}
+
```
john-hill, line 343: will add a test to the autowired test

PR #3809, opened 2019-10-25
Plfm 5170 ii

brucehoff:
Previously there were three modes of authentication handled by the AuthenticationFilter:
1) sessionToken
2) HMAC
3) anonymous (no authentication header(s))

For each mode a downstream Controller would see a userId as follows:
sessionToken:	userId
HMAC:		userId
anonymous:	anon-user-id

We remove OAuthAccessTokenFilter, moving the logic into AuthenticationFilter and updating web.xml to apply AuthenticationFilter to the userinfo services.  The reason for combining the different kinds of logic into one filter (instead of using multiple filters) is to reduce the chance of misconfiguring auth filter(s) that secure our services.

With this change, there are four variations allowed for authenticating requests:
1) sessionToken
2) HMAC
3) anonymous
4) oauth access token

With this change, a Controller sees not only the current 'userId' request param but also a Authorization header of type Bearer:
sessionToken:			userId; 			full access token
HMAC:				userId; 			full access token
anonymous:			anon-user-id; 		anon-access token
oauth access token:	NO user id;		scoped access token

This means that all current services can continue to use 'userId' to determine authorization.  Omitting userId when authenticating via access token means that current services won't accidentally allow unauthorized access when a scoped token is passed.
Services can switch from userId to access token when ready, an will have the correctly scoped token, regardless of how the request was authenticated.

Currently at the service level the 'userId' is exchanged for a UserInfo object, containing the userId and teams the user is in.  We introduce the UserAuthorization object which translates an access token into (1) the UserInfo, (2) the scopes granted by the access token and (3) the OIDC claims from the access token.  This can then be passed to downstream managers to use as they use UserInfo today. This PR updates the User Info service to follow this pattern. (See OpenIDConnectServiceImpl, OpenIDConnectManagerImpl.)

OIDCTokenHelper has methods to create full-access and anonymous-access tokens.


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/oauth/OIDCTokenHelperImpl.java
```diff
@@ -118,6 +122,18 @@ public String createOIDCaccessToken(
 		return createSignedJWT(claims);
 	}
 	
+	@Override
+	public String createTotalAccessToken(Long principalId) {
+		String issuer = null; // doesn't matter -- it's only important to the client (which will never see this token, used internally)
+		String subject = principalId.toString(); // we don't encrypt the subject
+		String oauthClientId = ""+AuthorizationConstants.SYNAPSE_OAUTH_CLIENT_ID;
```
brucehoff, line 129: what if clientId=0 is used elsewhere, like making a authorization request?

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/oauth/OpenIDConnectManagerImpl.java
```diff
@@ -204,15 +211,27 @@ public OAuthAuthorizationResponse authorizeClient(UserInfo userInfo,
 
 	// As per, https://openid.net/specs/openid-connect-core-1_0.html#PairwiseAlg
 	public String ppid(String userId, String clientId) {
+		if (AuthorizationConstants.SYNAPSE_OAUTH_CLIENT_ID.equals(clientId)) {
```
brucehoff, line 217: add a comment explaining why we don't encrypt / decrypt

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/oauth/OpenIDConnectManagerImplUnitTest.java
```diff
@@ -705,5 +764,74 @@ public void testGetUserInfoAsJWT() {
...
+		when(mockOauthClientDao.getSectorIdentifierSecretForClient(OAUTH_CLIENT_ID)).thenReturn(clientSpecificEncodingSecret);
+		
+		String ppid = openIDConnectManagerImpl.ppid(USER_ID, OAUTH_CLIENT_ID);
+		claims.setSubject(ppid);
+
+		// method under test
+		assertEquals(USER_ID, openIDConnectManagerImpl.getUserId(token));
+		
+		verify(oidcTokenHelper).parseJWT(token);
+	}
```
brucehoff, line 849: add a test to getUserInfo when client ID is Synapse ("0")

services/repository/src/main/java/org/sagebionetworks/auth/AuthenticationFilter.java
```diff
@@ -46,126 +43,111 @@
...
-		if (userId == null && !allowAnonymous) {
-			String reason = "The session token provided was missing, invalid or expired.";
-			reject(req, (HttpServletResponse) servletResponse, reason);
-			log.warn("Anonymous not allowed");
-			return;
-		}
-		
+
 		// If the user has been identified, check if they have accepted the terms of use
 		if (userId != null) {
```
brucehoff: whenever we have a userId we also have a bearer token, so eliminate the case: if (userId!=null)

services/repository/src/main/java/org/sagebionetworks/auth/services/AuthenticationService.java
```diff
@@ -68,11 +68,17 @@
 	 * @throws {@link UnauthorizedException} If the alias belongs to a team.
 	 */
 	public PrincipalAlias lookupUserForAuthentication(String alias);
+	
 	/**
 	 * Has the user accepted the terms of use?
 	 */
 	public boolean hasUserAcceptedTermsOfUse(Long userId) throws NotFoundException;
```
brucehoff: remove?

services/repository/src/main/java/org/sagebionetworks/auth/services/AuthenticationService.java
```diff
@@ -68,11 +68,17 @@
...
 
+	/**
+	 * Has the user accepted the terms of use?
+	 */
+	public boolean hasUserAcceptedTermsOfUse(String accessToken) throws NotFoundException;
```
brucehoff, line 75: be consistent in variable name (bearer token, jwt token or access token)

services/repository/src/main/java/org/sagebionetworks/auth/AuthenticationFilter.java
```diff
@@ -46,126 +43,111 @@
...
 		try {
 			// Pass along, including the user ID
 			Map<String, String[]> modParams = new HashMap<String, String[]>(req.getParameterMap());
-			modParams.put(AuthorizationConstants.USER_ID_PARAM, new String[] { userId.toString() });
-			HttpServletRequest modRqst = new ModHttpServletRequest(req, null, modParams);
+			if (userId!=null) {
+				modParams.put(AuthorizationConstants.USER_ID_PARAM, new String[] { userId.toString() });
+			}
+			Map<String, String[]> modHeaders = HttpAuthUtil.filterAuthorizationHeaders(req);
+			HttpAuthUtil.setBearerTokenHeader(modHeaders, bearerToken);
```
brucehoff: set not as Authorization header but as a Synapse-specific header name so as not to collide with Basic auth header used in special purpose filters

services/repository/src/main/webapp/WEB-INF/web.xml
```diff
@@ -80,28 +80,33 @@
 		<url-pattern>/*</url-pattern>
 	</filter-mapping>
 
+	<!--  Any filters depending on Authorization header should go before the general authFilter -->
```
brucehoff: undo reordering  of filters

services/repository/src/test/java/org/sagebionetworks/auth/AuthenticationFilterTest.java
```diff
@@ -45,7 +58,24 @@
 @ExtendWith({MockitoExtension.class})
 @MockitoSettings(strictness = Strictness.LENIENT)
```
brucehoff: remove lenient

marcomarasca (approved):
Looks great, thanks!

PR #3816, opened 2019-10-29
PLFM-5913: Async message processing

marcomarasca:
Always process messages in a worker independently from the number of recipients, before we would have a special case when there was a single recipient and the message was sent right away. We changed instead the system to always process the messages in a worker. Failures are sent as a notification to the sender.

### DIFF
--- a/services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/MessageManager.java
+++ b/services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/MessageManager.java
@@ -122,7 +122,7 @@ public interface MessageManager {
 	public void sendPasswordChangeConfirmationEmail(long userId);
 	
 	/**
-	 * Sends a delivery failure notification based on a template
+	 * Sends a delivery failure notification based on a template unless the message is a notification message
 	 */
 	public void sendDeliveryFailureEmail(String messageId, List<String> errors) throws NotFoundException;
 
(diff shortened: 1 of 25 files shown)


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/MessageManager.java
```diff
@@ -131,11 +127,7 @@ public void sendNewPasswordResetEmail(String passwordResetPrefix, PasswordResetS
 	public void sendDeliveryFailureEmail(String messageId, List<String> errors) throws NotFoundException;
```
john-hill, line 127: Marco wants to not send delivery failures for "notification emails" as those "should" be system auto-sent.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/MessageManagerImpl.java
```diff
@@ -456,12 +405,8 @@ public void markMessageStatus(UserInfo userInfo, MessageStatus status) throws No
 					settings = new Settings();
 				}
 				MessageStatusType userMessageStatus = null; // setting to null tells the DAO to use the default value
-				// This marks a user as a recipient of the message
-				if (singleTransaction) {
-					messageDAO.createMessageStatus_SameTransaction(dto.getId(), userId, userMessageStatus);
```
john-hill, line 461: can remove same transaction for both create and update.

john-hill (approved):

PR #3821, opened 2019-10-30
PLFM-5913: Allows submission creation if notification fails

marcomarasca:
Temporary fix for staging so that this can be released next week.

### DIFF
--- a/services/repository/src/main/java/org/sagebionetworks/repo/web/service/EvaluationServiceImpl.java
+++ b/services/repository/src/main/java/org/sagebionetworks/repo/web/service/EvaluationServiceImpl.java
@@ -4,6 +4,8 @@ import java.util.List;
 
 import javax.servlet.http.HttpServletRequest;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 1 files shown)


services/repository/src/main/java/org/sagebionetworks/repo/web/service/EvaluationServiceImpl.java
```diff
@@ -170,10 +174,17 @@ public Submission createSubmission(Long userId, Submission submission, String en
 		Long versionNumber = submission.getVersionNumber();
 		EntityBundle bundle = serviceProvider.getEntityBundleService().getEntityBundle(userId, entityId, versionNumber, mask);
 		Submission created = submissionManager.createSubmission(userInfo, submission, entityEtag, submissionEligibilityHash, bundle);
-		List<MessageToUserAndBody> messages = submissionManager.
-				createSubmissionNotifications(userInfo,created,submissionEligibilityHash,
-						challengeEndpoint, notificationUnsubscribeEndpoint);
-		notificationManager.sendNotifications(userInfo, messages);
+		
+		try {
```
john-hill, line 178: We agree this is a "hack" but is needed because the "real" fix is too risky to put in staging.

john-hill (approved):

PR #3829, opened 2019-11-05
Plfm 4741

brucehoff:


brucehoff:
The root cause of the bug was that the sort column was joined using an *inner* join when it should have been a *left outer* join.  I rewrote the queries using 'join ... on' syntax and introducing the left outer join so that the original query:

```
SELECT annoBlob.ANNOTATIONS_BLOB 
FROM SUBSTATUS_ANNOTATIONS_OWNER annoOwner  
inner join  SUBSTATUS_ANNOTATIONS_BLOB annoBlob  
inner join  SUBSTATUS_STRINGANNOTATION srt 
WHERE annoOwner.EVALUATION_ID=42 
and annoOwner.SUBMISSION_ID = annoBlob.SUBMISSION_ID 
and annoOwner.SUBMISSION_ID = srt.SUBMISSION_ID 
and srt.ATTRIBUTE = 'string anno_null'
ORDER BY srt.VALUE asc ;
```

becomes:

```
SELECT annoBlob.ANNOTATIONS_BLOB 
FROM SUBSTATUS_ANNOTATIONS_OWNER annoOwner  
inner join  SUBSTATUS_ANNOTATIONS_BLOB annoBlob on annoOwner.SUBMISSION_ID = annoBlob.SUBMISSION_ID  
left outer join  SUBSTATUS_STRINGANNOTATION srt on annoOwner.SUBMISSION_ID = srt.SUBMISSION_ID and srt.ATTRIBUTE = 'string anno_null'
WHERE annoOwner.EVALUATION_ID=42 
ORDER BY srt.VALUE asc ;

```

john-hill (approved):

PR #3837, opened 2019-11-09
unconditionally fail table builds on any error

john-hill:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/table/TableStatusDAOImpl.java
```diff
@@ -135,7 +138,7 @@ private void attemptToSetTableEndState(IdAndVersion idAndVersion,
 			throw new IllegalArgumentException("This method cannot be used to change the state to PROCESSING because it does not change the reset-token");
 		}
 		DBOTableStatus current = selectResetTokenForUpdate(idAndVersion);
-		if(!current.getResetToken().equals(resetToken)) {
+		if(resetToken != null && !current.getResetToken().equals(resetToken)) {
```
xschildw, line 141: Need to update the token

PR #3841, opened 2019-11-13
Plfm 5759

marcomarasca:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/oauth/OpenIDConnectManagerImpl.java
```diff
@@ -391,4 +408,51 @@ public Object getUserInfo(UserAuthorization userAuthorization, String oauthClien
...
+		if (oauthClientDao.isOauthClientVerified(clientId)) {
+			return;
+		}
+		
+		StringBuilder errorMessage = new StringBuilder("The client is not verified yet.");
+
+		if (includeACTMessageCondition.getAsBoolean()) {
+			String actEmail = getVerificationContactEmail();
+			if (actEmail != null) {
+				errorMessage.append(" Pleast contact the Synapse Access and Compliance Team (" + actEmail + ") to verify your client.");
```
marcomarasca: Maybe just point to https://docs.synapse.org/articles/using_synapse_as_an_oauth_server.html instead of including the ACT email.

lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/oauth/OAuthClient.json
```diff
@@ -30,7 +30,7 @@
 		},
 		"verified": {
 			"type": "boolean",
-			"description": "initially false, becomes true once the client creator demonstrates ownership of the redirect URI host",
+			"description": "initially false, to be verified please contact the AC Team.",
```
marcomarasca: Maybe add link to https://docs.synapse.org/articles/using_synapse_as_an_oauth_server.html

services/repository/src/main/java/org/sagebionetworks/auth/OpenIDConnectController.java
```diff
@@ -74,7 +74,7 @@ JsonWebKeySet getOIDCJsonWebKeySet() {
 	
 	/**
 	 * Create an OAuth 2.0 client.  Note:  After creating the client one must also set the client secret
-	 * and verify ownership of the host(s) in the registered redirect URIs.
+	 * and ask ACT to have their client verified.
```
marcomarasca: Maybe add link to https://docs.synapse.org/articles/using_synapse_as_an_oauth_server.html

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/oauth/OAuthClientManagerImpl.java
```diff
@@ -224,7 +229,29 @@ public OAuthClient updateOpenIDConnectClient(UserInfo userInfo, OAuthClient toUp
...
+		OAuthClient currentClient = oauthClientDao.selectOAuthClientForUpdate(clientId);
+		
+		if (verifiedStatus != BooleanUtils.isTrue(currentClient.getVerified())) {
+			currentClient.setVerified(verifiedStatus);
+			currentClient.setModifiedOn(new Date());
+			currentClient.setEtag(UUID.randomUUID().toString());
+			oauthClientDao.updateOAuthClient(currentClient);
+		}
 
+		return oauthClientDao.getOAuthClient(clientId);
```
brucehoff: why can't you just return currentClient and avoid another database call?

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/oauth/OpenIDConnectManagerImpl.java
```diff
@@ -86,7 +93,7 @@ public void setClaimProviders(Map<OIDCClaimName, OIDCClaimProvider> claimProvide
 	 */
 	public static List<OAuthScope> parseScopeString(String s) {
 		if (StringUtils.isEmpty(s)) {
-			return Collections.EMPTY_LIST;
+			return Collections.emptyList();
```
brucehoff, line 89: Please consider throwing a 4xx status in getAuthenticationRequestDescription() if the client is unverified

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/auth/OAuthClientDaoImpl.java
```diff
@@ -222,8 +221,19 @@ public boolean checkOAuthClientSecretHash(String clientId, String secretHash) {
...
+		String secretHash;
+		
+		try {
+			secretHash = jdbcTemplate.queryForObject(CLIENT_SECRET_HASH_SQL_SELECT, String.class, clientId);
+		} catch (EmptyResultDataAccessException e) {
+			throw clientNotFoundException(clientId);
+		}
+		
+		if (secretHash == null) {
+			throw new NotFoundException("OAuth client (" + clientId + ") not initialized");
```
brucehoff: This error message is an improvement but I'd go further and replace 'not initialized' with something like 'does not have a secret defined'

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/oauth/OAuthClientNotVerifiedException.java
```diff
@@ -0,0 +1,9 @@
+package org.sagebionetworks.repo.manager.oauth;
+
+public class OAuthClientNotVerifiedException extends RuntimeException {
```
brucehoff: Will this become a 4xx or a 5xx status?

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/oauth/OpenIDConnectManagerImpl.java
```diff
@@ -391,4 +406,43 @@ public Object getUserInfo(UserAuthorization userAuthorization, String oauthClien
...
+	 */
+	private void validateClientVerificationStatus(String clientId, BooleanSupplier includeOAuthDocsURLCondition) throws NotFoundException, OAuthClientNotVerifiedException {
+		
+		if (oauthClientDao.isOauthClientVerified(clientId)) {
+			return;
+		}
+		
+		StringBuilder errorMessage = new StringBuilder("The client is not verified yet.");
+
+		if (includeOAuthDocsURLCondition.getAsBoolean()) {
```
brucehoff: I feel that the response should behave the same regardless of who the caller/user is.   In general when we try to anticipate user behavior and do something unexpected, we can end up causing trouble.  E.g., if the developer uses two Synapse accounts for testing, our check will not identify the developer.  If the developer learns to expect one error message and then gets a modified one, they may wonder why.  I think it's sufficient to say that the reason the call failed is that "The client is not verified".  The user who set up the client should have read the set-up instructions and/or will refer back to the instructions when they see this message, and the instructions will tell them what to do next.

brucehoff (approved):
LOoks good.

PR #3843, opened 2019-11-14
PLFM-5613

nickgros:
https://sagebionetworks.jira.com/browse/PLFM-5613

### DIFF
--- a/lib/communicationUtilities/src/main/java/org/sagebionetworks/utils/MD5ChecksumHelper.java
+++ b/lib/communicationUtilities/src/main/java/org/sagebionetworks/utils/MD5ChecksumHelper.java
@@ -5,29 +5,36 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.UnsupportedEncodingException;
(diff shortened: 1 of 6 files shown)


lib/communicationUtilities/src/main/java/org/sagebionetworks/utils/MD5ChecksumHelper.java
```diff
@@ -1,113 +1,14 @@
...
 
-/**
- * Code lifted from http://www.rgagnon.com/javadetails/java-0416.html
- * 
- * Found similar code later in org/apache/commons/codec/digest/DigestUtils.java
- * 
- * @author deflaux
- * 
- */
 public class MD5ChecksumHelper {
```
nickgros, line 15: This class no longer handles calculating MD5 checksums (correctness of the calculation was  untested). Calculations are now delegated (via this class) to Apache's DigestUtils.

nickgros: Probably Rename to MD5ChecksumUtil

brucehoff (approved):
Looks great!

PR #3849, opened 2019-11-22
PLFM-5970

marcomarasca:


integration-test/src/test/java/org/sagebionetworks/IT500SynapseJavaClientTeamTest.java
```diff
@@ -650,31 +650,33 @@ public void testMembershipInvitationAPI() throws Exception {
...
+//		assertEquals(inviteeId, mis.getInviteeId());
+//
+//		// delete the second invitation
+//		synapseOne.deleteMembershipInvitation(mis.getId());
+//		try {
+//			synapseOne.getMembershipInvitation(mis.getId());
+//			fail("Failed to delete membership invitation.");
+//		} catch (SynapseException e) {
+//			// as expected
+//		}
```
brucehoff, line 679: We do not have commented out / dead code in our code base.  The lines should be removed altogether.

PR #3855, opened 2019-12-03
Plfm 5965

john-hill:


lib/jdomodels/src/main/resources/schema/functions/GetEntityPath.ddl.sql
```diff
@@ -0,0 +1,35 @@
+CREATE FUNCTION `getEntityPath`(inputEntityId BIGINT) RETURNS varchar(1000) CHARSET utf8mb4
```
marcomarasca: An alternative to consider is to use a recursive cte (See https://dev.mysql.com/doc/refman/8.0/en/with.html#common-table-expressions-recursive), something on the lines of:

WITH RECURSIVE PATH(ID, PARENT_ID, DISTANCE) AS (
-- Initial row set, replace ? with the id of the entity
SELECT ID, PARENT_ID, 1
FROM JDONODE WHERE ID = ?
-- Distinct allows loop detection as well
UNION DISTINCT
-- Recursion
SELECT JDONODE.ID, JDONODE.PARENT_ID, PATH.DISTANCE+ 1
FROM JDONODE JOIN PATH ON (JDONODE.ID = PATH.PARENT_ID)
-- Break point
WHERE JDONODE.ID IS NOT NULL AND DISTANCE <= 100
)
SELECT * FROM PATH ORDER BY DISTANCE;

john-hill: Interesting.
I setup a test on a hierarchy 12 deep to compare this query to the function and they have identical run times (not surprising).  I think the main advantage of the "WITH RECURSIVE" approach is we can select additional columns like name and type.  With the function, I need a second call to gather name and type for the results.

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/NodeDAOImpl.java
```diff
@@ -1033,10 +1067,9 @@ public boolean doesNodeRevisionExist(String nodeId, Long revNumber) {
 	}
 
 	@Override
-	public EntityHeader getEntityHeader(String nodeId, Long versionNumber) throws DatastoreException, NotFoundException {
+	public EntityHeader getEntityHeader(String nodeId) throws DatastoreException, NotFoundException {
 		Reference ref = new Reference();
 		ref.setTargetId(nodeId);
-		ref.setTargetVersionNumber(versionNumber);
```
marcomarasca, line 1039: At line 1073 below if(ref.getTargetVersionNumber() != null) block should probably be removed together with the EntityHeader clone = SerializationUtils.cloneJSONEntity(original); one line above.

lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/dao/NodeDaoUnitTest.java
```diff
@@ -5,6 +5,7 @@
 import static org.junit.Assert.assertTrue;
 import static org.mockito.Mockito.verify;
 
+import java.util.Collections;
```
marcomarasca, line 8: Organize imports for the whole codebase next PR :)

lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/table/QueryResultBundle.json
```diff
@@ -39,9 +39,13 @@
 				"$ref": "org.sagebionetworks.repo.model.table.FacetColumnResult"
 			}
 		},
-		"sumFileSizes":{
-			"description":"The sum of the file size for all files in the given view query. Use mask = 0x40 to include in the bundle.",
+		"sumFileSizes": {
+			"description": "The sum of the file size for all files in the given view query. Use mask = 0x40 to include in the bundle.",
 			"$ref": "org.sagebionetworks.repo.model.table.SumFileSizes"
+		},
+		"estimatedUpToDate": {
```
marcomarasca, line 46: Maybe rethink this name, it's a tiny bit confusing. What does it mean if the up to date estimate is false? Maybe an indexStatus is enough? Or **isStale**?

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/table/TableManagerSupportImpl.java
```diff
@@ -535,13 +535,8 @@ public Long getViewTypeMask(IdAndVersion idAndVersion){
...
 
 	@Override
-	public Set<Long> getEntityPath(IdAndVersion idAndVersion) {
-		List<EntityHeader> headers = nodeDao.getEntityPath(idAndVersion.getId().toString());
-		Set<Long> results = new HashSet<Long>(headers.size());
-		for(EntityHeader header: headers){
-			results.add(KeyFactory.stringToKey(header.getId()));
-		}
-		return results;
+	public List<Long> getEntityPath(IdAndVersion idAndVersion) {
```
marcomarasca, line 538: getEntityPathIds

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/AuthorizationManagerImplUnitTest.java
```diff
@@ -697,31 +697,24 @@ private static void addEntityHeaderTo(String id, Collection<EntityHeader>c) {
...
-		}
-		when(mockNodeDao.getEntityPath(newParentId)).thenReturn(newParentAncestors);
+		List<Long> newAncestorIds = new ArrayList<Long>();
+		newAncestorIds.add(KeyFactory.stringToKey(newParentId));
+		newAncestorIds.add(888L);
+		when(mockNodeDao.getEntityPathIds(newParentId)).thenReturn(newAncestorIds);
 		
 		List<String> diff = Arrays.asList("1");
 		when(mockAccessRequirementDAO.getAccessRequirementDiff(ancestorIds, newAncestorIds, RestrictableObjectType.ENTITY)).thenReturn(new LinkedList<String>());
+		when(mockNodeDao.isNodeAvailable(parentId)).thenReturn(true);
```
marcomarasca, line 713: Can be removed?

services/repository/src/main/java/org/sagebionetworks/repo/web/service/EntityServiceImpl.java
```diff
@@ -483,9 +483,9 @@ public boolean hasAccess(String entityId, Long userId, String accessType)
 	}
 
 	@Override
-	public EntityHeader getEntityHeader(Long userId, String entityId, Long versionNumber) throws NotFoundException, DatastoreException, UnauthorizedException {
+	public EntityHeader getEntityHeader(Long userId, String entityId) throws NotFoundException, DatastoreException, UnauthorizedException {
 		UserInfo userInfo = userManager.getUserInfo(userId);
-		return entityManager.getEntityHeader(userInfo, entityId, versionNumber);
+		return entityManager.getEntityHeader(userInfo, entityId);
```
marcomarasca, line 488: Double check if the API is broken

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/NodeDAOImpl.java
john-hill: Verified that there is no path from the API to call this method with a non-null version.

marcomarasca (approved):

PR #3863, opened 2019-12-10
PLFM-5773: Defaults to true if inclusion flags are not specified for stats

marcomarasca:


marcomarasca:
Mixup with the PLFM number

PR #3866, opened 2019-12-11
error message fix for Entity views (PLFM-5999)

zimingd:


lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/model/QualifiedJoin.java
```diff
@@ -14,12 +14,7 @@
...
-	JoinCondition joinCondition;// optional
-
-
-	QualifiedJoin(TableReference tableReferenceLHS, TableReference tableReferenceRHS) {
-		this(tableReferenceLHS,tableReferenceRHS,null);
-	}
+	JoinCondition joinCondition;// condition for join
 
 	public QualifiedJoin(TableReference tableReferenceLHS, TableReference tableReferenceRHS, JoinCondition joinCondition) {
 		this.tableReferenceLHS = tableReferenceLHS;
```
john-hill, line 20: throw exception

john-hill (approved):

PR #3874, opened 2019-12-19
PLFM-5932: Allows deletion of nodes with more than 15 levels of depth

marcomarasca:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/NodeDAOImpl.java
```diff
@@ -1575,11 +1601,51 @@ public ProjectHeader mapRow(ResultSet rs, int rowNum) throws SQLException {
...
+		ValidateArgument.required(parentIds, "parentIds");
+		ValidateArgument.requirement(minDistance > 0, "The min distance must be greater than 0");
+		
+		Map<String, Object> parameters = new HashMap<String, Object>(2);
+		
+		parameters.put(PARAM_NAME_IDS, parentIds);
+		parameters.put(PARAM_NAME_MIN_DISTANCE, minDistance);
+		parameters.put(BIND_LIMIT, limit);
+		
+		return namedParameterJdbcTemplate.query(
```
john-hill: if you use the "streamingJdbcTemplate" stream over the results and batch delete as you go.  The "WITH RECURSIVE" would only execute once.

lib/models/src/main/java/org/sagebionetworks/repo/model/NodeDAO.java
```diff
@@ -604,5 +605,18 @@
...
+	 * the respective input nodes. 
+	 * 
+	 * The keys in the sorted map is the distance value, the value for each key is the set of nodes that match the distance.
+	 *  
+	 * @param parentIds The input root nodes
+	 * @param minDistance The min distance between the input node a container in its subtree to filter by
+	 * @param limit The max number of container nodes to be fetched
+	 * @return A map sorted by the distance (desc) of the container nodes to their parents
+	 */
+	SortedMap<Integer, Set<Long>> getAllContainerIdsOrderByDistanceDesc(Collection<Long> parentIds, int minDistance, int limit);
```
john-hill: can remove

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/NodeDAOImpl.java
```diff
@@ -620,28 +626,54 @@ public Node getNodeForVersion(String id, Long versionNumber){
...
+		if (ids.isEmpty()) {
+			// no need to update database if not deleting anything
 			return 0;
 		}
-		
-		for(long id : ids){
+
+		for (long id : ids) {
 			String stringID = KeyFactory.keyToString(id);
 			transactionalMessenger.sendDeleteMessageAfterCommit(stringID, ObjectType.ENTITY);
```
marcomarasca: We can additionally send out a message for each container in the sub-trees.

marcomarasca:
Note: reduce the timer of the trash worker to 1 hour.

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/NodeDAOImpl.java
marcomarasca: Tried with the streaming template and unfortunately one cannot submit other queries on the same connection while streaming :( See https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-implementation-notes.html: "There are some caveats with this approach. You must read all of the rows in the result set (or close it) before you can issue any other queries on the connection, or an exception will be thrown.".

marcomarasca:
Closing for now as we decided to go in another direction

PR #3878, opened 2019-12-23
PLFM-5955

brucehoff:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/verification/VerificationMigrationTypeListener.java
```diff
@@ -0,0 +1,60 @@
...
+			}
+			
+			// if there is just one captured email then it must have been the notification email
+			if(dto.getEmails().size()==1) {
+				dto.setNotificationEmail(dto.getEmails().get(0));
+			} else {
+				// most of the remaining can be disambiguated by the current notification email:
+				//
+				// Note: org.sagebionetworks.repo.model.migration.MigrationType shows that 
+				// VERIFICATION_SUBMISSION migrates *after* NOTIFICATION_EMAIL
```
marcomarasca: Double checked with John and the verified in the migration-utility code base. Migration jobs for each primary type are run in parallel, so we cannot rely on the order of the MigrationType. We could write a dedicated asynchronous job to run after migration or implement a worker that listen to change messages sent out once we get out of read only mode?

marcomarasca (approved):

PR #3901, opened 2020-01-21
PLFM-6057: Sanitize empty base key when creating a storage location

marcomarasca:


marcomarasca:
This was branched from release-294. It's currently failing due to instability of tests (See https://sagebionetworks.jira.com/browse/PLFM-6056).

marcomarasca:
We decided to unify the validation on the baseKey given in input to avoid having different behavior between S3 and GC, see ProjectSettingManagerImpl validate methods and getBaseKey() usage.

PR #3909, opened 2020-01-24
PLFM-6051: Fixed Numeric Type issue depends on MySql Version

OCEO-YJ:


lib/lib-table-cluster/src/test/java/org/sagebionetworks/table/cluster/TableIndexDAOImplTest.java
```diff
@@ -1034,7 +1030,7 @@ public void testColumnInfoAndCardinality(){
 		assertTrue(info.hasIndex());
 		assertEquals("_C13_idx_", info.getIndexName());
 		assertEquals(MySqlColumnType.TINYINT, info.getType());
-		assertEquals(new Integer(1), info.getMaxSize());
```
zimingd, line 1037: add test case where we have a VARCHAR MySqlType and assert that maxSize's value equals an expected value

lib/lib-table-cluster/src/test/java/org/sagebionetworks/table/cluster/MySqlColumnTypeTest.java
```diff
@@ -28,9 +26,9 @@ public void testParserTypeNoSize(){
 	
 	@Test
 	public void testParseSizeBigInt(){
-		String typeString = "bigint(20)";
+		String typeString = "BIGINT";
 		Integer size = MySqlColumnType.parseSize(typeString);
-		assertEquals(new Integer(20), size);
+		assertNull(size);
```
zimingd, line 31: add test case for VARCHAR of the parseSize() method. 

String typeString = "VARCHAR(56)";

zimingd (approved):

PR #3941, opened 2020-02-12
merge 297 into develop

DwayneJengSage:
This is a no-op pull request that merges release-297 into develop for git consistency.


DwayneJengSage:
Verbal confirmation from Marco that admin-override merge of this change is fine.

PR #3966, opened 2020-03-05
PLFM-6012 STS Sample Code

DwayneJengSage:
https://sagebionetworks.jira.com/browse/PLFM-6012

I put the sample code in our Synapse-Repository-Services, to make it easier to share with others.

I also made an integration test for our sample code to ensure that it works and that it will continue to work as we make changes in the future.

### DIFF
--- a/client/sample-code/src/main/java/org/sagebionetworks/sample/sts/MigrateS3Bucket.java
+++ b/client/sample-code/src/main/java/org/sagebionetworks/sample/sts/MigrateS3Bucket.java
@@ -17,8 +17,6 @@ import java.io.IOException;
 import java.util.List;
 
 public class MigrateS3Bucket {
-	private static final String DELIMITER = "/";
-
 	private final DigestUtils md5DigestUtils;
 	private final AmazonS3 s3Client;
 	private final String s3Bucket;
@@ -45,7 +43,8 @@ public class MigrateS3Bucket {
 		// https://docs.synapse.org/articles/sts_storage_locations.html
 		this.synapseFolderId = synapseFolderId;
 
-		// This is the storage location ID that is set on the Synapse folder.
+		// This is the storage location ID that is set on the Synapse folder. Note that you must be the owner of the
+		// storage location in order to import existing S3 files into Synapse this way.
 		this.storageLocationId = storageLocationId;
 	}
 
@@ -67,7 +66,7 @@ public class MigrateS3Bucket {
 	private void executeForSubfolder(String s3SubFolder, String synapseSubFolderId) throws IOException,
 			SynapseException {
 		ListObjectsRequest listObjectsRequest = new ListObjectsRequest().withBucketName(s3Bucket)
-				.withPrefix(s3SubFolder).withDelimiter(DELIMITER);
+				.withPrefix(s3SubFolder).withDelimiter("/");
 		ObjectListing objectListing = s3Client.listObjects(listObjectsRequest);
 		boolean hasNext;
 		do {
@@ -75,14 +74,6 @@ public class MigrateS3Bucket {
 			List<S3ObjectSummary> objectSummaryList = objectListing.getObjectSummaries();
 			if (objectSummaryList != null) {
 				for (S3ObjectSummary objectSummary : objectSummaryList) {
-					String s3Key = objectSummary.getKey();
-					if (s3Key.equals(s3SubFolder)) {
-						// S3 does this weird thing where if you list objects inside of a folder, one of the results
-						// is the folder itself. Creating folders is handled in a different loop (common prefixes,
-						// below), so just skip this one.
-						continue;
-					}
-
 					migrateOneFile(objectSummary.getBucketName(), objectSummary.getKey(), synapseSubFolderId);
 				}
 			}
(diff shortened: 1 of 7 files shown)


client/sample-code/src/main/java/org/sagebionetworks/sample/sts/MigrateS3Bucket.java
```diff
@@ -0,0 +1,175 @@
...
+	 *
+	 * @param s3SubFolder
+	 *         the S3 subfolder from which we migrate files; if this is null then we export from the S3 bucket root
+	 * @param synapseSubFolderId
+	 *         the Synapse folder to which we migrate the S3 files
+	 */
+	private void executeForSubfolder(String s3SubFolder, String synapseSubFolderId) throws IOException,
+			SynapseException {
+		ListObjectsRequest listObjectsRequest = new ListObjectsRequest().withBucketName(s3Bucket)
+				.withPrefix(s3SubFolder).withDelimiter(DELIMITER);
```
DwayneJengSage: Try prefix includes the trailing delimiter, so we don't need that weird if statement.

client/sample-code/src/main/java/org/sagebionetworks/sample/sts/MigrateS3Bucket.java
```diff
@@ -0,0 +1,175 @@
...
+		this.s3Bucket = s3Bucket;
+		this.s3Folder = s3Folder;
+
+		// This is the entity ID of the folder in Synapse to which we migrate the S3 files. This folder must have an
+		// external S3 storage location with STS-enabled, as described in
+		// https://docs.synapse.org/articles/sts_storage_locations.html
+		this.synapseFolderId = synapseFolderId;
+
+		// This is the storage location ID that is set on the Synapse folder.
+		this.storageLocationId = storageLocationId;
```
DwayneJengSage: Only the storage location owner can import files from S3.

client/sample-code/src/main/java/org/sagebionetworks/sample/sts/MigrateSynapseProject.java
```diff
@@ -0,0 +1,131 @@
...
+	 * @param sourceSubFolderId
+	 *         the subfolder from which we migrate files
+	 * @param destinationSubFolderId
+	 *         the folder to which we migrate the files
+	 */
+	private void executeForSubfolder(String sourceSubFolderId, String destinationSubFolderId) throws IOException,
+			SynapseException {
+		EntityChildrenRequest entityChildrenRequest = new EntityChildrenRequest();
+		entityChildrenRequest.setParentId(sourceSubFolderId);
+		entityChildrenRequest.setIncludeTypes(CHILD_ENTITY_TYPE_LIST);
```
DwayneJengSage: Inline the constant for readability for sample code.

integration-test/pom.xml
```diff
@@ -269,6 +269,12 @@
 			<scope>test</scope>
 		</dependency>
 
+		<dependency>
+			<groupId>org.sagebionetworks</groupId>
+			<artifactId>sample-code</artifactId>
+			<version>develop-SNAPSHOT</version>
```
DwayneJengSage: The version should be in the root pom.xml under dependencyManagement.

integration-test/src/test/java/org/sagebionetworks/ITStorageLocation.java
```diff
@@ -196,7 +221,150 @@ public void synapseS3StorageLocationWithSts() throws SynapseException {
...
+				externalS3Bucket, baseKey, folder.getId(), externalS3StorageLocationSetting.getStorageLocationId());
+		migration.execute();
+
+		// Verify results.
+		// Get files in baseKey (f1, f2, and owner.txt).
+		Map<String, EntityHeader> childrenByName = getChildren(folder.getId(), EntityType.file);
+		assertEquals(3, childrenByName.size());
+		assertTrue(childrenByName.containsKey("f1"));
+		assertTrue(childrenByName.containsKey("f2"));
+		assertTrue(childrenByName.containsKey("owner.txt"));
```
DwayneJengSage: Make a comment in the sample code that we migrate owner.txt. You can add a check to prevent this.

client/sample-code/src/main/java/org/sagebionetworks/sample/sts/MigrateS3Bucket.java
DwayneJengSage: I don't know what I did wrong when I first tested this, but apparently I don't need to filter out the subfolder at all.

marcomarasca: The documentation I found is inconclusive on this. On one side the official docs (https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/model/ListObjectsRequest.html) seems to suggest that the delimiter is needed in the prefix to get our expected results but others suggest the behavior you reported initially. If I'm not wrong when you create a "folder" through the AWS console, AWS automatically adds a key named as the folder. The build is green so I'm guessing it's not an issue.

marcomarasca (approved):
Thanks!

PR #3968, opened 2020-03-06
switch to stored procedures for transaction managment, consitency and…

john-hill:
… performance


lib/id-generator/pom.xml
```diff
@@ -98,7 +98,7 @@
 	</dependencies>
 
 	<properties>
```
marcomarasca, line 100: we can remove the properties block

lib/id-generator/src/main/java/org/sagebionetworks/ids/IdGeneratorImpl.java
```diff
@@ -139,40 +119,34 @@ public void afterPropertiesSet() throws Exception {
...
-		if(count > 1){
-			long now = System.currentTimeMillis();
-			idGeneratorJdbcTemplate.update(String.format(INSERT_SQL_INCREMENT, type.name()), lastId, now);
+	
+	public void createStoredProcedure(String name) {
+		String ddl = loadDDLString(name);
+		try {
+			idGeneratorJdbcTemplate.execute(ddl);
+		} catch (DataAccessException e) {
+			if(!e.getMessage().contains("already exists")) {
```
marcomarasca, line 114: We could check for a specific error code

lib/id-generator/pom.xml
marcomarasca: Double check for unused dependencies (since we removed the transactional context)

lib/id-generator/src/main/resources/generateNewId.ddl.sql
```diff
@@ -0,0 +1,23 @@
+CREATE PROCEDURE generateNewId(IN typeName VARCHAR(256))
+    MODIFIES SQL DATA
+    SQL SECURITY INVOKER
+BEGIN
+	DECLARE newId BIGINT;
+    DECLARE typeLock VARCHAR(256);
+	
+	SET SESSION TRANSACTION ISOLATION LEVEL READ COMMITTED;
```
marcomarasca, line 8: We might not need a transaction nor the lock table

lib/id-generator/src/main/resources/generateNewId.ddl.sql
```diff
@@ -0,0 +1,25 @@
+CREATE PROCEDURE generateNewId(IN typeName VARCHAR(256))
+    MODIFIES SQL DATA
+    SQL SECURITY INVOKER
+BEGIN
+	DECLARE newId BIGINT;
+    DECLARE typeLock VARCHAR(256);
```
marcomarasca, line 6: Can be removed

lib/id-generator/src/main/resources/reserveId.ddl.sql
```diff
@@ -0,0 +1,16 @@
+CREATE PROCEDURE reserveId(IN idToReserve BIGINT, IN typeName VARCHAR(256))
+    MODIFIES SQL DATA
+    SQL SECURITY INVOKER
+BEGIN
+    DECLARE typeLock VARCHAR(256);
```
marcomarasca, line 5: Can be removed

marcomarasca (approved):

PR #3970, opened 2020-03-12
PLFM-5908: Allows to disable certification requirement for project

marcomarasca:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/ProjectSettingsManagerImpl.java
```diff
@@ -111,19 +139,32 @@ public ProjectSetting createProjectSetting(UserInfo userInfo, ProjectSetting pro
 
 		// make sure the project id is a project
 		EntityType nodeType = nodeManager.getNodeType(userInfo, parentId);
```
john-hill, line 141: If of the certified type then add the extra check but all other logic still applies.

john-hill (approved):

PR #3980, opened 2020-03-26
PLFM-6046

xschildw:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/wikiV2/V2DBOWikiPageDaoImpl.java
```diff
@@ -359,8 +359,6 @@ public V2WikiPage updateWikiPage(V2WikiPage wikiPage,
 			basicDao.createBatch(attachmentsToInsert);
 		}
 		
-		// Send the change message
-		transactionalMessenger.sendMessageAfterCommit(newDbo, ChangeType.UPDATE);
```
marcomarasca, line 363: We should not remove this, the WIKI object type has a SNS topic in the stack builder.

lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/dao/DBOActivityDAOImplTest.java
```diff
@@ -92,7 +92,6 @@ public void testLockActivityAndIncrementEtag() throws Exception {
 		
 		ChangeType type = ChangeType.UPDATE;
 		String returnedNewEtag = activityDao.lockActivityAndGenerateEtag(id.toString(), oldEtag, type);
-		verify(mockMessenger).sendMessageAfterCommit(eq(mockDbo), eq(type));
```
marcomarasca, line 95: We should not remove this, the WIKI object type has a SNS topic in the stack builder.

marcomarasca (changes requested):
We should not remove the WIKI messages, but rather the ObjectType.FORUM ones (there is no SNS topic for FORUM)

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/DBOActivityDAOImpl.java
```diff
@@ -136,7 +135,6 @@ public String lockActivityAndGenerateEtag(String id, String eTag, ChangeType cha
 		// Get a new etag
 		DBOActivity dbo = getDBO(id);
 		dbo.seteTag(UUID.randomUUID().toString());
-		transactionalMessenger.sendMessageAfterCommit(dbo, changeType);
 		return dbo.getEtag();
 	}
```
marcomarasca, line 136: We should also remove the message that is sent when deleting the activity

lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/dao/DBOActivityDAOImplTest.java
xschildw: This one still needed (activity)

marcomarasca: Right, wrong class :)

marcomarasca (approved):

PR #3984, opened 2020-03-30
change STS GetFederationToken to AssumeRole

DwayneJengSage:
https://sagebionetworks.jira.com/browse/PLFM-6008

GetFederationToken cannot be called by an IAM Role, and Elastic Beanstalk uses IAM roles by default. We switched from GetFederationToken to AssumeRole.

Requires changes to Stack Builder (https://github.com/Sage-Bionetworks/Synapse-Stack-Builder/pull/328). Requires you to add an entry to ~/.m2/settings:

```
<org.sagebionetworks.sts.iam.arn>(insert role arn here)</org.sagebionetworks.sts.iam.arn>
```


DwayneJengSage (commented):
In STS Autowire test and integ test, only run the tests if the temp credentials ARN exists.

DwayneJengSage:
Locally tested this against a bucket in the Bridge account, and it works.

marcomarasca (approved):

PR #4018, opened 2020-04-24
Fix UNNEST() on a list column only returning 5 results (PLFM-6206)

zimingd:


lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/model/FunctionReturnType.java
```diff
@@ -13,12 +17,18 @@
 	LONG(ColumnType.INTEGER),
 	DOUBLE(ColumnType.DOUBLE),
 	// Cases where the return type matches the parameter type.
-	MATCHES_PARAMETER(null);
-	
-	private ColumnType columnType;
-	
+	MATCHES_PARAMETER(new MatchParameterFunctionReturnTypeHandler()),
```
zimingd, line 20: missing info about string max length if it is a string type

john-hill (approved):

PR #4023, opened 2020-04-29
temporary code: re-serialization of columnmodel w/out validation for migration (PLFM-6213)

zimingd:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/persistence/table/DBOColumnModel.java
```diff
@@ -127,13 +140,36 @@ public MigrationType getMigratableTableType() {
...
-				DBOColumnModel modifiedDBO = ColumnModelUtils.createDBOFromDTO(columnModel, Integer.MAX_VALUE);
+				DBOColumnModel modifiedDBO = createDBOFromDTO(columnModel);
 				return modifiedDBO;
 			}
 		};
 	}
 
+	@TemporaryCode(author="ziming", comment = "one-time migration change. remove after stack 309")
+	public static DBOColumnModel createDBOFromDTO(ColumnModel dto) {
+		try {
```
marcomarasca: set the id of the dto to null before computing the hash and saving it back

marcomarasca (approved):

PR #4029, opened 2020-05-01
change list max length sql and push validation down to DAO level (PLFM-6168)

zimingd:


lib/lib-table-cluster/src/test/java/org/sagebionetworks/table/cluster/TableIndexDAOImplUnitTest.java
```diff
@@ -0,0 +1,218 @@
...
+
+
+		ObjectType objectType = ObjectType.FILE;
+		long viewId = 123L;
+		long viewTypeMask = 1L;
+		Set<Long> allContainersInScope = Sets.newHashSet(111L,222L);
+		Set<Long> objectIdFilter = null;
+
+		// method under test
+		spyDao.validateMaxListLengthInAnnotationReplication(objectType,viewId,viewTypeMask,
```
marcomarasca, line 73: You might want to check that there are no DB interactions instead and wrap it into the assertDoesNotThrow to give it context

lib/lib-table-cluster/src/test/java/org/sagebionetworks/table/cluster/TableIndexDAOImplTest.java
```diff
@@ -1417,7 +1417,7 @@ public void testGetMaxListSizeForAnnotations_nullScope() throws ParseException {
 		Set<String> annotationNames = Sets.newHashSet("foo", "bar");
 
 		String errorMessage = assertThrows(IllegalArgumentException.class, () ->
-				tableIndexDAO.getMaxListSizeForAnnotations(objectType, tableId.getId(),
+				((TableIndexDAOImpl) tableIndexDAO).getMaxListSizeForAnnotations(objectType, tableId.getId(),
```
marcomarasca, line 1420: This kind of methods might go into the new unit test class

services/workers/src/test/java/org/sagebionetworks/table/worker/TableViewIntegrationTest.java
```diff
@@ -1575,6 +1574,45 @@ public void testEntityView_multipleValueColumnRoundTrip_changeMaxListLength() th
...
+		RowSet rowset = result.getQueryResult().getQueryResults();
+		rowset.getRows().get(0).setValues(Collections.singletonList("[\"val1\",\"val2\",\"val3\",\"val4\"]"));
+
+		// wait for the change to complete
+//		startAndWaitForJob(adminUserInfo, transaction, TableUpdateTransactionResponse.class);
+
+		String error = assertThrows(AsynchJobFailedException.class, () ->
+				updateView(rowset,fileViewId)
+		).getMessage();
+		assertEquals("Value at [0,16] was not a valid STRING_LIST. Exceeds the maximum number of list elements defined in the ColumnModel (3): \"[\"val1\",\"val2\",\"val3\",\"val4\"]\"", error);
```
marcomarasca, line 1614: Can you double check if we truncate the number of values at the end of the message? The error message might become extremely long, maybe we can remove the list of values in the error message (e.g. we might just need the total count)?

marcomarasca (approved):

PR #4035, opened 2020-05-07
Plfm 6158

marcomarasca:


lib/lib-table-cluster/src/test/java/org/sagebionetworks/table/cluster/SQLUtilsTest.java
```diff
@@ -3020,7 +3023,66 @@ public void testGetDeleteRowsFromViewSql() {
...
+		boolean filterByObjectId = true;
+		
+		ViewScopeFilter scopeFilter = getSQLScopeFilter(subTypes, filterByObjectId);
+		
+		String result = SQLUtils.getViewScopeFilterColumn(scopeFilter);
+		
+		assertEquals(TableConstants.OBJECT_REPLICATION_COL_OBJECT_ID, result);
+	}
+	
+	private ViewScopeFilter getSQLScopeFilter(List<Enum<?>> subTypes, boolean filterByObjectId) {
```
john-hill, line 3085: rename

lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/metadata/ObjectFieldTypeMapper.java
```diff
@@ -10,6 +11,11 @@
  * @author Marco Marasca
  */
 public interface ObjectFieldTypeMapper {
+	
+	/**
+	 * @return The object type this mapper applies to
+	 */
+	ObjectType getObjectType();
```
john-hill, line 18: remove it not needed

lib/models/src/main/java/org/sagebionetworks/repo/model/table/ViewScopeFilter.java
```diff
@@ -0,0 +1,65 @@
...
+		this.subTypes = subTypes;
+		this.filterByObjectId = filterByObjectId;
+		this.containerIds = containerIds;
+	}
+	
+	public ObjectType getObjectType() {
+		return objectType;
+	}
+
+	public List<Enum<?>> getSubTypes() {
```
john-hill, line 32: Consider a List<String> here since at this level it does not matter that they were enums.

john-hill (approved):

PR #4041, opened 2020-05-12
additional LIKE filter API (PLFM-6192)

zimingd:


lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/table/LikeQueryFilter.json
```diff
@@ -0,0 +1,17 @@
+{
+	"implements": [{"$ref":"org.sagebionetworks.repo.model.table.QueryFilter"}],
+	"description": "A LIKE filter against a column. Values within the values list have an OR relationship",
+	"properties": {
+		"columnName":{
+			"type": "string",
+			"description": "name of the column to filter"
+		},
```
marcomarasca: Consider naming this as ColumnValueFilter and use an Enum with the supported Operators so we could extend it in the future.

lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/SQLUtils.java
```diff
@@ -2046,4 +2047,63 @@ public static String getViewScopeSubTypeFilter(ViewScopeFilter scopeFilter) {
 		
 		return builder.toString();
 	}
+
+	public static String appendAdditionalFilters(List<QueryFilter> additionalFilters){
```
marcomarasca: Consider move this to the SQLTranslatorUtils and rename it to translateQueryFilters

lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/SqlQuery.java
```diff
@@ -178,6 +182,20 @@ public SqlQuery(
 			SelectList expandedSelectList = SQLTranslatorUtils.createSelectListFromSchema(tableSchema);
 			this.model.replaceSelectList(expandedSelectList);
 		}
+
+
+		//Append additionalFilters onto the WHERE clause
+		if(additionalFilters != null && !additionalFilters.isEmpty()) {
+			String additionalFilterSearchCondition = SQLUtils.appendAdditionalFilters(additionalFilters);
+			StringBuilder whereClauseBuilder = new StringBuilder();
+			FacetUtils.appendFacetWhereClauseToStringBuilderIfNecessary(whereClauseBuilder,additionalFilterSearchCondition, this.model.getTableExpression().getWhereClause());
```
marcomarasca: We could probably make this method generic (maybe move it to some query/sql util class) and reuse it for facets

marcomarasca (approved):

marcomarasca (approved):

PR #4045, opened 2020-05-14
Plfm 6161 f

john-hill:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/schema/JsonSchemaManagerImpl.java
```diff
@@ -366,5 +368,59 @@ public ListJsonSchemaVersionInfoResponse listSchemaVersions(ListJsonSchemaVersio
...
 	}
+	
+	/**
+
+	 * 
+	 * @param id
+	 * @return
+	 */
+	@Override
+	public JsonSchema getValidationSchema(String id) {
```
marcomarasca, line 379: Let us add a unit test

services/workers/src/test/resources/pets/Cat.json
```diff
@@ -0,0 +1,15 @@
...
+	"description": "...",
+	"allOf": [
+		{
+			"$ref": "my.organization/pets.Pet"
+		}
+	],
+	"properties": {
+		"breed": {
+			"$ref": "my.organization/pets.cat.Breed"
+		}
```
marcomarasca, line 13: We might want to add an additional integration test where we have a similar hierarchy but additionally we have a "loop" of $ref, for example here we could add a property named dogFriends that are of type pets.Dog.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/schema/SchemaTranslatorImp.java
```diff
@@ -33,7 +34,7 @@
 	@Override
 	public ObjectSchemaImpl loadSchemaFromClasspath(String id) {
```
marcomarasca, line 35: Consider renaming the class from SchemaTranslatorImp -> SchemaTranslatorImpl

marcomarasca (approved):

PR #4046, opened 2020-05-14
 PLFM-6211: allow Docker client to use an access token instead of a password

brucehoff:


services/repository/src/main/java/org/sagebionetworks/auth/filter/DockerClientAuthFilter.java
```diff
@@ -56,43 +62,58 @@ protected boolean reportBadCredentialsMetric() {
 
 	@Override
 	protected boolean validCredentials(UserNameAndPassword credentials) {
```
marcomarasca: We might want to modify the abstract method to return some sort of object that could be used as a signal in the subsequent step that there is a different type of "credentials".

services/repository/src/main/java/org/sagebionetworks/auth/filter/DockerClientAuthFilter.java
```diff
@@ -56,43 +62,58 @@ protected boolean reportBadCredentialsMetric() {
...
+					PrincipalAlias alias = authenticationService.lookupUserForAuthentication(username);
+					userId = alias.getPrincipalId();
+					accessToken = oidcTokenHelper.createTotalAccessToken(userId);
+				} catch (NotFoundException e) {
+					rejectRequest(response, getInvalidCredentialsMessage());
+					return;
+				}
+			}
+			HttpAuthUtil.setBearerTokenHeader(modHeaders, accessToken);
+			if (!authenticationService.hasUserAcceptedTermsOfUse(userId)) {
```
marcomarasca: Consider introducing a dedicated filter for the term of use check that comes after all the authentication filters and could be applied to any endpoint regardless.

services/repository/src/main/java/org/sagebionetworks/auth/filter/AcceptTermsOfUseFilter.java
```diff
@@ -0,0 +1,82 @@
...
+	
+	private AuthenticationService authenticationService;
+
+	@Autowired
+	public AcceptTermsOfUseFilter(AuthenticationService authenticationService) {
+		this.authenticationService = authenticationService;
+	}
+	
+	@Override
+	public final void doFilter(ServletRequest request, ServletResponse response, FilterChain filterChain)
```
marcomarasca: A little trickery in the (CGLIB) proxies that we use for the spring beans, I'm afraid that the method cannot be final

marcomarasca (approved):

PR #4056, opened 2020-05-26
PLFM-6159 - Part B: Submission metadata index implementation

marcomarasca:
This follows from https://github.com/Sage-Bionetworks/Synapse-Repository-Services/pull/4052, the following stack builder PR is needed before merging: https://github.com/Sage-Bionetworks/Synapse-Stack-Builder/pull/334

PR #4062, opened 2020-05-27
PLFM-6159 - Part C: New submission view entity

marcomarasca:
This builds on top of https://github.com/Sage-Bionetworks/Synapse-Repository-Services/pull/4056 which should be merged back before reviewing.


lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/table/ViewScope.json
```diff
@@ -7,15 +7,15 @@
...
 			"items": {
 				"type": "string"
 			}
 		},
-		"objectType": {
-			"description": "The type of the objects in the rows of a view (E.g. ENTITY for entity views)",
-			"$ref": "org.sagebionetworks.repo.model.table.ViewObjectType"
+		"viewEntityType": {
+			"description": "The entity type of the view, if not supplied defaults to entityview. Supported values are entityview and submissionview",
+			"$ref": "org.sagebionetworks.repo.model.EntityType"
```
john-hill, line 18: new ViewEntityType (too many unused options in EntityType)

services/repository/src/main/java/org/sagebionetworks/repo/web/service/metadata/SubmissionViewMetadataProvider.java
```diff
@@ -0,0 +1,57 @@
...
+			throws InvalidModelException, NotFoundException, DatastoreException, UnauthorizedException {
+		validateScopeAccess(event.getUserInfo(), view.getScopeIds());		
+	}
+	
+	private void validateScopeAccess(UserInfo userInfo, List<String> scope) {
+		if (scope == null || scope.isEmpty()) {
+			return;
+		}
+		scope.forEach((evaluationId) -> {
+			evaluationPermissionManager.hasAccess(userInfo, evaluationId, ACCESS_TYPE.READ_PRIVATE_SUBMISSION);
```
john-hill, line 53: is there a way to pass the scope and get back the sub-set the user has the  passed permission.  That would be a single call instead of n calls.

john-hill (approved):

PR #4076, opened 2020-06-03
Plfm 6209

brucehoff:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/RestrictionInformationManagerImpl.java
```diff
@@ -0,0 +1,82 @@
...
+		ValidateArgument.required(request.getObjectId(), "RestrictionInformationRequest.objectId");
+		ValidateArgument.required(request.getRestrictableObjectType(), "RestrictionInformationRequest.restrictableObjectType");
+		RestrictionInformationResponse info = new RestrictionInformationResponse();
+
+		boolean userIsFileCreator=false;
+		if (RestrictableObjectType.ENTITY == request.getRestrictableObjectType()) {
+			// if the user is the owner of the entity (and the entity is a File), then she automatically 
+			// has access to the object and therefore has no unmet access requirements
+			Long principalId = userInfo.getId();
+			Node node = nodeDao.getNode(request.getObjectId());
```
john-hill, line 45: can use nodeDao.getCreatedBy()

services/repository-managers/src/main/resources/private/managers-spb.xml
```diff
@@ -281,6 +281,8 @@
 
 	<bean id="accessRequirementManager" class="org.sagebionetworks.repo.manager.AccessRequirementManagerImpl" />
 
+	<bean id="restrictionInformationManager" class="org.sagebionetworks.repo.manager.RestrictionInformationManagerImpl" />
```
john-hill, line 284: consider an annotation @Service and no xml bean.

john-hill (approved):

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/RestrictionInformationManagerImpl.java
brucehoff: Actually we can't!  We have to get both the creator and the entity type otherwise we will report that the owner of a Folder has no access restrictions, while they may be unable to access restricted  files in the folder (created by others).

PR #4080, opened 2020-06-04
PLFM-5753 - Code review changes and additional DAO methods for token limits

nickgros:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/auth/OAuthRefreshTokenDaoImpl.java
```diff
@@ -85,6 +86,21 @@
...
+			+ " AND " + COL_OAUTH_REFRESH_TOKEN_CLIENT_ID + " = :" + PARAM_CLIENT_ID
+			+ " AND " + COL_OAUTH_REFRESH_TOKEN_LAST_USED + " > (NOW() - INTERVAL  :" + PARAM_DAYS + " DAY)";
+
+	private static final String DELETE_LEAST_RECENTLY_USED_ACTIVE_TOKEN = 
+			"DELETE FROM " + TABLE_OAUTH_REFRESH_TOKEN
+				+ " WHERE " + COL_OAUTH_REFRESH_TOKEN_PRINCIPAL_ID + " = :" + PARAM_PRINCIPAL_ID
+				+ " AND " + COL_OAUTH_REFRESH_TOKEN_CLIENT_ID + " = :" + PARAM_CLIENT_ID
+				+ " AND " + COL_OAUTH_REFRESH_TOKEN_LAST_USED
+				+ " > (NOW() - INTERVAL  :" + PARAM_DAYS + " DAY)"
+				+ " ORDER BY " + COL_OAUTH_REFRESH_TOKEN_LAST_USED + " ASC"
```
brucehoff: instead of
" AND lastUsed > threshold ORDER BY  lastUsed ASC LIMIT 1"
what do you think of
"ORDER BY lastUsed DESC OFFSET 100"
(where 100 is just an example of the number of tokens you want to retain).
I.e. this deletes all but the most recent 100 tokens.  It has three advantages:
(1) You don't need to know the count before executing the query;
(2) There is no 'race condition' in which you check the count and then it changes (a token expires) just before you run the delete query;
(3) It cleans up old, expired tokens.
Essentially the operation is idempotent.  It just cleans up all but the most recent 100 tokens.

brucehoff (changes requested):

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/auth/OAuthRefreshTokenDaoImpl.java
nickgros: I think this is a great idea. I will make that change.

nickgros: I think that the COUNT query introduced in this PR is also obsoleted by this change (there is no need to get a count if we have a "safe" DELETE query), so I will remove that as well.

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/auth/OAuthRefreshTokenDaoImpl.java
```diff
@@ -85,6 +87,25 @@
...
 			+ " AND " + COL_OAUTH_REFRESH_TOKEN_CLIENT_ID + " = :" + PARAM_CLIENT_ID;
 
+	/*
+	 * We use a JOIN because
+	 *   - MySQL doesn't support OFFSET when using DELETE
+	 *   - MySQL doesn't support LIMIT & IN/ALL/ANY/SOME subquery (so we can't do DELETE WHERE id IN (SELECT ... LIMIT x OFFSET y))
+	 *
+	 * https://stackoverflow.com/a/42030157
+	 */
+	private static final String DELETE_LEAST_RECENTLY_USED_ACTIVE_TOKEN = "DELETE t FROM " + TABLE_OAUTH_REFRESH_TOKEN + " t "
```
brucehoff: DELETE_LEAST_RECENTLY_USED_ACTIVE_TOKENS (i.e. plural)

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/auth/OAuthRefreshTokenDaoImpl.java
```diff
@@ -85,6 +87,25 @@
...
+			+ " JOIN ("
+				+ "SELECT tt." + COL_OAUTH_REFRESH_TOKEN_ID
+				+ " FROM " + TABLE_OAUTH_REFRESH_TOKEN + " tt "
+				+ " WHERE " + COL_OAUTH_REFRESH_TOKEN_PRINCIPAL_ID + " = :" + PARAM_PRINCIPAL_ID
+				+ " AND " + COL_OAUTH_REFRESH_TOKEN_CLIENT_ID + " = :" + PARAM_CLIENT_ID
+				+ " AND " + COL_OAUTH_REFRESH_TOKEN_LAST_USED
```
brucehoff: I don't think you need AND COL_OAUTH_REFRESH_TOKEN_LAST_USED > (NOW() - INTERVAL : PARAM_DAYS DAY).  If you omit it then this SQL guarantees that it never leaves more than PARAM_MAX_NUM_TOKENS for a user+client.  If you include it then while you clean up excessive, unexpired tokens, any expired ones will just linger forever.

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/auth/OAuthRefreshTokenDaoImpl.java
```diff
@@ -252,4 +269,18 @@ public void deleteAllTokensForUserClientPair(String userId, String clientId) {
 		params.addValue(PARAM_CLIENT_ID, clientId);
 		namedParameterJdbcTemplate.update(DELETE_TOKEN_BY_CLIENT_USER_PAIR, params);
 	}
+
+	@Override
+	public void deleteLeastRecentlyUsedTokensIfOverLimit(String userId, String clientId, Long maxLeaseLengthInDays, Long maxNumberOfTokens) {
```
brucehoff: As stated above, I think you can omit maxLeaseLengthInDays and just delete based on maxNumberOfTokens

brucehoff (changes requested):
Please see the in-line comments.  I think the clean-up SQL can be slightly simpler, filtering on just the count, not the expiration date.

lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/auth/OAuthRefreshTokenDaoAutowiredTest.java
```diff
@@ -392,4 +403,24 @@ void deleteTokensForUserClientPair() {
...
+		// Note: this date is still in the "active" window (in this case, 180 days)
+		Date leastRecentlyUsedDate = new Date(System.currentTimeMillis() - ONE_DAY_MILLIS * 30);
+
+		// Create a bunch of tokens, one of which will be the "expected" retrieval based on date
+		OAuthRefreshTokenInformation toRemove = createRefreshToken("abcd", leastRecentlyUsedDate);
+		OAuthRefreshTokenInformation doNotRemove1 = createRefreshToken("abcd", new Date());
+		OAuthRefreshTokenInformation doNotRemove2 = createRefreshToken("abcd", new Date(System.currentTimeMillis() - 1000 * 5));
+
+		// We will create an expired token -- it will be deleted, but this is an implementation detail, rather than a requirement
+		OAuthRefreshTokenInformation expired = createRefreshToken("abcd", new Date(System.currentTimeMillis() - ONE_YEAR_MILLIS));
```
brucehoff: rename to avoid confusion

brucehoff (approved):

PR #4092, opened 2020-06-14
PLFM-5958

xschildw:


xschildw:
After discussion with Marco, decided to move read-only user creation to stack builder.

PR #4100, opened 2020-06-17
PLFM-6234

leem42:
Removed isConsistent from Query Service
- Query service now exclusively runs with a read lock on the table and includes the etag
- removed isConsistent from Query class definition, usage across all files calling query service, in tests, docs
- Removed two tests entirely that tested specific behavior of the isConsistent parameter. One tested the default value for isConsistent, the other was the behavior of running a query with isConsistent = false. 

Made a few other small changes - removed unused variable, fixed a typo

The github link to my jenkins build is broken, it doesn't include the p in my jenkins account name, can view the passing build here http://build-system-synapse.sagebase.org:8081/job/devpmlee/9/console


brucehoff (approved):
Looks good!

PR #4111, opened 2020-06-25
PLFM-6306: Fix for race condition on replication

marcomarasca:


lib/lib-table-cluster/src/test/java/org/sagebionetworks/table/cluster/TableIndexDAOImplTest.java
```diff
@@ -1347,6 +1347,57 @@ public void testEntityReplication(){
...
+		
+		// Re-adding to the index should simply update the object without throwing
+		tableIndexDAO.addObjectData(objectType, Collections.singletonList(objectData));
+		
+		// Makes sure the data was updated
+		result = tableIndexDAO.getObjectData(ViewObjectType.ENTITY, objectData.getId());
+		
+		assertEquals(objectData, result);
+	}
+
```
john-hill, line 1381: test with batch of objects with different ids.

john-hill (approved):

PR #4117, opened 2020-07-01
PLFM-6241 - Log the OAuth client id from the basic auth creds

nickgros:


services/repository/src/main/java/org/sagebionetworks/repo/web/AccessInterceptor.java
```diff
@@ -51,9 +51,19 @@
...
+		 * There are two different places the client ID might be:
+		 *  - in the access token, if the OAuth client is acting on behalf of a user
+		 *  - in the basic auth credentials, if using an auth code/refresh token
+		 */
 		String accessToken = HttpAuthUtil.getBearerTokenFromStandardAuthorizationHeader(request);
-		if (accessToken==null) return null;
-		return oidcTokenHelper.parseJWT(accessToken).getBody().getAudience();
+		if (accessToken != null) {
+			return oidcTokenHelper.parseJWT(accessToken).getBody().getAudience();
+		} else if (HttpAuthUtil.usesBasicAuthentication(request)) {
```
marcomarasca: A small warning about this approach: this might not be enough since we use basic authentication in other filters that are not related to oauth and the "username" extracted might not actually be an OAuth client id (e.g. might be the user for docker registry or cloud mail in, a service call etc.). You might want to check for the special OAUTH_VERIFIED_CLIENT_ID_HEADER injected in the request instead (see the implementation of OAuthClientAuthFilter)

brucehoff: "+1". 

The two choices I see for changing this are (1) make sure it's the oauth client credentials rather than some other client credentials or (2) add a separate field, BASIC_AUTH_USER_NAME, which can be an oauth client or a non-oauth client.

nickgros: Thanks for the heads up, @marcomarasca ! It sounds like the right approach would be to use the injected header to conditionally insert the username into the OAuthClientId value, and as @brucehoff said, to log the basic auth username for all requests that include it.

services/repository/src/test/java/org/sagebionetworks/repo/web/AccessInterceptorTest.java
```diff
@@ -196,7 +199,54 @@ public void testStatusCode400() throws Exception{
...
+		Claims claims = new DefaultClaims();
+		claims.setAudience(OAUTH_CLIENT_ID);
+		Jwt<JwsHeader,Claims> jwt = new DefaultJwt(null, claims);
+		when(oidcTokenHelper.parseJWT(any())).thenReturn(jwt);
+
+		// Start
+		interceptor.preHandle(mockRequest, mockResponse, mockHandler);
+		interceptor.setReturnObjectId("returnId");
+		// Wait to add some elapse time
+		testClock.sleep(234);
```
brucehoff, line 219: why do we sleep?

services/repository/src/test/java/org/sagebionetworks/repo/web/AccessInterceptorTest.java
```diff
@@ -196,7 +199,54 @@ public void testStatusCode400() throws Exception{
...
+	public void testGetOAuthClientIdFromBearerJwt() throws Exception {
+		// Put the client ID in the JWT access token
+		when(mockRequest.getHeader("Authorization")).thenReturn(BEARER_TOKEN_HEADER);
+		Claims claims = new DefaultClaims();
+		claims.setAudience(OAUTH_CLIENT_ID);
+		Jwt<JwsHeader,Claims> jwt = new DefaultJwt(null, claims);
+		when(oidcTokenHelper.parseJWT(any())).thenReturn(jwt);
+
+		// Start
+		interceptor.preHandle(mockRequest, mockResponse, mockHandler);
```
brucehoff, line 216: // method under test

brucehoff (approved):

PR #4119, opened 2020-07-02
Plfm 6325

john-hill:


services/repository-managers/pom.xml
```diff
@@ -58,7 +59,17 @@
...
+		<!-- Force maven to use our modified version of json.org by including it 
+			before com.github.everit-org.json-schema which also depends on it. -->
+		<dependency>
+			<groupId>org.json</groupId>
+			<artifactId>JSON-Java</artifactId>
+		</dependency>
+		<!-- JSON schema validation library -->
+		<dependency>
+			<groupId>com.github.everit-org.json-schema</groupId>
+			<artifactId>org.everit.json.schema</artifactId>
```
marcomarasca, line 69: Please add the exclusion block on the conflicting org.json dependency so we are sure that it's not added to the classpath.

marcomarasca: This might potentially be added to the root pom.xml in the dependency management section

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/schema/SchemaTranslatorImp.java
```diff
@@ -177,15 +177,17 @@ JsonSchema translate(ObjectSchema objectSchema, boolean isRoot) {
 
 	void translateConcreteType(ObjectSchema objectSchema, JsonSchema jsonSchema) {
 		if(objectSchema.getType() == null || TYPE.OBJECT.equals(objectSchema.getType())) {
-			Map<String, JsonSchema> properties = jsonSchema.getProperties();
-			if(properties == null) {
-				properties = new LinkedHashMap<String, JsonSchema>(1);
+			if(objectSchema.getId() != null) {
```
marcomarasca, line 180: Could you renamed the class to SchemaTranslatorImpl.java

marcomarasca (approved):

PR #4121, opened 2020-07-02
PLFM-6286

nickgros:


services/repository/src/test/java/org/sagebionetworks/auth/filter/AuthenticationFilterTest.java
```diff
@@ -299,7 +302,7 @@ public void testFilter_validCredentials() throws Exception {
 		assertEquals(""+userId, requestCaptor.getValue().getParameter(AuthorizationConstants.USER_ID_PARAM));
 		assertEquals("Bearer "+BEARER_TOKEN, requestCaptor.getValue().getHeader(AuthorizationConstants.SYNAPSE_AUTHORIZATION_HEADER_NAME));
 	}
-	
+
```
brucehoff, line 305: revisit testFilter_invalid_AccessToken which might be a better place to test than in the integration test

brucehoff (approved):

PR #4155, opened 2020-08-10
Plfm 6368

brucehoff:
We have to be careful when requiring 'authorize' scope.  The scope is powerful, allowing operations like granting access to third party app's and creating personal access tokens.  Users shouldn't have to grant this scope in order to do common operations.  This PR updates the scope required by various services to remove 'authorize' where not needed:

GET /bearerToken  has no required scope.  (The returned token is limited by the scope(s) in the access token.)
POST /oauth2/session has no required scope.  (It's not an authenticated request in the normal sense and scope is not checked.)
POST /entity/id/acl no longer requires 'authorize' scope, just 'view' and 'modify'.
PUT /entity/id/acl no longer requires 'authorize' scope, just 'view' and 'modify'.
DELETE /entity/id/acl no longer requires 'authorize' scope, just 'modify'.
PUT /form/group/id/acl no longer requires 'authorize' scope, just 'view' and 'modify'.
PUT /schema/organization/id/acl no longer requires 'authorize' scope, just 'view' and 'modify'.
PUT /team/acl no longer requires 'authorize' scope, just 'view' and 'modify'.
PUT /evaluation/acl no longer requires 'authorize' scope, just 'view' and 'modify'.


services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/DockerManagerImplUnitTest.java
```diff
@@ -45,6 +50,13 @@
...
 
+import com.google.common.collect.ImmutableList;
+
+import io.jsonwebtoken.Claims;
+import io.jsonwebtoken.JwsHeader;
+import io.jsonwebtoken.Jwt;
+import io.jsonwebtoken.Jwts;
+
 
 @RunWith(MockitoJUnitRunner.class)
```
marcomarasca, line 61: Might be worth updating to Junit5

marcomarasca (approved):

PR #4162, opened 2020-08-13
PLFM-6195: Unhandled exception return 500, log to cloudwatch

marcomarasca:


john-hill (approved):

john-hill:
Consider if 500 is what we want for all exceptions or wait to see what types of errors we see with the new logging.

PR #4229, opened 2020-11-05
PLFM-6494

nickgros:
https://sagebionetworks.jira.com/browse/PLFM-6494

PR #4231, opened 2020-11-16
PLFM-6494 - Handle correct exception type and add autowired test

nickgros:
https://sagebionetworks.jira.com/browse/PLFM-6494

The prior fix (#4229 ) did not handle the correct exception (the exception was caught and wrapped in a different exception class at a lower level). I added a test to the autowired suite to ensure that the correct exception is thrown and handled.

### DIFF
--- a/services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/authentication/PersonalAccessTokenManagerImpl.java
+++ b/services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/authentication/PersonalAccessTokenManagerImpl.java
@@ -125,8 +125,12 @@ public class PersonalAccessTokenManagerImpl implements PersonalAccessTokenManage
 
 		try {
 			record = personalAccessTokenDao.createTokenRecord(record);
-		} catch (DuplicateKeyException e) {
-			throw new IllegalArgumentException(DUPLICATE_TOKEN_NAME_MSG, e);
+		} catch (IllegalArgumentException e) {
+			if (e.getMessage().contains("org.springframework.dao.DuplicateKeyException")) {
+				throw new IllegalArgumentException(DUPLICATE_TOKEN_NAME_MSG, e);
+			} else {
+				throw e;
+			}
 		}
 		AccessTokenGenerationResponse response = new AccessTokenGenerationResponse();
 		response.setToken(oidcTokenHelper.createPersonalAccessToken(oauthEndpoint, record));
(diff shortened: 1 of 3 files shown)


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/authentication/PersonalAccessTokenManagerImpl.java
```diff
@@ -125,8 +125,12 @@ public AccessTokenGenerationResponse issueToken(UserInfo userInfo, String access
 
 		try {
 			record = personalAccessTokenDao.createTokenRecord(record);
-		} catch (DuplicateKeyException e) {
-			throw new IllegalArgumentException(DUPLICATE_TOKEN_NAME_MSG, e);
+		} catch (IllegalArgumentException e) {
+			if (e.getMessage().contains("org.springframework.dao.DuplicateKeyException")) {
```
brucehoff: In the interest of *strong typing* I would suggest an alternative fix, something like:
Define a class org.sagebionetworks.DuplicateKeyException that extends IllegalArgumentException.
In DBOBasicDaoImpl.insert, catch org.springframework.dao.DuplicateKeyException and throw org.sagebionetworks.DuplicateKeyException.
Here, in PersonalAccessTokenManagerImpl, catch org.sagebionetworks.DuplicateKeyException and throw an exception (either  org.sagebionetworks.DuplicateKeyException or IllegalArgumentException) with DUPLICATE_TOKEN_NAME_MSG (as currently shown at line 130).

brucehoff (changes requested):

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/authentication/PersonalAccessTokenManagerImpl.java
nickgros: I found elsewhere we're using `e.getCause() instanceof org.springframework.dao.DuplicateKeyException` to conditionally handle this type of error. I've switched my implementation to use this. Creating a new exception and changing the behavior of DBOBasicDaoImpl would have likely been a much larger change.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/authentication/PersonalAccessTokenManagerImpl.java
```diff
@@ -125,8 +125,12 @@ public AccessTokenGenerationResponse issueToken(UserInfo userInfo, String access
 
 		try {
 			record = personalAccessTokenDao.createTokenRecord(record);
-		} catch (DuplicateKeyException e) {
-			throw new IllegalArgumentException(DUPLICATE_TOKEN_NAME_MSG, e);
+		} catch (IllegalArgumentException e) {
+			if (e.getCause() instanceof DuplicateKeyException) {
```
brucehoff, line 129: I double checked that this works if e.getCause() == null

brucehoff (approved):

PR #4235, opened 2020-12-04
code review changes

john-hill:


lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/download/Sort.json
```diff
@@ -0,0 +1,13 @@
+{
+	"description": "Identifies both the field and direction for a single sort operation",
+	"properties": {
+		"field": {
+			"description": "..",
```
marcomarasca, line 5: Might want to fill this out :)

marcomarasca (approved):

PR #4239, opened 2020-12-11
use absolute  for root schemas for both basica an validation schemas

john-hill:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/schema/JsonSchemaManagerImpl.java
```diff
@@ -64,6 +64,8 @@
 
 	public static final int MAX_ORGANZIATION_NAME_CHARS = 250;
 	public static final int MIN_ORGANZIATION_NAME_CHARS = 6;
+	
+	public static final String ABSOLUTE_$ID_TEMPALTE = "https://repo-prod.prod.sagebase.org/repo/v1/schema/type/registered/%s";
```
marcomarasca: Since this is public information (The absolute URL prefix) we probably should expose it (e.g. a constant or a static function) and reuse it.

marcomarasca (approved):

marcomarasca (approved):

PR #4244, opened 2021-01-05
Plfm 6551 server

john-hill:


lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/migration/RangeChecksum.json
```diff
@@ -14,11 +14,11 @@
 			"type": "string"
 		},
 		"minimumId": {
-			"description": "The ID of the minimum row in this checksum range (inclusive).",
+			"description": "The ID of the minimum row in this checksum range.",
```
marcomarasca, line 17: We can remove this together with the maximum id as it's not used anymore in the migration client.

marcomarasca (approved):

PR #4250, opened 2021-01-12
changed password check and failed lockout to only update the database…

john-hill:
… when required.  also changed the lock increment to occur in the database so a transaction and select for update lock are not needed


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/auth/LoginLockoutStatusDaoImpl.java
```diff
@@ -0,0 +1,76 @@
...
+import org.sagebionetworks.util.ValidateArgument;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.dao.EmptyResultDataAccessException;
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public class LoginLockoutStatusDaoImpl implements LoginLockoutStatusDao {
+
+	public static final String CURRENT_TIME_AS_UNSIGNED = "CAST(UNIX_TIMESTAMP(CURRENT_TIMESTAMP(3)) * 1000 AS SIGNED)";
```
marcomarasca: Please add a comment on why we need the current_timestamp(3) (e.g. we need milliseconds precision) and why we need to cast the result of the unix_timestamp (e.g. it returns a decimal that cannot be used afterwards with the timestamps we are using that are integer).

marcomarasca: Please renamed, _UNSIGNED -> SIGNED

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/auth/LoginLockoutStatusDaoImpl.java
```diff
@@ -0,0 +1,76 @@
...
+		}
+	}
+
+	@NewWriteTransaction
+	@Override
+	public void incrementLockoutInfoWithNewTransaction(Long userId) {
+		ValidateArgument.required(userId, "userId");
+		String sql = "INSERT INTO " + TABLE_UNSUCCESSFUL_LOGIN_LOCKOUT + "(" + COL_UNSUCCESSFUL_LOGIN_KEY + ","
+				+ COL_UNSUCCESSFUL_LOGIN_COUNT + "," + COL_UNSUCCESSFUL_LOGIN_LOCKOUT_EXPIRATION_TIMESTAMP_MILLIS
+				+ ") VALUES (?, 1, " + CURRENT_TIME_AS_UNSIGNED + "+(1 << 1)) ON DUPLICATE KEY UPDATE "
```
marcomarasca: I would probably set the default lockout the first time around to a fixed amount: e.g. 2ms rather than (1 << 1)

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/auth/LoginLockoutStatusDaoImpl.java
```diff
@@ -0,0 +1,76 @@
...
+				+ COL_UNSUCCESSFUL_LOGIN_COUNT + " = " + COL_UNSUCCESSFUL_LOGIN_COUNT + "+1, "
+				+ COL_UNSUCCESSFUL_LOGIN_LOCKOUT_EXPIRATION_TIMESTAMP_MILLIS + " = " + CURRENT_TIME_AS_UNSIGNED
+				+ "+(1 << " + COL_UNSUCCESSFUL_LOGIN_COUNT + ")";
```
marcomarasca: Please add a comment about the exponential doubling of the the lockout time

marcomarasca: We should probably add a bound to the expression, e.g. (1 << MIN(COL_UNSUCCESSFUL_LOGIN_COUNT, 54)). 54 is around 200 days.

lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/auth/LoginLockoutStatusDaoImplTest.java
```diff
@@ -0,0 +1,181 @@
...
+	}
+	
+	@Test
+	public void testResetLockoutInfoWithNewTransactionWithNoData() {
+		// call under test
+		loginLockoutStatusDao.resetLockoutInfoWithNewTransaction(userOneId);
+		LockoutInfo info = loginLockoutStatusDao.getLockoutInfo(userOneId);
+		assertNotNull(info);
+		assertEquals(0L, info.getNumberOfFailedLoginAttempts());
+		assertTrue(info.getRemainingMillisecondsToNextLoginAttempt() < 0L);
```
marcomarasca: We should probably protect ourselves from potential really fast tests "< 0L" -> "<= 0L"

lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/auth/LoginLockoutStatusDaoImplTest.java
```diff
@@ -0,0 +1,181 @@
...
+		loginLockoutStatusDao.incrementLockoutInfoWithNewTransaction(userOneId);
+		LockoutInfo info = loginLockoutStatusDao.getLockoutInfo(userOneId);
+		assertNotNull(info);
+		assertEquals(1L, info.getNumberOfFailedLoginAttempts());
+		// call under test
+		loginLockoutStatusDao.resetLockoutInfoWithNewTransaction(userOneId);
+		info = loginLockoutStatusDao.getLockoutInfo(userOneId);
+		assertNotNull(info);
+		assertEquals(0L, info.getNumberOfFailedLoginAttempts());
+		assertTrue(info.getRemainingMillisecondsToNextLoginAttempt() < 0L);
```
marcomarasca: We should probably protect ourselves from potential really fast tests "< 0L" -> "<= 0L"

lib/models/src/main/java/org/sagebionetworks/repo/model/auth/LockoutInfo.java
```diff
@@ -0,0 +1,79 @@
...
+ * 
+ *
+ */
+public class LockoutInfo {
+
+	/**
+	 * The number of failed login attempts for this user since the last successful
+	 * login.
+	 */
+	private Long numberOfFailedLoginAttempts;;
```
marcomarasca: I know they are free, but probably not needed ;

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/UserCredentialValidatorImpl.java
```diff
@@ -76,33 +56,50 @@ public boolean checkPassword(final long userId, final String password) {
...
+			lockoutDao.resetLockoutInfoWithNewTransaction(userId);
 		}
 
+		if (!credentialsCorrect) {
+			/*
+			 * When the wrong credentials are provided the lockout information is
+			 * incremented.
+			 */
+			lockoutDao.incrementLockoutInfoWithNewTransaction(userId);
+		}
```
marcomarasca, line 95: Probably it's better (for readibility):

```
boolean inccorrectCrenditals = !checkPassword();

if (inccorrectCrenditals) {
	/*
		* When the wrong credentials are provided the lockout information is
		* incremented.
		*/
	lockoutDao.incrementLockoutInfoWithNewTransaction(userId);
} else if (lockoutInfo.getNumberOfFailedLoginAttempts() > 0) {
	/*
		* The user was previously locked out but has now correctly authenticated so a
		* one-time reset of the previous lockout information is needed.
		*/
	lockoutDao.resetLockoutInfoWithNewTransaction(userId);
}
```

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/UserCredentialValidatorImplTest.java
```diff
@@ -1,151 +1,195 @@
...
-import org.sagebionetworks.repo.manager.loginlockout.LoginLockoutStatus;
 import org.sagebionetworks.repo.manager.loginlockout.UnsuccessfulLoginLockoutException;
 import org.sagebionetworks.repo.model.auth.AuthenticationDAO;
+import org.sagebionetworks.repo.model.auth.LockoutInfo;
+import org.sagebionetworks.repo.model.auth.LoginLockoutStatusDao;
 import org.sagebionetworks.securitytools.PBKDF2Utils;
 
-@RunWith(MockitoJUnitRunner.class)
+@ExtendWith(MockitoExtension.class)
 public class UserCredentialValidatorImplTest {
```
marcomarasca: Can we call this UnitTest?

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/UserCredentialValidatorImplTest.java
```diff
@@ -1,151 +1,195 @@
...
 	}
 
 	@Test
-	public void checkPasswordWithThrottling_WrongPassword(){
-		when(mockAuthDAO.checkUserCredentials(userId, PBKDF2Utils.hashPassword(password, salt))).thenReturn(false);
-
-		//method under test
-		assertFalse(authenticationManagerUtil.checkPasswordWithThrottling(userId, password));
-
+	public void testCheckPasswordWithThrottlingWithNoLocksValidCredentials() {
```
marcomarasca: Seems to be the same as testCheckPasswordWithThrottlingWithNoLock?

marcomarasca (approved):

marcomarasca:
org.sagebionetworks.client.exceptions.SynapseServiceUnavailable: {"reason":"Service temporarily unavailable, please try again later."}
	at org.sagebionetworks.IT990AuthenticationController.testChangePasswordWithOldPassword(IT990AuthenticationController.java:138)

marcomarasca (changes requested):
As discussed, during the change password the user group is updated in an outer transaction. The FK in the login lockout creates a deadlock, so we can drop the FK to solve the issue.

marcomarasca (approved):

PR #4262, opened 2021-01-21
restored ddl for session token including a unique index on the token

john-hill:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/persistence/DBOSessionToken.java
```diff
@@ -66,7 +96,7 @@ public MigrationType getMigratableTableType() {
 
 	@Override
 	public MigratableTableTranslation<DBOSessionToken, DBOSessionToken> getTranslator() {
-		return new BasicMigratableTableTranslation<DBOSessionToken>();	
+		return new BasicMigratableTableTranslation<DBOSessionToken>();
```
marcomarasca, line 99: This might break migration since we changed the validatedOn from a Date field to a Long. Potential solutions:
1. Introduce a DBOSessionTokenBackup object with the Date field and do the transformation here (Later on we will have to somehow get rid of that).
2. Keep the Date field here? But it might not work with the BasicDao on create/update
3. Use a DATETIME instead of long in the DB DDL itself and change everything to take this into account (e.g. all the queries? might not change anything).
4. Leave the Annotation Based DBO but add the unique index in the "constraints" property of the Table annotation

marcomarasca (changes requested):
See comment about migration.

PR #4269, opened 2021-01-27
PLFM-6582: Return unversioned DOI for Tables in unversioned bundle request

nickgros:
For versionable entities, by default we return the DOI for the most recent version if no version is specified in the entity bundle request.

This is the wrong thing to do for tables because the current version is mutable. Instead, we return the unversioned DOI. A user can still retrieve the versioned DOI in a bundle by specifying the table version.

### DIFF
--- a/services/repository/src/main/java/org/sagebionetworks/repo/web/service/EntityBundleServiceImpl.java
+++ b/services/repository/src/main/java/org/sagebionetworks/repo/web/service/EntityBundleServiceImpl.java
@@ -8,10 +8,8 @@ import java.util.List;
 
 import org.sagebionetworks.repo.manager.UserManager;
 import org.sagebionetworks.repo.manager.dataaccess.AccessRequirementManager;
-import org.sagebionetworks.repo.model.ACCESS_TYPE;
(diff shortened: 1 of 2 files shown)


services/repository/src/main/java/org/sagebionetworks/repo/web/service/EntityBundleServiceImpl.java
```diff
@@ -160,8 +158,9 @@ public EntityBundle getEntityBundle(Long userId, String entityId,
 		}
 		if(isTrue(request.getIncludeDOIAssociation()) ){
 			try {
-				if (versionNumber == null && (entity instanceof VersionableEntity)) {
-					// DOIs on VersionableEntity cannot be versionless, so we want to get the DOI for the current version
+				if (versionNumber == null && (entity instanceof VersionableEntity) && !(entity instanceof Table)) {
```
john-hill: This works.  In my opinion files are the special case.  If it is a file we use the current version, for all other cases we use the version number.

john-hill (approved):

PR #4288, opened 2021-02-24
PLFM-6616 Apply limit and offset to getFavoritesEntityHeader

nickgros:
The limit and offset parameters were not in the SQL statement.

I am maintaining the current behavior of `getTotalResults`, which does return the total number of results. For performance reasons, in other places we have changed the behavior to return something like `offset + numberOfPageResults + 1` if there are more results. Let me know if this is necessary here.

### DIFF
--- a/lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/dao/DBOFavoriteDAOImplAutowiredTest.java
+++ b/lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/dao/DBOFavoriteDAOImplAutowiredTest.java
@@ -262,24 +262,33 @@ public class DBOFavoriteDAOImplAutowiredTest {
 		favoritesToDelete.add(fav1created);
 		Favorite fav2created = favoriteDao.add(fav2);
 		favoritesToDelete.add(fav2created);
-		
+
 		PaginatedResults<EntityHeader> favs = favoriteDao.getFavoritesEntityHeader(creatorUserGroupId.toString(), Integer.MAX_VALUE, 0);
 
 		assertEquals(2, favs.getTotalNumberOfResults());		
 		assertEquals(2, favs.getResults().size());
 		
-		EntityHeader eh1 = null;
-		EntityHeader eh2 = null;
-		for(EntityHeader eh : favs.getResults()) {
-			if(eh.getName().equals(node1Name)) eh1 = eh;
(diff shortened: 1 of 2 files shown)


lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/dao/DBOFavoriteDAOImplAutowiredTest.java
```diff
@@ -262,24 +262,33 @@ public void testGetFavoritesEntityHeader() throws Exception {
...
 		assertEquals(2, favs.getResults().size());
 		
-		EntityHeader eh1 = null;
-		EntityHeader eh2 = null;
-		for(EntityHeader eh : favs.getResults()) {
-			if(eh.getName().equals(node1Name)) eh1 = eh;
-			if(eh.getName().equals(node2Name)) eh2 = eh;
-		}
+		EntityHeader eh1 = favs.getResults().get(0);
+		EntityHeader eh2 = favs.getResults().get(1);
```
nickgros, line 272: I added an `ORDER BY id` so the order results are returned should be deterministic, based on entity ID. In this test, the entities are created synchronously, so this order should always hold. The `for` loop is no longer necessary.

brucehoff: I suggest adding a test in which pagination is not trivial/default, e.g., make offset=1 and page size=1 and make sure you get exactly one page back.

brucehoff (changes requested):
I suggest adding a test in which pagination is not trivial/default, e.g., make offset=1 and page size=1 and make sure you get exactly one page back.

lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/dao/DBOFavoriteDAOImplAutowiredTest.java
nickgros: I think I capture what you're talking about in lines 288-291. If that's not what you're looking for, I need some further clarification

brucehoff: Yes, you are right.  My mistake, I didn't see those two tests.

brucehoff (approved):

PR #4291, opened 2021-03-01
PLFM-6583: Add `isLatestVersion` field to versionable entities

nickgros:


john-hill (commented):
Nick,
Why does it need to be added to EntityHeader, VersionInfo, and VersionableEntity?

nickgros:
> Why does it need to be added to EntityHeader, VersionInfo, and VersionableEntity?

These objects all contain version-related fields for entities (version number, version label, version comment). To me, it makes sense to bundle 'isLatestVersion' with these other fields. It's important version information.

john-hill (approved):

PR #4296, opened 2021-03-03
fully remove EvaluationStatus (PLFM-6621)

zimingd:


zimingd:
Staging will need to be altered with:
```
ALTER TABLE `jdoevaluation`
CHANGE COLUMN `STATUS` `STATUS` INT NULL DEFAULT NULL ;
```

marcomarasca (approved):

PR #4302, opened 2021-03-04
support ISO 8601 time format in tables (PLFM-5843)

zimingd:


lib/models/src/main/java/org/sagebionetworks/repo/model/table/parser/DateParser.java
```diff
@@ -1,29 +0,0 @@
-package org.sagebionetworks.repo.model.table.parser;
-
-
-import java.util.Date;
-
-import org.sagebionetworks.repo.model.table.ValueParser;
-
-public class DateParser implements ValueParser {
```
zimingd, line 8: This class was unused except in its own test

john-hill (approved):

PR #4303, opened 2021-03-05
Plfm 6508 i

john-hill:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/DDLUtilsImpl.java
```diff
@@ -102,11 +105,40 @@ public static String loadSQLFromClasspath(String fileName) {
...
+		StringBuilder builder = new StringBuilder();
+		boolean withinMultiLineComment = false;
+		for (int i=0; i < sql.length(); i++) {
+			char c = sql.charAt(i);
+			if(c == '/' && sql.charAt(i+1) == '*') {
+				withinMultiLineComment = true;
+				i++;
+				continue;
+			}
+			if(c == '*' && sql.charAt(i+1) == '/') {
```
marcomarasca, line 131: Careful with index out of bound here (and up)

marcomarasca (approved):

PR #4306, opened 2021-03-08
remove UPLOAD from all places except two

john-hill:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/sts/StsManagerImpl.java
```diff
@@ -103,8 +103,8 @@ public StsCredentials getTemporaryCredentials(UserInfo userInfo, String entityId
 		authManager.canAccess(userInfo, entityId, ObjectType.ENTITY, ACCESS_TYPE.DOWNLOAD)
 				.checkAuthorizationOrElseThrow();
 		if (permission == StsPermission.read_write) {
-			// For write permissions, we also need upload access.
-			authManager.canAccess(userInfo, entityId, ObjectType.ENTITY, ACCESS_TYPE.UPLOAD)
+			// For write permissions, we also need update access.
+			authManager.canAccess(userInfo, entityId, ObjectType.ENTITY, ACCESS_TYPE.UPDATE)
```
marcomarasca, line 107: Maybe we should check for both CREATE and DELETE as well

marcomarasca (approved):

PR #4327, opened 2021-03-27
PLFM-6658 Fix Jira integration

brucehoff:
This change references the component in the SG Jira project by ID rather than by name, since that's less likely to change.  Also, I made the error message more verbose so logs will be more informative in the future.


xschildw (approved):
Good idea to print status and body!

PR #4337, opened 2021-04-08
DW

xschildw:


services/workers/src/main/java/org/sagebionetworks/object/snapshot/worker/utils/AclObjectRecordWriter.java
```diff
@@ -51,6 +56,7 @@ public static AclRecord buildAclRecord(AccessControlList acl, ObjectType ownerTy
 	@Override
 	public void buildAndWriteRecords(ProgressCallback progressCallback, List<ChangeMessage> messages) throws IOException {
 		List<ObjectRecord> toWrite = new LinkedList<ObjectRecord>();
+		List<AclKinesisLogRecord> kinesisToWrite = new LinkedList<AclKinesisLogRecord>();
```
marcomarasca, line 59: We can use an array list with the messages.size() instead

lib/lib-audit/src/main/java/org/sagebionetworks/audit/utils/ObjectRecordBuilderUtils.java
```diff
@@ -1,5 +1,6 @@
 package org.sagebionetworks.audit.utils;
 
+import org.sagebionetworks.kinesis.AwsKinesisLogRecord;
```
marcomarasca, line 3: Unused

lib/lib-audit/src/main/java/org/sagebionetworks/audit/kinesis/ObjectRecordLoggerImpl.java
```diff
@@ -0,0 +1,18 @@
+package org.sagebionetworks.audit.kinesis;
+
+import org.sagebionetworks.kinesis.AwsKinesisFirehoseLogger;
+import org.sagebionetworks.kinesis.AwsKinesisLogRecord;
+import org.springframework.beans.factory.annotation.Autowired;
+
+import java.util.List;
+
+public class ObjectRecordLoggerImpl implements ObjectRecordLogger {
```
marcomarasca, line 9: We do not need this abstraction for now, let us add it later if needed

lib/lib-audit/src/main/java/org/sagebionetworks/audit/kinesis/AclKinesisLogRecordUtils.java
```diff
@@ -0,0 +1,11 @@
+package org.sagebionetworks.audit.kinesis;
+
+import org.sagebionetworks.repo.model.audit.AclRecord;
+
+public class AclKinesisLogRecordUtils {
+
+	public static AclKinesisLogRecord buildAclKinesisLogRecord(AclRecord aclRecord, long timestamp) {
```
marcomarasca, line 7: This can be moved in the client using it

PR #4342, opened 2021-04-14
fix dissappearance of expected error when resizing maximimSize of STRING_LIST (PLFM-6690)

zimingd:


zimingd:
I upgraded my own dev stack's MySQL to 8.0.23 to test the build, but other tests are failing. So, I reverted back to 8.0.16 to make sure thos changes won't break existing tests. To fully upgrade the version, we'll need to make more changes

john-hill (approved):

PR #4348, opened 2021-04-25
PLFM-6709: Throw IllegalArgumentException instead of DuplicateKeyException

xschildw:
[PLFM-6709]: This should return a 400 instead of a 500 called with duplicate label.

[PLFM-6709]: https://sagebionetworks.jira.com/browse/PLFM-6709

### DIFF
--- a/lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/NodeDAOImpl.java
+++ b/lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/NodeDAOImpl.java
@@ -122,6 +122,7 @@ import org.sagebionetworks.util.ValidateArgument;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.dao.DataIntegrityViolationException;
+import org.springframework.dao.DuplicateKeyException;
 import org.springframework.dao.EmptyResultDataAccessException;
 import org.springframework.jdbc.BadSqlGrammarException;
 import org.springframework.jdbc.core.BatchPreparedStatementSetter;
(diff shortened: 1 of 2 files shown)


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/NodeDAOImpl.java
```diff
@@ -2039,9 +2040,13 @@ public long snapshotVersion(Long userId, String nodeIdString, SnapshotRequest re
...
 		String label = request.getSnapshotLabel() != null ? request.getSnapshotLabel() : revisionNumber.toString();
-		this.jdbcTemplate.update(SQL_CREATE_SNAPSHOT_VERSION, request.getSnapshotComment(), label,
-				request.getSnapshotActivityId(), userId, modifiedOn, nodeId, revisionNumber);
-		return revisionNumber;
+		try {
+			this.jdbcTemplate.update(SQL_CREATE_SNAPSHOT_VERSION, request.getSnapshotComment(), label,
+					request.getSnapshotActivityId(), userId, modifiedOn, nodeId, revisionNumber);
+			return revisionNumber;
+		} catch (DuplicateKeyException e) {
+			throw new IllegalArgumentException(e);
```
marcomarasca: The exception thrown might not be friendly enough, we should probably check also which DuplicateKey was violated (e.g. might be something else than label, if id for example it should be a 500).

lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/dao/NodeDAOImplTest.java
```diff
@@ -4051,6 +4051,43 @@ public void testSnapshotVersion() throws InterruptedException {
...
+		SnapshotRequest request2 = new SnapshotRequest();
+		request2.setSnapshotComment("different comment");
+		request2.setSnapshotLabel("some label");
+		request2.setSnapshotActivityId(testActivity2.getId());
+
+		// call under test: this should fail
+		String id = node.getId();
+		assertThrows(
+			IllegalArgumentException.class, () -> {nodeDao.snapshotVersion(user1Id, id, request2);}
+		);
```
marcomarasca, line 4090: Check the error message and exception returned by the assertThrows

marcomarasca (changes requested):

marcomarasca (approved):

PR #4357, opened 2021-05-04
PLFM-6713: ReadOnly handling for scanner + cleanup

marcomarasca:


lib/models/src/main/java/org/sagebionetworks/repo/model/exception/RecoverableException.java
```diff
@@ -1,7 +1,8 @@
 package org.sagebionetworks.repo.model.exception;
 
 /**
- * Exception throw for operations that can be retried
+ * Exception throw for operations that can be retried, this can be used as an alternative the checked RecoverableMessageException when implementing an interface that
+ * cannot work with checked exceptions (e.g. An iterable)
```
john-hill, line 5: can we instead change RecoverableMessageException to be runtime?

john-hill (approved):

PR #4366, opened 2021-05-07
PLFM-4309 Documentation improvements

brucehoff:


services/repository/src/main/java/org/sagebionetworks/repo/web/controller/CertifiedUserController.java
```diff
@@ -169,13 +169,21 @@ public void deleteQuizResponse(
...
+	/**
+	 * Set a user's certification status.  For Synapse administrators only.
+	 * @param userId
+	 * @param principalId the ID of the user whose status is to be set.
+	 * @param isCertified true to set as certified or false to 'de-certify'
+	 * @throws NotFoundException
+	 */
 	@RequiredScope({modify})
 	@ResponseStatus(HttpStatus.NO_CONTENT)
 	@RequestMapping(value = UrlHelpers.ADMIN + UrlHelpers.CERTIFIED_USER_STATUS, method = RequestMethod.PUT)
```
marcomarasca, line 189: If we move this to the AdministrationController I think it is excluded from the public docs

services/repository/src/main/java/org/sagebionetworks/repo/web/controller/PrincipalController.java
```diff
@@ -167,7 +167,7 @@ public void additionalEmailValidation(
 	@RequestMapping(value = { UrlHelpers.EMAIL }, method = RequestMethod.POST)
 	public void addEmail(
 			@RequestParam(value = AuthorizationConstants.USER_ID_PARAM) Long userId,
-			@RequestParam(value = AuthorizationConstants.SET_AS_NOTIFICATION_EMAIL_PARM) Boolean setAsNotificationEmail,
+			@RequestParam(value = AuthorizationConstants.SET_AS_NOTIFICATION_EMAIL_PARM, required=false) Boolean setAsNotificationEmail,
```
marcomarasca, line 170: We can probably add defaultValue="false", even though I guess this handled down at the service layer

marcomarasca (approved):

PR #4370, opened 2021-05-12
create a manifest to capture all backup request information including…

john-hill:
… the secondary types used by the source


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/migration/MigrationManagerImpl.java
```diff
@@ -413,19 +420,43 @@ public BackupTypeResponse backupRequest(UserInfo user, BackupTypeRangeRequest re
...
-		return backupStreamToS3(request.getMigrationType(), dataStream, request.getAliasType(), request.getBatchSize());
+		return backupStreamToS3(request.getMigrationType(), dataStream, request.getAliasType(), request.getBatchSize(), backupKey);
 	}
 	
+	/**
+	 * Put the given manifest to S3.
+	 * @param manifestKey
+	 * @param manifest
+	 */
+	void putManifestToS3(String backupKey, BackupManifest manifest) {
```
marcomarasca, line 445: Maybe let us rename backupKey to backupKeyPrefix or move the line 447 up above so that it is clear that the string in input is the actual S3 key

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/migration/MigrationManagerImpl.java
```diff
@@ -526,19 +587,21 @@ public RestoreTypeResponse restoreRequest(UserInfo user, RestoreTypeRequest requ
 	 * @return
 	 */
 	@Override
-	public RestoreTypeResponse restoreStream(InputStream input, MigrationType primaryType,
-			BackupAliasType backupAliasType, long batchSize) {
+	public RestoreTypeResponse restoreStream(InputStream input, BackupManifest manifest) {
 		RestoreTypeResponse response = new RestoreTypeResponse();
-		if(!this.migratableTableDao.isMigrationTypeRegistered(primaryType)) {
+		if(!this.migratableTableDao.isMigrationTypeRegistered(manifest.getOriginalRequest().getMigrationType())) {
```
marcomarasca, line 592: We might have an edge case: if we want to promote a secondary to primary, but get rid of the original primary this will lead to not migrating any of the secondary. We might be able to avoid this doing this check in the loop below.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/migration/MigrationManagerImpl.java
```diff
@@ -526,19 +587,21 @@ public RestoreTypeResponse restoreRequest(UserInfo user, RestoreTypeRequest requ
 	 * @return
 	 */
 	@Override
-	public RestoreTypeResponse restoreStream(InputStream input, MigrationType primaryType,
-			BackupAliasType backupAliasType, long batchSize) {
+	public RestoreTypeResponse restoreStream(InputStream input, BackupManifest manifest) {
```
marcomarasca, line 590: This might be protected and not exposed in the interface

lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/migration/BackupManifest.json
```diff
@@ -0,0 +1,15 @@
+{
+	"description": "Manifest of all metadata associated with a single backup file",
+	"properties": {
+		"originalRequest": {
+			"$ref": "org.sagebionetworks.repo.model.migration.BackupTypeRangeRequest"
+		},
+		"secondaryTypes": {
```
marcomarasca, line 7: We might need an additional field containing the logical backup id for secondaries. This might be necessary because if we turn a secondary to a primary the new object will most likely have a different backup id but the restore should still use the old backup id.

marcomarasca (approved):

PR #4390, opened 2021-05-23
PLFM-6751

brucehoff:
This change allows a client to pass an access token (normally passed as a Bearer authorization header) as a "sessionToken" header.  The change keeps Synapse from breaking embedded app's that take the session token cookie and use the contents to authenticate requests to Synapse.

### DIFF
--- a/lib/models/src/main/java/org/sagebionetworks/repo/model/auth/JSONWebTokenHelper.java
+++ b/lib/models/src/main/java/org/sagebionetworks/repo/model/auth/JSONWebTokenHelper.java
@@ -15,11 +15,14 @@ import org.sagebionetworks.repo.model.oauth.JsonWebKeySet;
 import org.sagebionetworks.util.ValidateArgument;
 
 import io.jsonwebtoken.Claims;
+import io.jsonwebtoken.ExpiredJwtException;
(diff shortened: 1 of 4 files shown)


lib/models/src/main/java/org/sagebionetworks/repo/model/auth/JSONWebTokenHelper.java
```diff
@@ -94,7 +97,11 @@ public static RSAPublicKey getRSAPublicKeyForJsonWebKeyRSA(JsonWebKeyRSA jwkRsa)
 	}
 	
 	public static String getSubjectFromJWTAccessToken(String accessToken) {
-		return getUnsignedJWTFromToken(accessToken).getBody().getSubject();
+		try {
+			return getUnsignedJWTFromToken(accessToken).getBody().getSubject();
+		} catch (ExpiredJwtException | UnsupportedJwtException | MalformedJwtException | SignatureException | IllegalArgumentException e) {
```
marcomarasca, line 102: Maybe the Illegal should not be wrapped, also isn't there a common jwt ex?

marcomarasca (approved):

PR #4417, opened 2021-06-10
Plfm 6697

john-hill:


marcomarasca (changes requested):
Using the storage location id might not be what the user wants: given the constraints the users simply needs to know which files can be packaged and which ones cannot. Maybe can be based on a boolean instead. The storage location metadata is not public information and cannot be fetched if the user is not the creator.

PR #4425, opened 2021-06-21
PLFM-6408

brucehoff:
Update discovery document to say we don't support request parameter


jay-hodgson (approved):

jay-hodgson:
Unsigned request objects are unsupported (alg:none).  Thanks Bruce!

PR #4427, opened 2021-06-22
PLFM-6764: Re-link files during scanning

marcomarasca:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/file/FileHandleAssociationScannerJobManagerImpl.java
```diff
@@ -32,21 +36,24 @@
 		
 	static final long FLUSH_DELAY_MS = 250;
 	static final int KINESIS_BATCH_SIZE = 10000;
+	static final int RELINK_BACTH_SIZE = 1000;
```
john-hill, line 39: remove

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/file/FileHandleAssociationScannerJobManagerImpl.java
```diff
@@ -180,17 +193,40 @@ private static int getMessageDelay() {
...
+		
+		if (recordsBatch.isEmpty()) {
+			return 0;
+		}
+		
+		Set<Long> fileHandleIds = recordsBatch.stream()
+				.map(FileHandleAssociationRecord::getFileHandleId)
+				.collect(Collectors.toSet());
+		
+		List<Long> updatedIds = fileHandleDao.updateBatchStatus(new ArrayList<>(fileHandleIds), FileHandleStatus.AVAILABLE, FileHandleStatus.UNLINKED, 0);
```
john-hill, line 221: try a  query to find one row that would need updated from the batch, then do the update, we minimize write locks.  Also it should be super rare to have a hit.

john-hill (approved):

PR #4447, opened 2021-07-09
PLFM-6344

williamvietnguyen:


lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/TableIndexDAOImpl.java
```diff
@@ -144,6 +144,8 @@
 	private static final String KEY = "Key";
 	private static final String SQL_SHOW_COLUMNS = "SHOW FULL COLUMNS FROM ";
 	private static final String FIELD = "Field";
+	
+	private static final Long DEFAULT_MAX_STRING_SIZE = 50L;
```
marcomarasca: We can use the existing ColumnConstants.DEFAULT_STRING_SIZE instead.

lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/TableIndexDAOImpl.java
```diff
@@ -1183,8 +1185,12 @@ ColumnMetadata translateColumnModel(ColumnModel model, ObjectFieldModelResolver
 				}
 
 				model.setColumnType(type);
-				if(ColumnType.STRING == type || ColumnType.STRING_LIST==type) {
-					model.setMaximumSize(aggregation.getMaxStringElementSize());
+				if (ColumnType.STRING == type || ColumnType.STRING_LIST == type) {
+					if (aggregation.getMaxStringElementSize() == 0) {
```
marcomarasca: null check or 0L

lib/lib-table-cluster/src/test/java/org/sagebionetworks/table/cluster/TableIndexDAOImplUnitTest.java
```diff
@@ -208,4 +212,97 @@ public void testValidateMaxListLengthInAnnotationReplication_allUnderLimit() {
...
+		// call under test
+		List<ColumnModel> results = TableIndexDAOImpl.expandFromAggregation(Lists.newArrayList(zero,one,two,three));
+		assertEquals(6, results.size());
+		
+		// case 1: string column type with 0L max size, should give 50L max size
+		ColumnModel cm = results.get(0);
+		assertEquals("foo", cm.getName());
+		assertEquals(ColumnType.STRING, cm.getColumnType());
+		assertEquals(null, cm.getMaximumListLength());
+		assertEquals(50L, cm.getMaximumSize());
```
marcomarasca: Instead of 50L might be better to use the ColumnConstants.DEFAULT_STRING_SIZE value

marcomarasca (approved):

marcomarasca (approved):

PR #4448, opened 2021-07-15
Plfm 6819

john-hill:


lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/model/SQLElement.java
```diff
@@ -170,4 +167,73 @@ public boolean hasQuotesRecursive(){
...
+		}
+	}
+	
+	/**
+	 * Helper to build a children Iterable.
+	 * @param children
+	 * @return
+	 */
+	static <T extends Element> Iterable<Element> buildChildren(List<T> children){
+		if(children == null || children.isEmpty()) {
```
marcomarasca, line 197: This should turn around and call the buildChildren with the array of the list (careful with nulls)

lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/model/SQLElement.java
```diff
@@ -170,4 +167,73 @@ public boolean hasQuotesRecursive(){
...
+			return list;
+		}
+	}
+	
+	/**
+	 * Helper to build a children Iterable.
+	 * @param children
+	 * @return
+	 */
+	static <T extends Element> Iterable<Element> buildChildren(List<T> children){
```
marcomarasca, line 196: ```<? extends Element>```

lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/model/SQLElement.java
```diff
@@ -41,21 +44,30 @@ public String toString() {
...
+	 * @param <T>
+	 * @param list
+	 * @param type
+	 * @param element
 	 */
-	public <T extends Element> Iterable<T> createIterable(Class<T> type){
-		LinkedList<T> list = new LinkedList<T>();
-		checkElement(list, type, this);
-		return list;
+	static <T extends Element> void addRecurisve(List<T> list, Class<T> type, Element element) {
```
marcomarasca, line 64: addRecurisve -> addRecursive

lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/model/ArrayHasLikePredicate.java
```diff
@@ -48,11 +47,10 @@ public void toSql(StringBuilder builder, ToSqlParameters parameters) {
...
 		}
 	}
-	
+		
 	@Override
-	<T extends Element> void addElements(List<T> elements, Class<T> type) {
-		super.addElements(elements, type);
-		checkElement(elements, type, escapeCharacter);
+	public Iterable<Element> getChildren() {
+		return SQLElement.buildChildren(columnReferenceLHS, inPredicateValue, escapeCharacter);
```
marcomarasca, line 53: I would probably try to keep the super and return the concat the iterables

lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/model/IntervalLiteral.java
```diff
@@ -28,8 +43,8 @@ public void toSql(StringBuilder builder, ToSqlParameters parameters) {
 	}
 
 	@Override
-	<T extends Element> void addElements(List<T> elements, Class<T> type) {
-		// this is a leaf
```
marcomarasca, line 32: It might be a leaf for a "good" reason, please check the SQLTranslatorUtils and its exception for IntervalLiteral (or the other way around, e.g. maybe we do not need the exception in the SQLTranslatorUtils)

lib/lib-table-query/src/test/java/org/sagebionetworks/table/query/ArrayHasLikePredicateTest.java
```diff
@@ -76,4 +76,13 @@ public void testHasPredicateWithEscape() throws ParseException{
 		
 		assertEquals(expectedValues, values);
 	}
+	
+	@Test
+	public void testGetChidren() throws ParseException {
```
marcomarasca, line 81: Maybe a test without the escape character?

marcomarasca (approved):

lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/model/IntervalLiteral.java
john-hill: The spacial case for SQLTranslatorUtils for IntervalLiteral is still needed.  Since the child of IntervalLiteral is UnsignedInteger and not UnsignedLiteral, including the child in getChildren() does not change anything.

PR #4459, opened 2021-07-26
Plfm 6752

john-hill:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/entity/EntityAuthorizationManagerImpl.java
```diff
@@ -142,6 +143,7 @@ public UserEntityPermissions getUserPermissionsForEntity(UserInfo userInfo, Stri
 		permissions.setCanEnableInheritance(
 				determineCanDeleteACL(userInfo, stateProvider.getPermissionsState(entityIdLong))
```
marcomarasca, line 144: We might not need this call as we have it already up above.

marcomarasca (approved):

PR #4505, opened 2021-09-27
PLFM-6934: Handle annotations where type in the schema is defined by reference.

williamvietnguyen:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/schema/AnnotationsTranslatorImpl.java
```diff
@@ -410,22 +410,40 @@ void writeAnnotationValue(String key, AnnotationsValue value, JSONObject jsonObj
...
-							result.put(key, true);
 						}
 					}
 				}
 			}
 		}
 		return result;
 	}
+	
+	// returns true if an entry is added to the result, false otherwise
```
john-hill, line 433: void

john-hill (approved):

PR #4515, opened 2021-10-08
PLFM-6986: Add datasets entry to managerMap

nickgros:
Adding this to the configuration prevents error here when calling certain table services on a dataset: https://github.com/Sage-Bionetworks/Synapse-Repository-Services/blob/780f716dacdec2fc7cec6741bcfe1f87b2e00733/services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/table/TableTransactionManagerProviderImpl.java#L29


services/repository-managers/src/main/resources/private/managers-spb.xml
```diff
@@ -393,6 +393,9 @@
 				<entry key="submissionview">
 					<ref bean="tableViewTransactionManager" />
 				</entry>
+				<entry key="dataset">
+					<ref bean="tableViewTransactionManager" />
+				</entry>
```
nickgros: Move configuration to code and add integration test to DatasetIntegrationTest

services/repository-managers/src/main/resources/private/managers-spb.xml
```diff
@@ -382,19 +382,6 @@
...
-				</entry>
-				<entry key="entityview">
-					<ref bean="tableViewTransactionManager" />
-				</entry>
-				<entry key="submissionview">
-					<ref bean="tableViewTransactionManager" />
-				</entry>
-			</map>
-		</property>
 	</bean>
```
nickgros: Wasn't sure if I should remove the whole bean, so I only removed the map. I think replacing all of the XML in this file with annotations is a different task.

john-hill: I agree that the rest of the file is beyond this task.  However, we should move tableTransactionManagerProvider out of the xml.

john-hill (approved):
Looks good!  Thanks.

PR #4525, opened 2021-11-05
Plfm 6964

john-hill:


lib/lib-table-query/src/test/java/org/sagebionetworks/table/query/model/QualifiedJoinTest.java
```diff
@@ -1,15 +1,49 @@
...
+		QualifiedJoin join = new TableQueryParser("right outer join tableB on tableA.id = tableB.id").qualifiedJoin(lhs);
+		assertEquals("tableA RIGHT OUTER JOIN tableB ON tableA.id = tableB.id",join.toSql());
+	}
+	
+	@Test
+	public void testParseWithNoOn() throws ParseException {
+		TableReference lhs = new TableQueryParser("tableA").tableReference();
+		QualifiedJoin join = new TableQueryParser("join tableB").qualifiedJoin(lhs);
+		assertEquals("tableA JOIN tableB",join.toSql());
+	}
```
marcomarasca, line 46: Maybe test to check that outer by itself is not valid

marcomarasca (approved):

PR #4535, opened 2021-11-18
PLFM-6967: Materialized views CRUD

marcomarasca:


lib/lib-table-query/src/main/resources/table-query-parser.jj
```diff
@@ -32,7 +32,13 @@ public class TableQueryParser {
 	public static QuerySpecification parserQuery(String query)
 			throws ParseException {
 		TableQueryParser parser = new TableQueryParser(new StringReader(query));
-		return parser.querySpecification();
+		QuerySpecification parsedQuery = parser.querySpecification();
```
john-hill: we can remove the optional brackets from:
	<select> [ setQuantifier = setQuantifier()] selectList = selectList() [ tableExpression = tableExpression() ] <EOF>
That will require a table expression.

john-hill (changes requested):
See comment

john-hill (approved):

PR #4545, opened 2021-12-02
extended the scope limit to 30K and exclude empty containers from scope

john-hill:


lib/jdomodels/src/main/resources/sql/GetAllContainerIds.sql
```diff
@@ -0,0 +1,6 @@
+WITH RECURSIVE CONTAINERS (ID) AS (
+	 SELECT ID FROM JDONODE WHERE ID IN (:ids_param) AND NODE_TYPE IN ('project','folder')
+	 UNION DISTINCT
+	 SELECT N.ID FROM CONTAINERS AS C JOIN JDONODE AS N ON (C.ID = N.PARENT_ID AND N.NODE_TYPE IN ('project','folder'))
+)
+SELECT C.ID FROM CONTAINERS C LEFT JOIN JDONODE J ON C.ID = J.PARENT_ID WHERE J.ID IS NOT NULL LIMIT :bLimit
```
marcomarasca: Should be a DISTINCT

marcomarasca: Let also add a test

marcomarasca (changes requested):

marcomarasca (approved):

PR #4548, opened 2021-12-07
code review changes part one

john-hill:


lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/SQLTranslatorUtils.java
```diff
@@ -397,15 +457,25 @@ private static void replaceTextMatchesPredicate(BooleanPrimary booleanPrimary) {
...
-		IdAndVersion originalSynId = IdAndVersion
-				.parse(fromClause.getSingleTableName().orElseThrow(TableConstants.JOIN_NOT_SUPPORTED_IN_THIS_CONTEXT));
-		//replace from clause
-		fromClause.setTableReference(tableReferenceForName(SQLUtils.getTableNameForId(originalSynId, TableType.INDEX)));
-		return originalSynId;
+	static Optional<TableNameCorrelation> translateTableName(TableNameCorrelation tableNameCorrelation, TableAndColumnMapper mapper) {
+		Optional<TableInfo> optional = mapper.lookupTableNameCorrelation(tableNameCorrelation);
+		if(!optional.isPresent()) {
+			return Optional.empty();
+		}else {
```
marcomarasca, line 464: we don'[t need the else

lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/TableInfo.java
```diff
@@ -137,6 +137,23 @@ public String getTranslatedTableAlias() {
...
 	}
+	
+	/**
+	 * Does the passed TableNameCorrelation match this TableInfo?
+	 * 
+	 * @param tableNameCorrelation
+	 * @return
+	 */
+	public boolean isMatch(TableNameCorrelation tableNameCorrelation) {
+		String tableNameSql = tableNameCorrelation.getTableName().toSql();
```
marcomarasca, line 148: We might need to match on the alias too since the same table might be referenced more than once (in which case it would match always on the first TableInfo)

lib/lib-table-cluster/src/test/java/org/sagebionetworks/table/cluster/SQLTranslatorUtilsTest.java
```diff
@@ -749,77 +744,63 @@ public void testReadWithHeadersWithoutEtagRow() throws SQLException{
...
+		assertTrue(optional.isPresent());
+		assertEquals("T123_456",optional.get().toSql());
 	}
 	
 	@Test
-	public void testTranslateSelectDouble() throws ParseException{
-		ValueExpressionPrimary column = new TableQueryParser("aDouble").valueExpressionPrimary();
-		SQLTranslatorUtils.translateSelect(column, columnMap);
-		assertEquals("CASE WHEN _DBL_C777_ IS NULL THEN _C777_ ELSE _DBL_C777_ END", column.toSql());
+	public void testTranslateTableNameWithJoin() throws ParseException {
```
marcomarasca, line 775: Let us add a test where we have joins with tables with the same name (potentially not subsequent index)

marcomarasca (approved):

PR #4610, opened 2022-02-18
Plfm 7001.2

john-hill:


lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/description/MaterializedViewIndexDescription.java
```diff
@@ -92,12 +94,18 @@ public EntityType getTableType() {
 	}
 
 	@Override
-	public List<String> getColumnNamesToAddToSelect(SqlContext context, boolean includeEtag) {
+	public List<String> getColumnNamesToAddToSelect(SqlContext context, boolean includeEtag, boolean isAggregate) {
 		ValidateArgument.required(context, "SqlContext");
 		switch (context) {
 		case build:
+			if(isAggregate && !buildColumnsToAddToSelect.isEmpty()) {
```
nickgros, line 101: // explain why we throw this error
boolean hasRowLevelBenefactors = !buildColumnsToAddToSelect.isEmpty()

nickgros (approved):
Looks great!

PR #4618, opened 2022-03-01
Plfm 6956

john-hill:


lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/dao/NodeDAOImplTest.java
```diff
@@ -1610,16 +1611,23 @@ public void testGetEntityHeaderByReference() throws Exception {
 		assertEquals(parent.getModifiedOn(), header.getModifiedOn());
 		assertTrue(header.getIsLatestVersion());
 		
+		header = results.get(1);
+		assertEquals(childId, header.getId());
+		assertEquals("two", header.getVersionLabel());
+		assertEquals(new Long(2), header.getVersionNumber());
+		assertEquals(parentBenefactor, header.getBenefactorId());
+		assertTrue(header.getIsLatestVersion());
```
marcomarasca, line 1625: Let us add a test to reproduce the case when a non-existing version is supplied, and a result was given back (we should not get back a result instead).

marcomarasca (approved):

marcomarasca (approved):

PR #4641, opened 2022-04-07
Compile SWC dependencies to Java 8

nickgros:
This is the minimum set of dependencies I needed to have compile to Java 8 to get a successful SWC build

### DIFF
--- a/client/synapseJavaClient/pom.xml
+++ b/client/synapseJavaClient/pom.xml
@@ -107,6 +107,14 @@
 			</resource>
 		</resources>
 		<plugins>
+            <!-- Compile to Java 8 for compatibility with SWC -->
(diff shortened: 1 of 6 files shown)


marcomarasca (approved):
Thanks @nickgros! I'm tentatively accepting this, but I do have a question: the synapse java client also depends on lib-communicationUtilities and lib-securityUtilities. I wonder if those are unused? Is so I would probably remove them from the dependencies, if not we should probably be cautious and also target 1.8 for those so that we do not mistakenly code against > 1.8.

nickgros:
You are right that those are unused, and I'll be removing them from SWC's dependencies when I upgrade to the latest version of the dependencies that we are using.

PR #4670, opened 2022-06-14
PLFM-7307: check for null accessType

xschildw:
If null,then throw IllegalArgumentException.
Note1: need doc change as the parameter is now marked optional.
Note2: invalid value throws an IllegalArgumentException, albeit with the defult message. Add check as well?


services/repository/src/main/java/org/sagebionetworks/repo/web/service/EntityServiceImpl.java
```diff
@@ -493,6 +493,7 @@ public void deleteEntityACL(Long userId, String id)
 	public boolean hasAccess(String entityId, Long userId, String accessType)
 		throws NotFoundException, DatastoreException, UnauthorizedException {
 		UserInfo userInfo = userManager.getUserInfo(userId);
+		if (accessType == null) throw new IllegalArgumentException("AccessType cannot be null");
```
john-hill: {

}

john-hill (approved):

PR #4683, opened 2022-06-30
add null benefactors for user, as query remove them

SandhraSokhal:


services/workers/src/test/java/org/sagebionetworks/table/worker/MaterializedViewUpdateWorkerIntegrationTest.java
```diff
@@ -355,6 +355,68 @@ public void testMaterializedViewWithLeftJoinViews() throws Exception {
...
+				new Row().setRowId(2L).setVersionNumber(0L).setValues(Arrays.asList(fileIds.get(1), "222", "222")),
+				new Row().setRowId(3L).setVersionNumber(0L).setValues(Arrays.asList(fileIds.get(2), "111", null)),
+				new Row().setRowId(4L).setVersionNumber(0L).setValues(Arrays.asList(fileIds.get(3), "222", "222")),
+				new Row().setRowId(5L).setVersionNumber(0L).setValues(Arrays.asList(fileIds.get(4), "111", null))
+		);
+
+
+		// Wait for the query against the materialized view to have the expected results.
+		asyncHelper.assertQueryResult(userInfo, materializedQuery, (results) -> {
+			System.out.println(" result size "+ results.getQueryResult().getQueryResults().getRows().size());
```
john-hill: comment can be removed.

services/workers/src/test/java/org/sagebionetworks/table/worker/MaterializedViewUpdateWorkerIntegrationTest.java
```diff
@@ -355,6 +355,68 @@ public void testMaterializedViewWithLeftJoinViews() throws Exception {
...
+			assertEquals(expectedRows, results.getQueryResult().getQueryResults().getRows());
+		}, MAX_WAIT_MS);
+	}
```
john-hill, line 515: Lets add a test that covers the case where the user does not have permission to see some files on the left or the right.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/table/TableQueryManagerImpl.java
```diff
@@ -678,8 +678,16 @@ QuerySpecification addRowLevelFilter(UserInfo user, QuerySpecification query, In
 			} catch (BadSqlGrammarException e) { // table has not been created yet
 				tableBenefactors = Collections.emptySet();
 			}
-			// Get the sub-set of benefactors visible to the user.
+
+			boolean hasNullBenefactors =tableBenefactors.stream().anyMatch(Objects::isNull);
```
john-hill: Per discussion, we think that using a default of -1 would be a cleaner implementation that would avoid awkward database nulls.

john-hill (changes requested):

john-hill (approved):

PR #4685, opened 2022-07-07
allow enable of derived annotations

john-hill:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/schema/DBOJsonSchemaBindObject.java
```diff
@@ -42,6 +44,7 @@
 	private String objectType;
 	private Long createdBy;
 	private Timestamp createdOn;
+	private Boolean enableDerived;
```
marcomarasca, line 47: We need a piece of migration code to explicitly check if this value is null, in which case we set it as false (otherwise we end up with null values in the DB).

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/EntityManagerImpl.java
```diff
@@ -588,10 +588,14 @@ public JsonSchemaObjectBinding bindSchemaToEntity(UserInfo userInfo, BindSchemaT
 		ValidateArgument.required(request, "request");
 		ValidateArgument.required(request.getEntityId(), "request.entityId");
 		ValidateArgument.required(request.getSchema$id(), "request.schema$id");
+		boolean enableDerived = false;
+		if(request.getEnableDerivedAnnotations() != null) {
```
marcomarasca, line 592: Boolean.TRUE.equals(request.getEnableDerivedAnnotations())?

marcomarasca (changes requested):

marcomarasca (approved):

PR #4714, opened 2022-08-26
PLFM-7435: Added boolean canRequestMembership to Team class

harveyx22:


lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/Team.json
```diff
@@ -24,6 +24,10 @@
 			"type": "boolean",
 			"description": "true for teams which members can join without an invitation or approval"
 		},
+		"canRequestMembership": {
+			"type": "boolean",
+			"description": "true for teams where users can create a membership request to join"
```
marcomarasca: I would also add the default value if not present

lib/models/src/main/java/org/sagebionetworks/repo/model/principal/BootstrapTeam.java
```diff
@@ -41,6 +43,15 @@ public Boolean getCanPublicJoin() {
 	public void setCanPublicJoin(Boolean canPublicJoin) {
 		this.canPublicJoin = canPublicJoin;
 	}
+
+	public Boolean getCanRequestMembership() {
```
marcomarasca: We need also to modify the manager-spb.xml to reflect this for the boostrapped teams, or put a default value for it (which probably is already done in the manager)

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/team/TeamManagerImpl.java
```diff
@@ -849,6 +849,7 @@ public void bootstrapTeams() throws NotFoundException {
 				newTeam.setId(team.getId());
 				newTeam.setName(team.getName());
 				newTeam.setCanPublicJoin(team.getCanPublicJoin());
+				newTeam.setCanRequestMembership(team.getCanRequestMembership());
```
marcomarasca: We need to make sure that when a team is created (or updated), if the canRequestMembership flag is not present we assign a default value (true iff canPublicJoin is false or null, false if true).

What about the case when canPublicJoin is true but canRequestMemebership is true? Shouldn't that throw an exception?

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/team/MembershipRequestManagerImpl.java
```diff
@@ -113,6 +113,10 @@ public MembershipRequest create(UserInfo userInfo,
 			throw new IllegalArgumentException("This team is already open for the public to join, membership requests are not needed.");
 		}
 
+		if (team.getCanRequestMembership() != null && !team.getCanRequestMembership()) {
```
marcomarasca: Consider adding an enumeration (that is not exposed in the API), that translates the state of the two booleans into the state of the team E.g. (OPEN, PUBLIC, CLOSED) that is saved to the DB. Then instead of checking the booleans in the business logic we can work at the status level. Note that this status property also need to be migrated (but first we need to make sure we migrate the new requestMembership boolean).

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/persistence/DBOTeam.java
```diff
@@ -74,8 +77,24 @@ public Class<? extends DBOTeam> getDBOClass() {
...
 
 	};
-	
-	private static final MigratableTableTranslation<DBOTeam, DBOTeam> MIGRATION_MAPPER = new BasicMigratableTableTranslation<>();
+
+	@TemporaryCode(author = "peter.harvey@sagebase.org", comment = "One time migration of property canRequestMembership.  Can be removed after all teams have a canRequestMembership value.")
+	private static final MigratableTableTranslation<DBOTeam, DBOTeam> MIGRATION_MAPPER = new  MigratableTableTranslation<DBOTeam, DBOTeam>() {
+		@Override
+		public DBOTeam createDatabaseObjectFromBackup(DBOTeam backup) {
+			Team dto = TeamUtils.copyDboToDto(backup);
```
marcomarasca: Consider adding an enumeration (that is not exposed in the API), that translates the state of the two booleans into the state of the team E.g. (OPEN, PUBLIC, CLOSED) that is saved to the DB. Then instead of checking the booleans in the business logic we can work at the status level. Note that this status property also need to be migrated (but first we need to make sure we migrate the new requestMembership boolean).

marcomarasca (changes requested):

lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/persistence/DBOTeamTest.java
```diff
@@ -83,4 +89,54 @@ public void testRoundTrip() throws DatastoreException, NotFoundException, Unsupp
...
+	@TemporaryCode(author = "peter.harvey@sagebase.org", comment = "One time migration of state.  Can be removed after all teams have a state value.")
+	@Test
+	public void testCreateDatabaseObjectFromBackupWithPublicTeam() {
+		Team dto = new Team();
+		dto.setId("123");
+		dto.setEtag(UUID.randomUUID().toString());
+		dto.setCanPublicJoin(true);
+		dto.setCanRequestMembership(false);
+		DBOTeam dbo = new DBOTeam();
+		TeamUtils.copyDtoToDbo(dto, dbo);
```
marcomarasca, line 101: The utility will set the state here, so we are bypassing the actual testing of createDatabaseObjectFromBackup. Instead you can use the TeamUtils.copyToSerializedField to take the team and put it in the properties field of the DBOTeam object.

lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/persistence/DBOTeamTest.java
```diff
@@ -83,4 +89,54 @@ public void testRoundTrip() throws DatastoreException, NotFoundException, Unsupp
...
+		dto.setCanPublicJoin(false);
+		dto.setCanRequestMembership(false);
+		DBOTeam dbo = new DBOTeam();
+		TeamUtils.copyDtoToDbo(dto, dbo);
+
+		// Call under test
+		dbo = dbo.getTranslator().createDatabaseObjectFromBackup(dbo);
+
+		assertEquals(TeamState.CLOSED.name(), dbo.getState());
+	}
```
marcomarasca, line 157: We do need a test where the state is already set, to make sure we are not re-setting each time

lib/models/src/main/java/org/sagebionetworks/repo/model/TeamState.java
```diff
@@ -0,0 +1,35 @@
...
+    OPEN,
+    // A team is public when a user can join without creating a membership request
+    PUBLIC,
+    // A team is closed when a user cannot create a membership request
+    CLOSED;
+
+    public static TeamState from(Team team) {
+        boolean canPublicJoin = Boolean.TRUE.equals(team.getCanPublicJoin());
+        boolean canRequestMembership;
+        if (canPublicJoin) {
```
marcomarasca: I would add some comments why we do this (e.g. to maintain backward compatibility with existing clients etc)

lib/models/src/main/java/org/sagebionetworks/repo/model/TeamState.java
```diff
@@ -0,0 +1,35 @@
...
+            canRequestMembership = Boolean.TRUE.equals(team.getCanRequestMembership());
+        } else {
+            canRequestMembership = !Boolean.FALSE.equals(team.getCanRequestMembership());
```
marcomarasca: In this cases it might be easier to read if we have an explicit null check (e.g. if it is null we choose a default value according to the value of canPublicJoin)

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/team/MembershipRequestManagerImpl.java
```diff
@@ -108,11 +109,15 @@ public MembershipRequest create(UserInfo userInfo,
 			if (hasUnmetAccessRequirements(userInfo, mr.getTeamId()))
 				throw new UnauthorizedException("Requested member has unmet access requirements which must be met before asking to join the Team.");
 		}
-		Team team = teamDAO.get(mr.getTeamId());
-		if (team.getCanPublicJoin() != null && team.getCanPublicJoin()) {
+
+		if (teamDAO.getState(mr.getTeamId()).equals(TeamState.PUBLIC)) {
```
marcomarasca: I would switch the equals check here (TeamSteam.PUBLIC.equals(...))

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/team/TeamManagerImpl.java
```diff
@@ -851,7 +851,9 @@ public void bootstrapTeams() throws NotFoundException {
 				newTeam.setCanPublicJoin(team.getCanPublicJoin());
 				newTeam.setDescription(team.getDescription());
 				newTeam.setIcon(team.getIcon());
-				newTeam = bootstrapCreate(newTeam);	
+				newTeam.setCanRequestMembership(team.getCanRequestMembership());
```
marcomarasca: This is not needed for now

lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/dao/TeamUtilsTest.java
```diff
@@ -3,10 +3,13 @@
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 import java.util.Date;
+import java.util.UUID;
 
 import org.junit.jupiter.api.Test;
 import org.sagebionetworks.repo.model.Team;
+import org.sagebionetworks.repo.model.dbo.persistence.DBOAccessRequirement;
 import org.sagebionetworks.repo.model.dbo.persistence.DBOTeam;
+import org.sagebionetworks.util.TemporaryCode;
```
marcomarasca: We should add to this test that the state enum is added to the DBO

lib/models/src/main/java/org/sagebionetworks/repo/model/TeamDAO.java
```diff
@@ -184,4 +184,14 @@ public Team update(Team team) throws InvalidModelException,
...
+
+	/**
+	 * Retrieves the state of a team given its id
+	 *
+	 * @param id
+	 * @return
+	 * @throws DatastoreException
+	 * @throws NotFoundException
+	 */
+	public TeamState getState(String id) throws DatastoreException, NotFoundException;
```
marcomarasca: public modifier is not needed in an interface

lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/dao/DBOTeamDAOImplTest.java
```diff
@@ -643,4 +644,20 @@ private void bindTeamName(String name, Long teamId){
...
 	}
+
+	@Test
+	public void testGetState() {
+		UserGroup group = new UserGroup();
+		group.setId(userGroupDAO.create(group).toString());
+		teamsToDelete.add(group.getId());
+
+		Team team = new Team();
+		Long id = Long.parseLong(group.getId());
```
marcomarasca: why not use group.getId()?

lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/dao/DBOTeamDAOImplTest.java
```diff
@@ -643,4 +644,20 @@ private void bindTeamName(String name, Long teamId){
...
+		team.setId("" +id);
+		team.setCanPublicJoin(true);
+		team.setCanRequestMembership(false);
+		teamDAO.create(team);
+		// Call under test
+		assertEquals(TeamState.PUBLIC, teamDAO.getState(team.getId()));
```
marcomarasca, line 670: We need a test for the case when the team does not exist. Also you should probably have a test for each of the state values (e.g. to test the persistence of the state)

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/DBOTeamDAOImpl.java
```diff
@@ -196,6 +198,8 @@ public class DBOTeamDAOImpl implements TeamDAO {
 
 	private static final String SELECT_CHECK_TEAM_EXISTS = "SELECT COUNT(*) FROM " + TABLE_TEAM + " WHERE " + COL_TEAM_ID + " = ?";
 
+	private static final String SELECT_TEAM_STATE = "SELECT " + COL_TEAM_STATE + " FROM " + TABLE_TEAM + " WHERE " + COL_TEAM_ID + "=:"+COL_TEAM_ID;
```
marcomarasca: You probably do not need a named parameter for this

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/DBOTeamDAOImpl.java
```diff
@@ -652,4 +656,11 @@ public List<String> getAllTeamsUserIsAdmin(String userId) {
 	public void truncateAll() {
 		jdbcTemplate.update("DELETE FROM " + TABLE_TEAM);
 	}
+
+	@Override
+	public TeamState getState(String teamId) throws DatastoreException, NotFoundException {
+		MapSqlParameterSource param = new MapSqlParameterSource();
+		param.addValue(COL_TEAM_ID, teamId);
+		return namedJdbcTemplate.queryForObject(SELECT_TEAM_STATE, param, TeamState.class);
```
marcomarasca: This would throw a EmptyResultDataAccessException if the query returned no rows, we can catch that and throw a NotFoundException

marcomarasca (changes requested):

lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/dao/TeamUtilsTest.java
```diff
@@ -39,4 +37,21 @@ public void testRoundTrip() {
...
+		dto.setCreatedOn(new Date());
+		dto.setCanPublicJoin(false);
+		dto.setCanRequestMembership(true);
+
+		DBOTeam dbo = new DBOTeam();
+
+		// call under test
+		TeamUtils.copyDtoToDbo(dto, dbo);
+
+		assertEquals(dbo.getState(), TeamState.OPEN.name());
```
marcomarasca, line 55: The "expected" value is the first argument

marcomarasca (approved):

PR #4716, opened 2022-08-29
PaginationIterator stops getting pages when done

john-hill:


lib/lib-utils/src/main/java/org/sagebionetworks/util/PaginationIterator.java
```diff
@@ -26,16 +27,25 @@ public PaginationIterator(PaginationProvider<T> provider, long limit) {
...
+		if(isDone) {
+			return false;
+		}
 		if(currentPage == null || !currentPage.hasNext()) {
 			// Fetch the next page
 			List<T> page = provider.getNextPage(limit, offset);
 			offset += limit;
 			currentPage = page.iterator();
-			return currentPage.hasNext();
+			boolean hasNext =  currentPage.hasNext();
```
marcomarasca, line 43: If you check the page.size() < limit you avoid querying once more when there are results and you reached the last page

marcomarasca (approved):

PR #4726, opened 2022-09-08
PLFM-6275

nickgros:
- Add 'additionalFilters' to TransformSqlWithFacetsRequest object
- Support transforming QueryFilters to Synapse SQL
- Refactor to reuse less complex code

Draft while pending API design review

### DIFF
--- a/lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/util/FacetUtils.java
+++ b/lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/util/FacetUtils.java
@@ -1,10 +1,17 @@
 package org.sagebionetworks.table.query.util;
 
 import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
 
+import org.sagebionetworks.repo.model.table.ColumnMultiValueFunctionQueryFilter;
+import org.sagebionetworks.repo.model.table.ColumnSingleValueQueryFilter;
+import org.sagebionetworks.repo.model.table.QueryFilter;
+import org.sagebionetworks.repo.model.table.TextMatchesQueryFilter;
 import org.sagebionetworks.table.query.ParseException;
 import org.sagebionetworks.table.query.TableQueryParser;
 import org.sagebionetworks.table.query.model.QuerySpecification;
+import org.sagebionetworks.table.query.model.TextMatchesPredicate;
 import org.sagebionetworks.table.query.model.WhereClause;
 import org.sagebionetworks.util.ValidateArgument;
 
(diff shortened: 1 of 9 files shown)


lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/util/FacetUtils.java
```diff
@@ -15,39 +22,157 @@ public class FacetUtils {
...
-					builder.append(searchCondition);
-				}
-			}
-		}
+
+		String joinedConditions = conditionStrings.stream()
+				// Remove nulls
+				.filter(s -> s != null && !s.isEmpty())
+				// Join with " AND "
+				.collect(Collectors.joining(" AND "));
```
nickgros, line 77: IMO this is much more clear than the imperative style loop

lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/util/FacetUtils.java
```diff
@@ -15,39 +22,157 @@ public class FacetUtils {
...
+	 */
+	private static String concatConditionStrings(List<String> conditionStrings) {
+		ValidateArgument.required(conditionStrings, "conditionStrings");
+
 		StringBuilder builder = new StringBuilder();
 		builder.append("(");
 		int initialSize = builder.length(); //length with the "( " included
-		
-		for(FacetRequestColumnModel facetColumn : validatedFacets){
-			if(columNameToIgnore == null || !facetColumn.getColumnName().equals(columNameToIgnore)){ //make sure not to include the ignored column
```
nickgros, line 30: `columnNameToIgnore` is now filtered in `concatFacetSearchConditionStrings`

lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/util/FacetUtils.java
```diff
@@ -15,39 +22,157 @@ public class FacetUtils {
 	 * @param columNameToIgnore the name of the column to exclude from the concatenation
 	 * @return the concatenated string or null if there was nothing to concatenate
 	 */
-	public static String concatFacetSearchConditionStrings(List<FacetRequestColumnModel> validatedFacets, String columNameToIgnore){
+	public static String concatFacetSearchConditionStrings(List<FacetRequestColumnModel> validatedFacets, Optional<String> columNameToIgnore){
```
john-hill, line 25: Cannot have Optional as method param
Can overload the signature.
In method can use
Name valueOption = Optional.ofNullable(param)

lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/util/FacetUtils.java
```diff
@@ -15,39 +22,157 @@ public class FacetUtils {
...
+
+		return concatConditionStrings(filters.stream().map(FacetUtils::getSearchConditionString).collect(Collectors.toList()));
+	}
+
+	/**
+	 * Concatenates a list of SQL conditional statements with " AND "
+	 * @param conditionStrings
+	 * @return
+	 */
+	private static String concatConditionStrings(List<String> conditionStrings) {
```
john-hill, line 66: This means there is no unit test for this.

lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/util/FacetUtils.java
```diff
@@ -15,39 +22,157 @@ public class FacetUtils {
...
+				.filter(facet -> {
+					if (columNameToIgnore.isPresent()) {
+						return !facet.getColumnName().equals(columNameToIgnore.get());
+					} else {
+						return true;
+					}
+				})
+				.collect(Collectors.toList());
+
+		List<String> conditionStrings = facetColumnsWithoutIgnoredColumn.stream().map(FacetRequestColumnModel::getSearchConditionString).collect(Collectors.toList());
```
john-hill, line 42: can this be a single line stream with a single line filter.

john-hill: ...filter(Not(facet.getColumnName().equal(columNameToIgnore)))

lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/util/FacetUtils.java
```diff
@@ -15,39 +22,157 @@ public class FacetUtils {
...
+		return builder.toString();
+	}
+
+	/**
+	 * Converts a QueryFilter to a Synapse SQL condition string
+	 * @param filter
+	 * @return a standalone condition, such as "( \"foo\" LIKE '%bar%' )"
+	 */
+	public static String getSearchConditionString(QueryFilter filter) {
+		if (filter instanceof ColumnSingleValueQueryFilter) {
```
john-hill, line 95: Find a way to move this logic out of a static util and into a PoJo with methods like toSql().

john-hill (changes requested):
We think this would be easier and cleaner to implement as an extension of the bundle.

john-hill:
We went another route

PR #4728, opened 2022-09-09
TableIndexManager is now a bean, to enable profiling, plus setIndexSc…

john-hill:
…hema() will now do nothing if the schema hash already matches the index hash


lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/TableIndexDAOImpl.java
```diff
@@ -389,6 +394,16 @@ public void setIndexVersionAndSchemaMD5Hex(IdAndVersion tableId, Long viewCRC, S
 		template.update(createOrUpdateStatusSql, viewCRC, schemaMD5Hex, viewCRC, schemaMD5Hex);
 	}
 	
+	@Override
+	public boolean doesIndexHashMatchSchemaHash(IdAndVersion tableId, List<ColumnModel> newSchema) {
+		ValidateArgument.required(tableId, "tableId");
+		ValidateArgument.required(newSchema, "newSchema");
+		String newSchemaHash = TableModelUtils
+				.createSchemaMD5Hex(newSchema.stream().map(ColumnModel::getId).collect(Collectors.toList()));
```
marcomarasca, line 402: Maybe just change the TableModelUtils.createSchemaMD5Hex to take thje list of columnMOdel?

marcomarasca (approved):

PR #4733, opened 2022-09-14
PLFM-7235: Fix for CVE-2021-33813

harveyx22:


marcomarasca (changes requested):
While it seems to work just updating to jdom2, from the migration documentation (https://github.com/hunterhacker/jdom/wiki/JDOM2-Migration-Issues) it looks like this is not enough. In particular they mention that some methods were deprecated in jdom2 and need to be changed. An example is the SAXBuilder.setValidation. Note that even though jdom was had the CVE our code is not affected (we are not using jdom for user input) so it is not strictly necessary to upgrade (but it'd be a good thing to be on the latest).

harveyx22:
Deprecated method setValidation replaced per http://www.jdom.org/docs/apidocs/org/jdom2/input/SAXBuilder.html#setValidation-boolean-

pom.xml
```diff
@@ -635,7 +635,7 @@
 			<dependency>
 				<groupId>org.sagebionetworks</groupId>
 				<artifactId>database-semaphore</artifactId>
-				<version>3.0.6</version>
+				<version>3.0.5</version>
```
marcomarasca: Why are we reverting all the versions?

marcomarasca (changes requested):
Looks like you reverted the database-semaphore, aws-utilities and simple-http-client version

marcomarasca (approved):

PR #4737, opened 2022-09-15
PLFM-7471: Upgrade to MySQL 8.0.30 

marcomarasca:
Depends on https://github.com/Sage-Bionetworks/Synapse-Stack-Builder/pull/482

### DIFF
--- a/lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/SQLUtils.java
+++ b/lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/SQLUtils.java
@@ -1017,10 +1017,13 @@ public class SQLUtils {
 			if(!isFirst){
 				builder.append(", ");
 			}
-			// There is no need to run the actual count for columns for which an index is not created 
+			// There is no need to run a distinct count for columns for which an index is not created 
 			// or for metadata columns (such as row id) that manage their own indices
+			// Using MAX with a constant is relatively cheap, note that when there are no rows in the table MAX will return NULL
 			if (info.isMetadata() || !info.getType().isCreateIndex()) {
+				builder.append("MAX(");
 				builder.append(TableConstants.COLUMN_NO_CARDINALITY);
+				builder.append(")");
 			} else {
 				builder.append("COUNT(DISTINCT ");
 				builder.append(info.getColumnName());
--- a/lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/SQLUtils.java
+++ b/lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/SQLUtils.java
@@ -105,7 +105,6 @@ public class SQLUtils {
 	public static final String COLUMN_PREFIX = "_C";
 	public static final String COLUMN_POSTFIX = "_";
 	public static final String UNNEST_SUFFIX = "_UNNEST";
-
 	private static final String DOUBLE_NAN = Double.toString(Double.NaN);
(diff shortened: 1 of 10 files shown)


lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/SQLUtils.java
```diff
@@ -1019,9 +1017,16 @@ public static String createCardinalitySql(List<DatabaseColumnInfo> list, IdAndVe
 			if(!isFirst){
 				builder.append(", ");
 			}
-			builder.append("COUNT(DISTINCT ");
-			builder.append(info.getColumnName());
-			builder.append(") AS ");
+			// There is no need to run the actual count for columns for which an index is not created 
+			// or for metadata columns (such as row id) that manage their own indices
+			if (info.isMetadata() || !info.getType().isCreateIndex()) {
+				builder.append(TableConstants.COLUMN_NO_CARDINALITY);
```
john-hill, line 1025: if we use max(1) then we always get one row back even for an empty table.  This should be super cheap.

lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/TableIndexDAOImpl.java
```diff
@@ -652,7 +652,11 @@ public void provideCardinality(final List<DatabaseColumnInfo> list,
 		if(list.isEmpty()){
 			return;
 		}
+		// If the table is empty, the query might not return any row if no count is performed for any column
+		list.forEach(info -> info.setCardinality(TableConstants.COLUMN_NO_CARDINALITY));
```
john-hill: will not need with the max(1) in the query.

lib/lib-table-cluster/src/test/java/org/sagebionetworks/table/cluster/SQLUtilsTest.java
```diff
@@ -1126,14 +1128,57 @@ public void testCreateCardinalitySqlEmpty(){
...
+			DatabaseColumnInfo column = new DatabaseColumnInfo();
+			column.setColumnName(name);
+			list.add(column);
+		}
+		list.sort(Comparator.comparing(DatabaseColumnInfo::getColumnName));
+		
+		String results = SQLUtils.createCardinalitySql(list, tableId);
+		
+		assertEquals(
+			"SELECT -1 AS ROW_BENEFACTOR, -1 AS ROW_ETAG, -1 AS ROW_ID, -1 AS ROW_SEARCH_CONTENT, -1 AS ROW_VERSION FROM T999", results
```
john-hill: this will use max

lib/lib-table-cluster/src/test/java/org/sagebionetworks/table/cluster/TableIndexDAOImplTest.java
```diff
@@ -1529,6 +1590,193 @@ public void testColumnInfoAndCardinality(){
...
+				
+		List<DatabaseColumnInfo> infoList = getAllColumnInfo(tableId);
+		tableIndexDAO.optimizeTableIndices(infoList, tableId, 4);
+		infoList = getAllColumnInfo(tableId);
+		assertEquals(5, infoList.size());
+		
+		DatabaseColumnInfo info = infoList.get(0);
+		
+		// ROW_ID
+		assertEquals("ROW_ID", info.getColumnName());
```
john-hill, line 1618: assertEquals(expected, result)

john-hill (changes requested):

PR #4738, opened 2022-09-15
try to run with concurrent true and lots of threads

john-hill:


services/workers/src/main/resources/table-cluster-sqs-spb.xml
```diff
@@ -42,12 +42,12 @@
 					</bean>
 				</property>
 				<property name="targetMethod" value="run" />
-				<property name="concurrent" value="false" />
+				<property name="concurrent" value="true" />
 			</bean>
 		</property>
 		<!-- We stager the start delay of each trigger to spread out the timing -->
 		<property name="startDelay" value="256" />
```
marcomarasca, line 49: With such a long repeatInterval we need to increase the start delay (since it might misfire the first time it starts up) to avoid having longer tests.

services/workers/src/main/resources/table-cluster-sqs-spb.xml
```diff
@@ -42,12 +42,12 @@
 					</bean>
 				</property>
 				<property name="targetMethod" value="run" />
-				<property name="concurrent" value="false" />
+				<property name="concurrent" value="true" />
```
marcomarasca, line 45: There is a chance that one machine would take all the semaphore locks defeating the distribution of workers in different machines

marcomarasca (changes requested):
You convinced yourself that this is not the way to go :)

PR #4740, opened 2022-09-16
PLFM-7476: Search index updates within transaction

marcomarasca:


lib/lib-table-cluster/src/test/java/org/sagebionetworks/table/cluster/TableIndexDAOImplTest.java
```diff
@@ -171,7 +173,7 @@ public ColumnType getBenefactorIdColumnType() {
 	public void after() {
 		// Drop the table
 		if (tableId != null && tableIndexDAO != null) {
-			tableIndexDAO.deleteTable(tableId);
+			//tableIndexDAO.deleteTable(tableId);
```
john-hill: oops

john-hill (changes requested):

PR #4741, opened 2022-09-22
PLFM-7194

harveyx22:


lib/lib-googlecloud/src/main/java/org/sagebionetworks/googlecloud/SynapseGoogleCloudStorageClientImpl.java
```diff
@@ -112,4 +112,9 @@ public Boolean bucketExists(String bucket) {
 		return storage.get(bucket, Storage.BucketGetOption.fields()) != null;
 	}
 
+	@Override
+	public Boolean doesObjectExist(String bucketName, String objectKey) {
+		return storage.get(BlobId.of(bucketName, objectKey)) != null;
```
john-hill, line 117: We might need an integration test that shows this works as expected.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/file/MultipartManagerV2Impl.java
```diff
@@ -115,16 +116,13 @@ <T extends MultipartRequest> MultipartUploadStatus startOrResumeMultipartRequest
 
 		// The MD5 is used to identify if this upload request already exists for this user.
 		String requestMD5Hex = MultipartRequestUtils.calculateMD5AsHex(request);
-
```
john-hill, line 118: if(!forceRestart && filleHandleExists, but storage object does not exist){
throw new IllegalArgumentException("You are attempting to re-upload a file that was already uploaded as FileHandle.id = 123.  However the file: <bucket>, <key> no longer exits in the cloud storage provider.  Please delete the Original FileHandle and try again.  Or attempt to re-upload again with forceRestart=true.")

We want them to fix the problem, not just silently sweep it under the rug.
}
An alternate solution would be to "fix" the broken file handle (not create a new one).  This would involve letting them re-upload the file, then change the broken file handle to point to the newly uploaded object.  This might be the best option because it allows them to truly fix what they broke.

john-hill (changes requested):
Our PM should decide on the best solution to this problem.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/file/MultipartManagerV2Impl.java
```diff
@@ -147,7 +145,18 @@ <T extends MultipartRequest> MultipartUploadStatus startOrResumeMultipartRequest
 		
 		return status.getMultipartUploadStatus();
 	}
-	
+
+	private boolean isNewUploadRequestRequired(boolean forceRestart, UserInfo user, String requestMD5Hex, CompositeMultipartUploadStatus status) {
```
john-hill: this method seems like a boolean question but it has an unexpected side-effect (it automatically re-starts the upload).

With a function that only returns a boolean (no side-effects):
then
if(forceRestart || fileHandleExistsFileNoLongerInCloudeStorage()){
        resetMultiPart()
}

john-hill (changes requested):
We still want the PM to decide on how we should handle this case.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/file/MultipartManagerV2Impl.java
```diff
@@ -6,6 +6,7 @@
 import java.util.List;
 import java.util.UUID;
 
+import com.google.cloud.storage.Acl;
```
marcomarasca: Not needed

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/file/MultipartManagerV2Impl.java
```diff
@@ -147,7 +150,13 @@ <T extends MultipartRequest> MultipartUploadStatus startOrResumeMultipartRequest
 		
 		return status.getMultipartUploadStatus();
 	}
-	
+
+	private boolean fileHandleExistsFileNoLongerInCloudStorage(UserInfo user, CompositeMultipartUploadStatus status) {
+		// If the existing upload request is complete, we check if the underlying object has been deleted
+		return (status != null && MultipartUploadState.COMPLETED.equals(status.getMultipartUploadStatus().getState())
+				&& !cloudDaoProvider.getCloudServiceMultipartUploadDao(status.getUploadType()).doesObjectExist(status.getBucket(), status.getKey()));
```
marcomarasca: For readability we should probably take the cloudDaoProvider.getCloudServiceMultipartUploadDao(status.getUploadType()) out in its own line

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/file/MultipartManagerV2Impl.java
```diff
@@ -115,16 +116,18 @@ <T extends MultipartRequest> MultipartUploadStatus startOrResumeMultipartRequest
...
-		if (forceRestart) {
-			multipartUploadDAO.setUploadStatusHash(user.getId(), requestMD5Hex, RESTARTED_HASH_PREFIX + requestMD5Hex + "_" + UUID.randomUUID().toString());
-		}
 
 		// Has an upload already been started for this user and request
 		CompositeMultipartUploadStatus status = multipartUploadDAO.getUploadStatus(user.getId(), requestMD5Hex);
-		
-		if (status == null) {
+
+		boolean fileHandleExistsFileNoLongerInCloudStorage = fileHandleExistsFileNoLongerInCloudStorage(user, status);
```
marcomarasca: I would rename this to a simpler name, e.g. isExistingUploadDataMissing?

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/file/MultipartManagerV2Impl.java
```diff
@@ -115,16 +116,18 @@ <T extends MultipartRequest> MultipartUploadStatus startOrResumeMultipartRequest
...
+
+		if (forceRestart || fileHandleExistsFileNoLongerInCloudStorage) {
+			// When forcing a restart or if the underlying object key of a file handle does not exist we clear the old hash changing the cache key, the previous multipart upload will be garbage collected by a worker
+			multipartUploadDAO.setUploadStatusHash(user.getId(), requestMD5Hex, RESTARTED_HASH_PREFIX + requestMD5Hex + "_" + UUID.randomUUID().toString());
```
marcomarasca, line 124: set the status to null so the line below does not have to recheck everything

marcomarasca: A better approach would be to move the "reset" of the upload hash when the status is missing. The forceRestart=true or the invalid existing upload can be treated the same way, e.g. the potentially existing upload is always considered invalid and we reset its hash unconditionally.

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/file/MultipartManagerV2ImplAutowireTest.java
```diff
@@ -546,7 +551,77 @@ public void testMultipartUploadCopyFromMultipartUploadWithMultiplePartsSinglePar
...
+		S3FileHandle sourceFileHandle = (S3FileHandle) doMultipartUpload(fileName, contentType, fileContent, storageLocationId, useContentTypeForParts, true);
+
+		s3Client.deleteObject(sourceFileHandle.getBucketName(), sourceFileHandle.getKey());
+
+		assertFalse(s3Client.doesObjectExist(sourceFileHandle.getBucketName(), sourceFileHandle.getKey()));
+
+		// Call under test
+		S3FileHandle resultFileHandle = (S3FileHandle) doMultipartUpload(fileName, contentType, fileContent, storageLocationId, useContentTypeForParts, false);
+
+		assertNotEquals(sourceFileHandle, resultFileHandle);
```
marcomarasca: I would probably check the ids only

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/file/MultipartManagerV2ImplAutowireTest.java
```diff
@@ -546,7 +551,77 @@ public void testMultipartUploadCopyFromMultipartUploadWithMultiplePartsSinglePar
...
+		String contentType = "plain/text";
+		String fileContent = "This is the content of the file";
+		Long storageLocationId = null;
+		boolean useContentTypeForParts = false;
+
+		GoogleCloudFileHandle sourceFileHandle = (GoogleCloudFileHandle) doMultipartUpload(fileName, contentType, fileContent, storageLocationId, useContentTypeForParts, true);
+
+		googleCloudStorageClient.deleteObject(sourceFileHandle.getBucketName(), sourceFileHandle.getKey());
+
+		assertFalse(s3Client.doesObjectExist(sourceFileHandle.getBucketName(), sourceFileHandle.getKey()));
```
marcomarasca: this should use the GC client

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/file/MultipartManagerV2ImplAutowireTest.java
```diff
@@ -546,7 +551,77 @@ public void testMultipartUploadCopyFromMultipartUploadWithMultiplePartsSinglePar
...
+	public void testGCStartOrResumeMultipartRequestWithObjectDeleted() throws Exception {
+		Assume.assumeTrue(stackConfiguration.getGoogleCloudEnabled());
+
+		String fileName = "foo.txt";
+		String contentType = "plain/text";
+		String fileContent = "This is the content of the file";
+		Long storageLocationId = null;
+		boolean useContentTypeForParts = false;
+
+		GoogleCloudFileHandle sourceFileHandle = (GoogleCloudFileHandle) doMultipartUpload(fileName, contentType, fileContent, storageLocationId, useContentTypeForParts, true);
```
marcomarasca: this cannot possibly work

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/file/MultipartManagerV2ImplAutowireTest.java
```diff
@@ -546,7 +551,77 @@ public void testMultipartUploadCopyFromMultipartUploadWithMultiplePartsSinglePar
...
+	}
+
+	@Test
+	public void testGCStartOrResumeMultipartRequestWithObjectDeleted() throws Exception {
+		Assume.assumeTrue(stackConfiguration.getGoogleCloudEnabled());
+
+		String fileName = "foo.txt";
+		String contentType = "plain/text";
+		String fileContent = "This is the content of the file";
+		Long storageLocationId = null;
```
marcomarasca: To be a GC upload set this to googleCloudStorageLocationSetting.getStorageLocationId();

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/file/MultipartManagerV2ImplAutowireTest.java
```diff
@@ -546,7 +551,77 @@ public void testMultipartUploadCopyFromMultipartUploadWithMultiplePartsSinglePar
...
+
+		// Call under test
+		S3FileHandle resultFileHandle = (S3FileHandle) doMultipartUpload(fileName, contentType, fileContent, storageLocationId, useContentTypeForParts, false);
+
+		assertNotEquals(sourceFileHandle, resultFileHandle);
+	}
+
+	@Test
+	public void testGCStartOrResumeMultipartRequestWithObjectDeleted() throws Exception {
+		Assume.assumeTrue(stackConfiguration.getGoogleCloudEnabled());
```
marcomarasca, line 582: To make this work you have to add this to your build:
export org_sagebionetworks_google_cloud_enabled=true
export org_sagebionetworks_google_cloud_key='key' (See the repo-develop-build)

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/file/MultipartManagerV2ImplAutowireTest.java
```diff
@@ -546,7 +551,77 @@ public void testMultipartUploadCopyFromMultipartUploadWithMultiplePartsSinglePar
...
+		boolean useContentTypeForParts = false;
+
+		GoogleCloudFileHandle sourceFileHandle = (GoogleCloudFileHandle) doMultipartUpload(fileName, contentType, fileContent, storageLocationId, useContentTypeForParts, true);
+
+		googleCloudStorageClient.deleteObject(sourceFileHandle.getBucketName(), sourceFileHandle.getKey());
+
+		assertFalse(s3Client.doesObjectExist(sourceFileHandle.getBucketName(), sourceFileHandle.getKey()));
+
+		// Call under test
+		GoogleCloudFileHandle resultFileHandle = (GoogleCloudFileHandle) doMultipartUpload(fileName, contentType, fileContent, storageLocationId, useContentTypeForParts, false);
```
marcomarasca: Please take the forceRestart value out in its own variable so we know what the value when reading the code

lib/lib-googlecloud/src/test/java/org/sagebionetworks/googlecloud/SynapseGoogleCloudStorageClientImplUnitTest.java
```diff
@@ -181,6 +182,18 @@ public void testGetObjects() {
 		assertEquals(expected, actual);
 	}
 
+	@Test
+	public void testDoesObjectExist() {
```
marcomarasca: What about the case when this is null?

marcomarasca (changes requested):

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/file/MultipartManagerV2Impl.java
```diff
@@ -151,10 +140,25 @@ <T extends MultipartRequest> MultipartUploadStatus startOrResumeMultipartRequest
 		return status.getMultipartUploadStatus();
 	}
 
-	private boolean fileHandleExistsFileNoLongerInCloudStorage(UserInfo user, CompositeMultipartUploadStatus status) {
-		// If the existing upload request is complete, we check if the underlying object has been deleted
-		return (status != null && MultipartUploadState.COMPLETED.equals(status.getMultipartUploadStatus().getState())
-				&& !cloudDaoProvider.getCloudServiceMultipartUploadDao(status.getUploadType()).doesObjectExist(status.getBucket(), status.getKey()));
+	Optional<CompositeMultipartUploadStatus> getCachedUploadOrRestart(UserInfo user, MultipartRequest request, boolean forceRestart) {
```
marcomarasca: Instead of having the forceRestart in this function and resetting the hash here, the caller can unconditionally reset the hash when a new upload is created. The forceRestart option will make it so that we do not even check for a cached upload. The goal of this method is to only retrieve a potentially cached upload if any that is still valid exist.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/file/MultipartManagerV2Impl.java
```diff
@@ -114,35 +114,24 @@ <T extends MultipartRequest> MultipartUploadStatus startOrResumeMultipartRequest
...
-		if (status == null || forceRestart || fileHandleExistsFileNoLongerInCloudStorage) {
+		CompositeMultipartUploadStatus status = getCachedUploadOrRestart(user, request, forceRestart).orElseGet(() -> {
 			StorageLocationSetting storageLocation = storageLocationDao.get(request.getStorageLocationId());
-			
+
 			// Since the status for this file does not exist, create it.
 			CreateMultipartRequest createRequest;
-			
+
+			String requestMD5Hex = MultipartRequestUtils.calculateMD5AsHex(request);
```
marcomarasca: No need to compute it twice

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/file/MultipartManagerV2Impl.java
```diff
@@ -114,35 +114,24 @@ <T extends MultipartRequest> MultipartUploadStatus startOrResumeMultipartRequest
...
-
-		boolean fileHandleExistsFileNoLongerInCloudStorage = fileHandleExistsFileNoLongerInCloudStorage(user, status);
-
-		if (forceRestart || fileHandleExistsFileNoLongerInCloudStorage) {
-			// When forcing a restart or if the underlying object key of a file handle does not exist we clear the old hash changing the cache key, the previous multipart upload will be garbage collected by a worker
-			multipartUploadDAO.setUploadStatusHash(user.getId(), requestMD5Hex, RESTARTED_HASH_PREFIX + requestMD5Hex + "_" + UUID.randomUUID().toString());
-		}
-
-		if (status == null || forceRestart || fileHandleExistsFileNoLongerInCloudStorage) {
+		CompositeMultipartUploadStatus status = getCachedUploadOrRestart(user, request, forceRestart).orElseGet(() -> {
```
marcomarasca: if (forceRestart) -> You can set the "cached upload" to an empty optional without even checking for a cached version.

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/file/MultipartManagerV2ImplTest.java
```diff
@@ -454,6 +455,67 @@ public void testStartOrResumeMultipartRequestWithUnsupportedOperationException()
...
+
+		when(mockCompositeStatus.getMultipartUploadStatus()).thenReturn(mockStatus);
+		when(mockStatus.getState()).thenReturn(MultipartUploadState.COMPLETED);
+		when(mockCloudDaoProvider.getCloudServiceMultipartUploadDao(any())).thenReturn(mockCloudDao);
+		when(mockCloudDao.doesObjectExist(any(), any())).thenReturn(true);
+
+		// Call under test
+		Optional<CompositeMultipartUploadStatus> status = manager.getCachedUploadOrRestart(user, mockRequest, forceRestart);
+
+		assertEquals(mockCompositeStatus, status.get());
```
marcomarasca, line 488: you probably want to add some verify to make suer you are using the right parameters in your calls

lib/lib-googlecloud/src/test/java/org/sagebionetworks/googlecloud/SynapseGoogleCloudStorageClientImplUnitTest.java
```diff
@@ -183,7 +183,7 @@ public void testGetObjects() {
 	}
 
 	@Test
-	public void testDoesObjectExist() {
+	public void testDoesObjectExistExists() {
```
marcomarasca: withExisting?

marcomarasca (changes requested):

marcomarasca (approved):

PR #4752, opened 2022-10-07
PLFM-7486: Support description field in views

marcomarasca:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/persistence/DBORevision.java
```diff
@@ -41,16 +46,46 @@
...
+					Annotations entityProperties = AnnotationUtils.decompressedAnnotationsV1(backup.getEntityPropertyAnnotations());
+					
+					Object descriptionValue = entityProperties.getSingleValue("description");
+					
+					if (descriptionValue != null && descriptionValue instanceof String) {
+						backup.setDescription((String) descriptionValue);
+					}
+					
+				} catch (IOException e) {
+					// Can't do nothing about it
```
john-hill: I would at least log.

lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/persistence/DBORevisionTest.java
```diff
@@ -104,21 +108,115 @@ public void testRoundTrip() throws Exception {
...
+	public void testDescriptionMigrationWithExisting() throws IOException {
+		MigratableTableTranslation<DBORevision,DBORevision> translator = new DBORevision().getTranslator();
+		
+		DBORevision backup = new DBORevision();
+		backup.setDescription("existing");
+		
+		Annotations entityProperties = new Annotations();
+		
+		entityProperties.addAnnotation("description", "a description");
+
```
john-hill, line 219: maybe we should remove the old from the annotations?

lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/TableIndexDAOImpl.java
```diff
@@ -1407,15 +1417,16 @@ public void populateViewFromSnapshot(IdAndVersion idAndVersion, Iterator<String[
 			String[] row = input.next();
 			long rowSize = SQLUtils.calculateBytes(row);
 			if (batchSize + rowSize > maxBytesPerBatch) {
-				template.batchUpdate(sql, batch);
+				writeTransactionTemplate.executeWithoutResult(txStatus -> template.batchUpdate(sql, batch));
 				batch.clear();
+				batchSize = 0;
 			}
```
john-hill, line 1423: Good Catch!

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/table/TableIndexManagerImpl.java
```diff
@@ -82,7 +82,7 @@ public class TableIndexManagerImpl implements TableIndexManager {
 
 	public static final int MAX_MYSQL_INDEX_COUNT = 60; // mysql only supports a max of 64 secondary indices per table.
 
-	public static final long MAX_BYTES_PER_BATCH = 1024*1024*5;// 5MB
+	public static final long MAX_BYTES_PER_BATCH = 1024*1024*10;// 10MB
```
john-hill: This is more of a concern for what the worker machine memory can handle rather than what the DB can handle.  If we allow 20+ threads on one machine to build views at the same time then we might need 2 GB of the machines 8 GB.

john-hill (approved):

PR #4756, opened 2022-10-14
PLFM-7521: Configure async job and typed message workers

marcomarasca:


services/workers/src/main/java/org/sagebionetworks/worker/config/WorkersConfig.java
```diff
@@ -33,16 +64,25 @@ public class WorkersConfig {
...
+	
+	@Bean
+	public StackStatusGate stackStatusGate() {
+		return new StackStatusGate();
 	}
 
 	@Bean
-	public ConcurrentManager concurrentStackManager(CountingSemaphore countingSemaphore, StackStatusDao stackStatusDao) {
+	public ConcurrentManager concurrentStackManager(StackStatusDao stackStatusDao) {
 		return new ConcurrentManagerImpl(countingSemaphore, amazonSQSClient, stackStatusDao);
```
john-hill: Can we add @service tag to this object.  Then it could be in the constructor and one less parameter each trigger needs to import.

john-hill (approved):

PR #4757, opened 2022-10-14
Plfm 7522

john-hill:


lib/lib-table-query/src/test/java/org/sagebionetworks/table/query/TableQueryParserTest.java
```diff
@@ -474,5 +475,106 @@ public void testMultipleJoinWithOneOn() throws ParseException {
...
+	public void testQueryExpressionWithUnionAndGropuByLastQuery() throws ParseException {
+		QueryExpression element = new TableQueryParser("select * from syn111 union select * from syn222 group by foo ")
+				.queryExpression();
+		assertEquals("SELECT * FROM syn111 UNION SELECT * FROM syn222 GROUP BY foo", element.toSql());
+	}
+
+	@Test
+	public void testQueryExpressionWithUnionAndOrderNotLastQuery() throws ParseException {
+		String message = assertThrows(ParseException.class, () -> {
+			new TableQueryParser("select * from syn111 order by foo union select * from syn222").queryExpression();
```
marcomarasca, line 534: Note that this should work if the query is in parenthesis (also pagination), e.g. (Select * From asdasd order by foo limit 20) UNION select....

lib/lib-table-query/src/test/java/org/sagebionetworks/table/query/TableQueryParserTest.java
```diff
@@ -474,5 +475,106 @@ public void testMultipleJoinWithOneOn() throws ParseException {
...
+		String message = assertThrows(ParseException.class, () -> {
+			new TableQueryParser("select * from syn111 order by foo limit 10 offset 1 union select * from syn222")
+					.queryExpression();
+		}).getMessage();
+		assertTrue(message.contains("Was expecting"));
+		assertTrue(message.contains("<EOF>"));
+	}
+
+	@Test
+	public void testQueryExpressionWithUnionAndOrderByPaginationLastQuery() throws ParseException {
```
marcomarasca, line 574: let us make sure we didn't introduce a nested query possibility

lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/model/QueryExpression.java
```diff
@@ -0,0 +1,13 @@
+package org.sagebionetworks.table.query.model;
+
+/**
+ * QueryExpression ::= {@link NonJoinQueryExpression}
+ *
+ */
+public class QueryExpression extends SimpleBranch {
+
+	public QueryExpression(NonJoinQueryExpression nonJoinQueryExpression) {
+		super(nonJoinQueryExpression);
```
marcomarasca, line 10: See QuerySpecification, we need to invoke .recursiveSetParent()

marcomarasca (approved):

PR #4763, opened 2022-10-20
Plfm-7487

SandhraSokhal:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/NodeDAOImpl.java
```diff
@@ -1855,6 +1855,32 @@ public ObjectDataDTO mapRow(ResultSet rs, int rowNum)
...
 					annotations = AnnotationsV2Utils.emptyAnnotations();
 				}
+
+				org.sagebionetworks.repo.model.Annotations entityPropertyAnnotation = null;
+				byte[] bytes = rs.getBytes(COL_REVISION_ENTITY_PROPERTY_ANNOTATIONS_BLOB);
+				if (bytes != null) {
+					try {
+						entityPropertyAnnotation = AnnotationUtils.decompressedAnnotationsV1(bytes);
+					} catch (IOException e) {
+						throw new DatastoreException(e);
```
john-hill, line 1865: Since this is not called by users, it would be okay to throw the IO exception.

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/NodeDAOImpl.java
```diff
@@ -1855,6 +1855,32 @@ public ObjectDataDTO mapRow(ResultSet rs, int rowNum)
...
+						dto.setFileMD5(checksum.toString());
+					}
+
+					Object size = entityPropertyAnnotation.getSingleValue("size");
+					if (size != null) {
+						dto.setFileSizeBytes(Long.parseLong(size.toString()));
+					}
+					Object count = entityPropertyAnnotation.getSingleValue("count");
+					if (count != null) {
+						dto.setItemCount(Integer.parseInt(count.toString()));
```
john-hill: we can use a long since it is BIGINT in the DB and the rest of the columns are longs.

lib/models/src/main/java/org/sagebionetworks/repo/model/table/ObjectDataDTO.java
```diff
@@ -35,6 +35,8 @@ public class ObjectDataDTO implements Comparable<ObjectDataDTO> {
 	private Long fileSizeBytes;
 	private Boolean isInSynapseStorage;
 	private String fileMD5;
+	private  Integer itemCount;
```
john-hill: Long

john-hill (changes requested):
We need integration tests that ensure a DataSet without the new fields sill worker and views of both the data and datasetcollection still work (null or empty values for the missing data).

john-hill (approved):

PR #4769, opened 2022-11-09
Adding dataset default column to dataset collection

SandhraSokhal:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/table/metadata/providers/Constants.java
```diff
@@ -37,6 +37,8 @@ public class Constants {
 	};
 
 	static final ObjectField[] DATASET_DEFAULT_COLUMNS = new ObjectField[]{
```
john-hill: DATASET_DEFAULT_COLUMNS  -> DATASET_SPECIFIC_COLUMNS

Also rename the FILE_DEFAULT_COLUMNS -> FILE_SPECIFIC_COLUMNS
Remove from the files everything that is in default.

john-hill (approved):

services/workers/src/test/java/org/sagebionetworks/table/worker/DatasetCollectionIntegrationTest.java
```diff
@@ -95,6 +102,10 @@ public void before() throws Exception {
 		datasetMD5 = defaultColumnMapper.getColumnModels(ViewObjectType.DATASET_COLLECTION,ObjectField.datasetMD5Hex).get(0);
```
marcomarasca: Since now we use all the default columns this specific columns can be removed

marcomarasca (changes requested):
An empty dataset (with an empty set of items) should return 0 as count and size

marcomarasca (approved):

PR #4773, opened 2022-11-17
Plfm 6607

marcomarasca:


john-hill (commented):
We both agree this is super awkward to try to do with our current code.  This will create maintenance problems for this code (leaky abstraction).  We would rather take full advantage of the JSON schemas to define the Schema for a project.  Then use the JSON schema to both validate annotation data and auto-generate the ColumnModels (schema) of a view.

PR #4782, opened 2022-12-06
Plfm 7605

john-hill:


lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/QueryTranslator.java
```diff
@@ -125,48 +147,78 @@ public class QueryTranslator {
...
-
-
-			SQLTranslatorUtils.translateModel(transformedModel, parameters, userId, tableAndColumnMapper);
+			
+			parts.forEach(p -> {
+				SQLTranslatorUtils.addMetadataColumnsToSelect(p.getQuerySpecification().getSelectList(),
+						new HashSet<>(p.getMapper().getTableIds()), columnsToAddToSelect);
+
+				// translate each part
+				SQLTranslatorUtils.translateModel(p.getQuerySpecification(), parameters, userId, p.getMapper());
```
marcomarasca, line 183: (SELECT * FROM syn123 
UNION
SELECT * FROM syn345 ) 
WHERE foo = bar?

marcomarasca (approved):

PR #4825, opened 2023-03-23
create schema from select

john-hill:


lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/SQLTranslatorUtils.java
```diff
@@ -1042,6 +1044,60 @@ static void appendSingleQuotedValueToStringBuilder(StringBuilder builder, String
...
+	 * the maximum of all columns from the same column index.
+	 * 
+	 * @param selectSchemas The schemas of each select statement. Tables, Views, and
+	 *                      materialized views without UNIONs should only provide
+	 *                      one schema.
+	 * @return Will always return the first schema of the provided list, with
+	 *         ColumnModels modified as needed.
+	 */
+	public static List<ColumnModel> createSchemaOfSelect(final List<List<ColumnModel>> selectSchemas) {
+		if (selectSchemas.size() < 2) {
```
marcomarasca, line 1063: This will fail with no schema?

marcomarasca (approved):

PR #4839, opened 2023-05-02
PLFM-7722 send principal snapshot to kinesis stream

SandhraSokhal:


services/workers/src/main/java/org/sagebionetworks/snapshot/workers/writers/PrincipalObjectRecordWriter.java
```diff
@@ -28,31 +24,44 @@
...
+
 @Service
 public class PrincipalObjectRecordWriter implements ObjectRecordWriter {
-
 	static private Logger log = LogManager.getLogger(PrincipalObjectRecordWriter.class);
-	
+    private static final String TEAM_SNAPSHOT_STREAM = "teamSnapshots";
+    private static final String TEAM_MEMBER_SNAPSHOT_STREAM = "teamMemberSnapshots";
+    private static final String USER_PROFILE_SNAPSHOT_STREAM = "userProfileSnapshots";
+    private static final String USER_GROUP = "userGroup";
```
john-hill: userGroupSnapshot

john-hill (approved):

services/workers/src/main/java/org/sagebionetworks/snapshot/workers/writers/PrincipalObjectRecordWriter.java
```diff
@@ -67,6 +76,7 @@ public void buildAndWriteRecords(ProgressCallback progressCallback, List<ChangeM
 				userGroup = userGroupDAO.get(principalId);
 				ObjectRecord objectRecord = ObjectRecordBuilderUtils.buildObjectRecord(userGroup, message.getTimestamp().getTime());
 				objectRecordDAO.saveBatch(Arrays.asList(objectRecord), objectRecord.getJsonClassName());
+                kinesisUserGroups.add(new KinesisJsonEntityRecord<>(message.getTimestamp().getTime(), userGroup));
```
marcomarasca, line 79: I was looking at this part of the code and noticed that you use the timestamp of the message as a generic timestamp. Was this changed? Didn't we decide to send both the change timestamp and the current (snapshot) timestamp? I still see that we use the KinesisObjectSnapshotRecord.map in the code for the node and file snapshots that automatically sets the current timestamp for the snapshot timestamp and the change timstamp from the message.

PR #4847, opened 2023-05-16
log file event in json format to kinesis stream

SandhraSokhal:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/statistics/FileEvent.java
```diff
@@ -2,7 +2,10 @@
 
 public enum FileEvent {
 
-	FILE_DOWNLOAD("fileDownloads", "fileDownloadsRecords"), FILE_UPLOAD("fileUploads", "fileUploadsRecords");
+	FILE_DOWNLOAD("fileDownloads", "fileDownloadsRecords"),
+	FILE_UPLOAD("fileUploads", "fileUploadsRecords"),
+	FILE_DOWNLOAD_RECORD("fileDownloadRecords", "fileDownloadsRecords"),
```
marcomarasca: We do not need an additional type of event, since we are already processing these events.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/events/EventsCollectorImpl.java
```diff
@@ -91,11 +96,41 @@ private <E extends SynapseEvent> void log(List<E> events) {
 
 		recordsMap.forEach((streamName, records) -> {
 			if (!records.isEmpty()) {
+				logToJsonStream(streamName,records);
 				firehoseLogger.logBatch(streamName, records);
 			}
 		});
 	}
 
+	private void logToJsonStream(String streamName, List<AwsKinesisLogRecord> records) {
```
marcomarasca: This logic is tied to the file download/upload, it does not belong in this class.

marcomarasca (changes requested):

PR #4851, opened 2023-05-26
PLFM-7799: Keep materialized view available while re-building

marcomarasca:


lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/TableIndexDAOImpl.java
```diff
@@ -1614,4 +1619,62 @@ public void update(String sql, Map<String, Object> parameters) {
...
+		}
+		
+		sqlBuilder.deleteCharAt(sqlBuilder.length() - 1);
+	
+		template.update(sqlBuilder.toString());
+		
+		return randomPrefix;
+	}
+	
+	public int dropStaleTableIndices() {
```
john-hill: if we remain the to the temp index, drops will occur automatically. (we can even drop it after the exclusive lock is released).

lib/lib-table-cluster/src/test/java/org/sagebionetworks/table/cluster/SQLUtilsTest.java
```diff
@@ -915,9 +929,9 @@ public void testCreateAlterListColumnIndexTable(){
 		String sql = SQLUtils.createAlterListColumnIndexTable(tableId, oldColumn, newColumn, false);
 		String expected = "ALTER TABLE T999_INDEX_C21_" +
 				" DROP INDEX _C21__UNNEST_IDX," +
-				" DROP FOREIGN KEY T999_INDEX_C21__FK," +
```
john-hill, line 918: we do not need to drop and recreate the FK since we are using _ibfk_

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/table/MaterializedViewManagerImpl.java
```diff
@@ -151,54 +163,63 @@ public void deleteViewIndex(IdAndVersion idAndVersion) {
...
+			throw new IllegalArgumentException("MaterializedView snapshots not currently supported");
+		}
+		
+		if (TableState.AVAILABLE == tableManagerSupport.getTableStatusState(idAndVersion).orElse(null)) {
+			// We use a negative id as the temporary id of the view
+			IdAndVersion temporaryId = IdAndVersion.newBuilder()
+				.setId(-idAndVersion.getId())
+				.setVersion(idAndVersion.getVersion().orElse(null)).build();
+			
+			LockContext parentContext = new LockContext(ContextType.UpdatingMaterializedView, temporaryId);
```
john-hill: I would use the real ID as the user might see it and be confused.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/table/MaterializedViewManagerImpl.java
```diff
@@ -212,22 +233,84 @@ void createOrRebuildViewHoldingExclusiveLock(ProgressCallback callback, LockCont
 		}
 	}
 
+	void rebuildAvailableViewHoldingTemporaryExclusiveLock(ProgressCallback callback, LockContext parentContext, IdAndVersion idAndVersion, IdAndVersion temporaryId) throws Exception {
+		
+		boolean recoverOnLockUnavailable = true;
```
john-hill: no longer need this boolean

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/table/MaterializedViewManagerImplTest.java
```diff
@@ -788,7 +873,247 @@ public void testCreateOrRebuildViewHoldingExclusiveLockWithNoDefiningSql() throw
...
+		verify(mockTableManagerSupport).tryRunWithTableNonExclusiveLock(eq(mockProgressCallback), eq(expectedLockContext), any(), eq(dependentIdAndVersions[0]), eq(dependentIdAndVersions[1]));
+		verify(mockTableManagerSupport).isTableSearchEnabled(idAndVersion);
+		verify(managerSpy).createOrRebuildViewHoldingWriteLockAndAllDependentReadLocks(queryCaptor.capture(), eq(viewSchema), eq(false));
+		
+		QueryTranslator query = queryCaptor.getValue();
+		
+		assertEquals("select * from syn123 join syn456", query.getInputSql());
+		assertEquals(temporaryIndex, query.getIndexDescription());
+		
+		verifyNoMoreInteractions(mockTableManagerSupport);
```
john-hill, line 1011: no more on->indexmanger

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/table/MaterializedViewManagerImplTest.java
```diff
@@ -788,7 +873,247 @@ public void testCreateOrRebuildViewHoldingExclusiveLockWithNoDefiningSql() throw
...
+		
+		assertThrows(RecoverableMessageException.class, ()->{
+			// call under test
+			managerSpy.rebuildAvailableViewHoldingTemporaryExclusiveLock(mockProgressCallback, expectedLockContext, idAndVersion, temporaryId);
+		});
+
+		verify(mockTableManagerSupport).getIndexDescription(idAndVersion);
+		verify(mockMaterializedViewDao).getMaterializedViewDefiningSql(idAndVersion);
+		verify(mockTableManagerSupport).getTableStatusOrCreateIfNotExists(IdAndVersion.parse("syn123"));
+		verifyNoMoreInteractions(mockTableManagerSupport);
```
john-hill, line 1042: no more on index

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/table/MaterializedViewManagerImplTest.java
```diff
@@ -788,7 +873,247 @@ public void testCreateOrRebuildViewHoldingExclusiveLockWithNoDefiningSql() throw
...
+		
+	}
+	
+	@Test
+	public void testRebuildAvailableViewHoldingTemporaryExclusiveLockWithExclusiveLockNonAvailable() throws Exception {
```
john-hill: this should not happen

john-hill (changes requested):

lib/lib-table-cluster/src/test/java/org/sagebionetworks/table/cluster/SQLUtilsTest.java
marcomarasca: I tried getting rid of this, but MySQL spits out an error revolving around "in place" vs "copy" algorithm due to the automatic rename of the FK. I think we cannot mix the alter table with a column rename and the table rename at the same time without dropping the FK.

john-hill (approved):

john-hill (approved):

PR #4863, opened 2023-06-19
row level filtering on multiple query specs

john-hill:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/table/TableQueryManagerImpl.java
```diff
@@ -596,6 +600,14 @@ private QuerySpecification parserQuery(String sql) {
 			throw new IllegalArgumentException(e);
 		}
 	}
+	
+	private QueryExpression parserQueryQuerExpression(String sql) {
```
marcomarasca, line 604: -> parserQueryExpression

marcomarasca (approved):

PR #4871, opened 2023-06-22
PLFM-7896: Add support for Maps in translation process

leonli33:
Added support for translating maps in the OpenAPI translation process.

This resolves: https://sagebionetworks.jira.com/jira/software/c/projects/PLFM/boards/1?modal=detail&selectedIssue=PLFM-7896

### DIFF
--- a/lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/schema/JsonSchema.json
+++ b/lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/schema/JsonSchema.json
@@ -140,6 +140,10 @@
 		"contains": {
 			"description": "See: <a href=\"https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.9.3.1.4\">contains</a>",
 			"$recursiveRef": "#"
+		},
+		"additionalProperties": {
+			"description": "See: <a href=\"https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.9.3.2.3\">additionalProperties</a>",
+			"$recursiveRef": "#"
 		}
 	}
 }
--- a/lib/lib-openapi/src/main/java/org/sagebionetworks/translator/ObjectSchemaUtils.java
+++ b/lib/lib-openapi/src/main/java/org/sagebionetworks/translator/ObjectSchemaUtils.java
@@ -17,13 +17,15 @@ import org.sagebionetworks.schema.ObjectSchema;
 import org.sagebionetworks.schema.TYPE;
 import org.sagebionetworks.util.ValidateArgument;
 
-public class ObjectSchemaUtils {	
+public class ObjectSchemaUtils {
 	/**
-	 * Generates a mapping of class id to an ObjectSchema that represents that class.
-	 * Starts out with all of the concrete classes found in `autoGen`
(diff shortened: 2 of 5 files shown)


lib/lib-openapi/src/main/java/org/sagebionetworks/translator/ObjectSchemaUtils.java
```diff
@@ -104,16 +104,31 @@ public JsonSchema translateObjectSchemaPropertyToJsonSchema(ObjectSchema propert
...
+			switch (propertyType) {
+				case ARRAY:
+					schema.setType(Type.array);
+					JsonSchema items = translateObjectSchemaPropertyToJsonSchema(property.getItems(), property.getItems().getType());
+					schema.setItems(items);
+					break;
+				case TUPLE_ARRAY_MAP:
+				case MAP:
+					schema.setType(Type.object);
+					Map<String, ObjectSchema> properties = new LinkedHashMap<>();
```
john-hill: It seems the schema for maps should be:
{
	"type": "object",
	"description": "desc",
	"additionalProperties": {
		"type": "string"
	}
}

john-hill (changes requested):
It looks like maps can be handled with addtionalProperites.

lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/schema/JsonSchema.json
```diff
@@ -140,6 +140,10 @@
 		"contains": {
 			"description": "See: <a href=\"https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.9.3.1.4\">contains</a>",
 			"$recursiveRef": "#"
```
john-hill, line 142: I think we are using this correctly.  However, we need a way to validate that when a JSON schema includes "additionalProperties" the 3rd party validation library accepts a map of keys and values and validates it as expected. See:
JsonSchemaValidationManagerImplAutowireTest for examples
Also the following PR shows an example:
https://github.com/Sage-Bionetworks/Synapse-Repository-Services/commit/5b2f23a4c665f81232ecbf87c76110a92e13d3c5

lib/lib-openapi/src/main/java/org/sagebionetworks/translator/ObjectSchemaUtils.java
```diff
@@ -65,85 +69,189 @@ JsonSchema translateObjectSchemaToJsonSchema(ObjectSchema objectSchema) {
...
 	 */
-	Map<String, JsonSchema> translatePropertiesFromObjectSchema(Map<String, ObjectSchema> properties) {
+	Map<String, JsonSchema> translatePropertiesFromObjectSchema(Map<String, ObjectSchema> properties, String schemaId) {
 		ValidateArgument.required(properties, "properties");
 		Map<String, JsonSchema> result = new LinkedHashMap<>();
 		for (String propertyName : properties.keySet()) {
 			ObjectSchema property = properties.get(propertyName);
-			TYPE propertyType = property.getType();
-			result.put(propertyName, translateObjectSchemaPropertyToJsonSchema(property, propertyType));
+			if (isSelfReferencing(property)) {
```
john-hill, line 93: this if/else is repeated 3 times.  Can it be refactored?

john-hill (approved):

PR #4883, opened 2023-07-06
PLFM-7833: Add support for JSON_OBJECT and JSON_ARRAY

marcomarasca:


marcomarasca:
Will close this, this has similar changes to #4877. Needs to be rebased on top of develop once #4877 is merged.

PR #4896, opened 2023-07-20
PLFM-7932: Allows mysql functions and cast on the left-hand side of a predicate

marcomarasca:


lib/lib-table-query/src/main/resources/table-query-parser.jj
```diff
@@ -945,7 +945,18 @@ Predicate predicate():
...
 }
-ComparisonPredicate comparisonPredicate(ColumnReference leftHandSide):
+PredicateLeftHandSide predicateLeftHandSide():
+{
+ 	ColumnReference columnReference = null;
+ 	MySqlFunction mySqlFunction = null;
+ 	CastSpecification castSpecification = null;
+}
+{
+	columnReference = columnReference() { return new PredicateLeftHandSide(columnReference); }
```
john-hill, line 955: suggest columnRef to be last, since the other two have no ambiguity.

john-hill (approved):

PR #4905, opened 2023-08-04
Plfm-6447

SandhraSokhal:


lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/audit/AccessRecord.json
```diff
@@ -17,6 +17,11 @@
 			"type": "string",
 			"description": "The HTTP method"
 		},
+		"objectConcreteType": {
```
marcomarasca: This should refer to the request, e.g. requestConcreteType

marcomarasca (changes requested):
We want to capture the request type, not the response

lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/audit/AccessRecord.json
```diff
@@ -17,6 +17,11 @@
 			"type": "string",
 			"description": "The HTTP method"
 		},
+		"requestConcreteType": {
+			"type": "string",
+			"description": "The concrete type of the API request object, e.g., org.sagebionetworks.repo.model.FileEntity. Concrete type can be null either request body is not available ( in case of GET, DELETE API method) or object does not have field concrete type, e.g, UserGroup.java."
```
marcomarasca:
```suggestion
			"description": "The concrete type of the API request object (if any) e.g. org.sagebionetworks.repo.model.FileEntity. The requestConcreteType may be null if the request body is not available (i.e. GET or DELETE) or if does not have a concreteType field (e.g. there is no ambiguity in the type).

```

lib/lib-shared-models/src/main/java/org/sagebionetworks/repo/model/AccessRecordExtractorUtil.java
```diff
@@ -0,0 +1,54 @@
...
+        }
+    }
+
+    /**
+     * Exact the concrete type from an object
+     *
+     * @param object
+     * @return
+     */
+    public static String getConcreteType(Object object) {
```
marcomarasca: No duplicate code in production. Also this implements logic that is not needed (e.g. instance of Long)

lib/lib-shared-models/src/test/java/org/sagebionetworks/repo/model/AccessRecordExtractorUtilTest.java
```diff
@@ -0,0 +1,41 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+
+public class AccessRecordExtractorUtilTest {
+	
+	@Test
+	public void testEntity(){
+		FileEntity ob = new FileEntity();
+		ob.setId("syn123");
+		ob.setConcreteType("concreateType");
```
marcomarasca: you still have the typo

lib/lib-shared-models/src/test/java/org/sagebionetworks/repo/model/AccessRecordExtractorUtilTest.java
```diff
@@ -0,0 +1,41 @@
...
+		String id = AccessRecordExtractorUtil.getObjectId(ob);
+		String concreteType = AccessRecordExtractorUtil.getConcreteType(ob);
+		assertEquals(ob.getId(), id);
+		assertEquals(ob.getConcreteType(), concreteType);
```
marcomarasca: assertEquals("concreteType", concreteType)

lib/lib-shared-models/src/test/java/org/sagebionetworks/repo/model/AccessRecordExtractorUtilTest.java
```diff
@@ -0,0 +1,41 @@
...
+public class AccessRecordExtractorUtilTest {
+	
+	@Test
+	public void testEntity(){
+		FileEntity ob = new FileEntity();
+		ob.setId("syn123");
+		ob.setConcreteType("concreateType");
+		String id = AccessRecordExtractorUtil.getObjectId(ob);
+		String concreteType = AccessRecordExtractorUtil.getConcreteType(ob);
+		assertEquals(ob.getId(), id);
```
marcomarasca: assertEquals("syn123", id)

lib/lib-shared-models/src/test/java/org/sagebionetworks/repo/model/AccessRecordExtractorUtilTest.java
```diff
@@ -0,0 +1,41 @@
...
+		assertEquals(ob.getConcreteType(), concreteType);
+	}
+
+	@Test
+	public void testFileHandle(){
```
marcomarasca: testGetConcreteTypeWithImplicitConcreteType

marcomarasca: Separate tests for getObjectId and getConcreteType

lib/lib-shared-models/src/test/java/org/sagebionetworks/repo/model/AccessRecordExtractorUtilTest.java
```diff
@@ -0,0 +1,41 @@
...
+		S3FileHandle ob = new S3FileHandle();
+		ob.setId("syn123");
+		String id = AccessRecordExtractorUtil.getObjectId(ob);
+		String concreteType = AccessRecordExtractorUtil.getConcreteType(ob);
+		assertEquals(ob.getId(), id);
+		assertEquals("org.sagebionetworks.repo.model.file.S3FileHandle", concreteType);
+	}
+
+	@Test
+	public void testEntityWithOutConcreteType(){
```
marcomarasca: withOut -> without

services/repository/src/main/java/org/sagebionetworks/repo/web/AccessRecordListener.java
```diff
@@ -0,0 +1,23 @@
+package org.sagebionetworks.repo.web;
+
+
+/**
+ * Abstraction for a listener to response accessRecord object.
+ * @author John
+ *
+ */
+public interface AccessRecordListener {
```
marcomarasca: AccessRecordDataListener. The description should be something meaningful. E.g. Listener used to provide additional information to an access record

lib/lib-shared-models/src/main/java/org/sagebionetworks/repo/model/AccessRecordExtractorUtil.java
```diff
@@ -0,0 +1,54 @@
...
+ * @author John
+ */
+public class AccessRecordExtractorUtil {
+    /**
+     * Exact the ID from an object
+     *
+     * @param object
+     * @return
+     */
+    public static String getObjectId(Object object) {
```
marcomarasca: Return Optional

services/repository/src/main/java/org/sagebionetworks/repo/web/ControllerAuditAspect.java
```diff
@@ -13,21 +15,37 @@
...
 	}
-	
+
+	/**
+	 * Look at all request objects
+	 *
+	 * @param
+	 */
+
+	@Before("within(@org.springframework.stereotype.Controller *) && (args(.., @RequestBody requestBody))")
```
marcomarasca: I don't know how this can work, the "within" expression matches "any join point (method execution only in Spring AOP) within the package". You should probably use @within or @annotation instead.

marcomarasca: You do not need the joinPOint argument?

marcomarasca (changes requested):

lib/lib-shared-models/src/main/java/org/sagebionetworks/repo/model/AccessRecordExtractorUtil.java
```diff
@@ -0,0 +1,31 @@
...
+     * @param object
+     * @return
+     */
+    public static Optional<String> getObjectFieldValue(Object object, String fieldName) {
+        if (object == null) return Optional.empty();
+        try {
+            Field field = object.getClass().getDeclaredField(fieldName);
+            field.setAccessible(true);
+            Object ob = field.get(object);
+            if (ob instanceof String || ob instanceof Long) {
```
marcomarasca, line 24: Revert to previous code, no need toString on a string

services/repository/src/main/java/org/sagebionetworks/repo/web/ControllerAuditAspect.java
```diff
@@ -38,14 +42,17 @@ public void inspectResponseBody(Object responseBody) {
...
-	public void inspectRequestBody(JoinPoint joinPoint, Object requestBody) {
+	@Before("@within(org.springframework.stereotype.Controller) && " +
+			"execution(* * (..,@org.springframework.web.bind.annotation.RequestBody (*),..))" )
+	public void inspectRequestBody(JoinPoint joinPoint) {
 		// extract the concrete type from the request body
-		String concreteType = AccessRecordExtractorUtil.getConcreteType(requestBody);
-		if (concreteType != null) {
-			listener.setRequestConcreteType(concreteType);
+		for( Object o : joinPoint.getArgs()){
+			if(o instanceof JSONEntity){
```
marcomarasca: We should not care about JSONEntity. Additionally this code is prone to errors if there are multiple arguments that are instances of JSONEntity. The pointcut should be narrowed down to methods whose arguments are annotated with the org.springframework.web.bind.annotation.RequestBody annotation. I think you can use args to expose the annotated argument.

marcomarasca: We can even narrow it down more to methods that are annotated with @RequestMapping

marcomarasca: This seems to be complicated and unecessary, I would give a look at RequestBodyAdvice. If this approach does not work I would close the ticket as won't do as we are spending too much time on this (arguably not needed) task.

marcomarasca (changes requested):

lib/lib-shared-models/src/main/java/org/sagebionetworks/repo/model/AccessRecordExtractorUtil.java
SandhraSokhal: We can not return object without toString in any case because return type is Optional<String>. so we have to convert object to string.

SandhraSokhal:
Not needed

PR #4907, opened 2023-08-10
PLFM-7967

SandhraSokhal:


services/workers/src/main/java/org/sagebionetworks/file/worker/FileEventRecordWorker.java
```diff
@@ -37,7 +37,8 @@ public void run(ProgressCallback progressCallback, Message message, FileEvent ev
 
 
         FileEventRecord record = new FileEventRecord().setUserId(event.getUserId()).setFileHandleId(event.getFileHandleId()).
-                setAssociateId(event.getAssociateId()).setAssociateType(event.getAssociateType()).setProjectId(projectId);
+                setAssociateId(event.getAssociateId()).setAssociateType(event.getAssociateType()).setProjectId(projectId)
+                .setDownloadedFileHandleId(event.getDownloadedFileHandleId());
```
marcomarasca, line 41: Shouldn't this be set as the fileHandleId if the downloadedFileHandleId is not there?

marcomarasca (approved):

PR #4929, opened 2023-09-19
run facet queries in parallel

john-hill:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/config/RepositoryConfiguration.java
```diff
@@ -243,4 +245,9 @@ public WriteReadSemaphore getWriteReadSemaphore(StackConfiguration config, Count
 		return new WriteReadSemaphoreImpl(countingSemaphore, config.getWriteReadSemaphoreRunnerMaxReaders());
 	}
 	
+	@Bean
+	public ExecutorService cachedThreadPool() {
```
marcomarasca, line 249: I would probably call this tableQueriesThreadPool?

marcomarasca (approved):

PR #4939, opened 2023-09-26
Plfm 7784

john-hill:


integration-test/src/test/java/org/sagebionetworks/warehouse/WarehouseTestHelperImpl.java
```diff
@@ -0,0 +1,173 @@
...
+			StackConfiguration stackConfig) {
+		super();
+		this.s3Client = s3Client;
+		this.athenaClient = athenaClient;
+		this.clock = clock;
+		this.stackConfig = stackConfig;
+	}
+
+	@Override
+	public void assertWarehouseQuery(String queryString, int maxNumberOfHours) throws Exception {
```
marcomarasca, line 52: Since we assume it's a count query that returns a result bigger than 0, why don't we take in input just the table and the where condition?

integration-test/src/test/java/org/sagebionetworks/warehouse/WarehouseTestHelperImpl.java
```diff
@@ -0,0 +1,173 @@
...
+	
+	Executable executeQueryAndAssertResults(String key) {
+		return ()->{
+			String previousQueryString = s3Client.getObjectAsString(BUCKET_NAME, key);
+			int count = executeCountQuery(previousQueryString);
+			assertTrue(count > 0, "Expected count > 0 for query: " + previousQueryString);
+		};
+	}
+	
+	int executeCountQuery(String previousQueryString) throws Exception {
```
marcomarasca, line 88: What if you try to compact all the queries into one with a UNION ALL? E.g.
select count(*) from blablabla
UNION ALL
select count(*) from blablabla

integration-test/src/test/java/org/sagebionetworks/warehouse/WarehouseTestHelperImplTest.java
```diff
@@ -0,0 +1,346 @@
...
+		verify(warehouseHelper).saveQueryToS3(query, nowInstant, callersPath, maxNumberHours);
+		verify(warehouseHelper).executeQueryAndAssertResults(keyOne);
+		verify(warehouseHelper).executeQueryAndAssertResults(keyTwo);
+		verify(warehouseHelper, never()).executeQueryAndAssertResults(keyThree);
+		verify(warehouseHelper, times(2)).executeQueryAndAssertResults(any());
+
+		verify(mockS3Client).deleteObject(WarehouseTestHelperImpl.BUCKET_NAME, keyOne);
+		verify(mockS3Client).deleteObject(WarehouseTestHelperImpl.BUCKET_NAME, keyTwo);
+		verify(mockS3Client, never()).deleteObject(WarehouseTestHelperImpl.BUCKET_NAME, keyThree);
+		verify(mockS3Client, times(2)).deleteObject(any(String.class), any(String.class));
```
marcomarasca, line 121: You can use noMoreInterations

marcomarasca (approved):

PR #4940, opened 2023-10-11
PLFM-8085: Added AWS customer parameters to CloudFront urls for backwards compatibility with python client

harveyx22:


integration-test/src/test/java/org/sagebionetworks/IT520SynapseJavaClientEvaluationTest.java
```diff
@@ -1112,10 +1112,9 @@ public void testGetFileTemporaryUrlForSubmissionFileHandle() throws Exception {
 		String actual = synapse.getFileTemporaryUrlForSubmissionFileHandle(sub1.getId(), fileHandle.getId()).toString();
 		
 		// trim time-sensitive params from URL (PLFM-2019)
-		String cloudFrontUrlTimeSensitiveParameterStart = "&Expires";
```
marcomarasca, line 1115: I think this needs to be there, the issue with this test is that the returned URLs contain a variable timestamp that can vary between the two calls at lines 1111 and 1112. Both "Expires" and "X-Amz-Date" (as well as the signature) are "time sensitive". Therefore, they were trimming everything after the "variable" parameters. I think it would be ok to just keep the base URL without any parameter (this test is just checking that the URL of the submission file is the same as the entity file).

harveyx22: The way the URL is built, CloudFront specific parameters like "Expires" will always come after the custom added parameters like "X-Amz-Date". Because of this, this test now trims everything after the first time sensitive parameter, which is "X-Amz-Date".

marcomarasca (approved):

marcomarasca:
I approved the PR, but I think this should go into staging rather than develop (e.g. the base branch should be release-473).

PR #4942, opened 2023-10-12
Plfm 8065

john-hill:


lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/QueryTranslator.java
```diff
@@ -109,9 +111,10 @@ public class QueryTranslator implements TranslatedQuery {
 	 * @throws ParseException
 	 */
 	QueryTranslator(String startingSql, SchemaProvider schemaProvider, Long maxBytesPerPage,
-			Boolean includeEntityEtag, Long userId, IndexDescription indexDescription, SqlContext sqlContextIn) {
+			Boolean includeEntityEtag, Long userId, IndexDescription indexDescription, SqlContext sqlContextIn, Long changeNumber) {
```
marcomarasca, line 114: UNused parameter

lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/QueryTranslator.java
```diff
@@ -101,6 +101,8 @@ public class QueryTranslator implements TranslatedQuery {
 	private final List<ColumnModel> tableSchema;
 	
 	private boolean isCommonTableExpression;
+	
+	private final String tableHash;
```
marcomarasca, line 105: We can get that from the indexDescription

lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/description/TableIndexDescription.java
```diff
@@ -18,10 +19,18 @@
 public class TableIndexDescription implements IndexDescription {
 	
 	private final IdAndVersion idAndVersion;
+	private final Long lastTableVersionNumber;
 	
 	public TableIndexDescription(IdAndVersion idAndVersion) {
```
marcomarasca, line 24: Consider removing this constructor

lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/description/IndexDescription.java
```diff
@@ -11,56 +13,109 @@
...
+	 */
+	default void recursiveAppendIdAndChangeNumber(StringBuilder builder) {
+		getLastTableChangeNumber().ifPresent(n -> {
+			builder.append("+");
+			builder.append(getIdAndVersion().toString());
+			builder.append("-");
+			builder.append(n);
+		});
+		for (IndexDescription dependency : getDependencies()) {
+			dependency.recursiveAppendIdAndChangeNumber(builder);
```
marcomarasca, line 88: Can we check if there is loop detection when we build a view that depends on something else/

marcomarasca (approved):

PR #4943, opened 2023-10-13
added replication hash code and view hash code for delta checks

john-hill:


lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/TableIndexDAOImpl.java
```diff
@@ -918,6 +918,10 @@ public void setValues(PreparedStatement ps, int i)
 					ps.setNull(parameterIndex++, java.sql.Types.VARCHAR);
 					ps.setNull(parameterIndex + updateOffset, java.sql.Types.VARCHAR);
 				}
+				int hashCode = dto.hashCode();
```
marcomarasca, line 921: you might want to make sure that the annotations are in a stable order before hashing

marcomarasca (approved):

PR #4947, opened 2023-10-17
PLFM-8010: removed exception thrown for methods that return void that don't redirect

harveyx22:


lib/lib-openapi/src/main/java/org/sagebionetworks/translator/ControllerToControllerModelTranslator.java
```diff
@@ -249,8 +249,11 @@ ResponseModel getResponseModel(ExecutableElement method, List<? extends DocTree>
 	 * @param blockTags the blocktags of the method
 	 * @return the description
 	 */
-	String getResponseDescription(List<? extends DocTree> blockTags) {
+	String getResponseDescription(List<? extends DocTree> blockTags, ExecutableElement method) {
 		Optional<String> returnComment = getReturnComment(blockTags);
+		if (returnComment.isEmpty() && method.getReturnType().getKind().equals(TypeKind.VOID)) {
+			return "The request was successful, but there is no response content.";
```
john-hill, line 255: This will never return null.
For the void case return "Void" for the default case return "Auto-generated description..."
Can use Optional.getOrElse() instead of the java tertiary.

lib/lib-openapi/src/test/resources/controller/ComplexExampleController.java
```diff
@@ -69,4 +69,14 @@ public class ComplexExampleController {
...
+
+	/**
+	 * Example of an endpoint that returns void but is not redirected.
+	 *
+	 * @param name
+	 * 	the name of the pet
+	 */
+	@ResponseStatus(HttpStatus.OK)
+	@RequestMapping(value = "/complex-pet/cat/{name}", method = RequestMethod.DELETE)
+	public void deleteCat(@PathVariable String name) {}
```
john-hill, line 81: since missing description get a default, add a case where not void but has not a description.

lib/lib-openapi/src/test/java/org/sagebionetworks/translator/ControllerModelDocletTest.java
```diff
@@ -247,4 +249,17 @@ public void testComplexTypesInPropertiesAreGeneratedAsReferences() {
...
+		JSONObject pathsObj = generatedOpenAPISpec.getJSONObject("paths");
+		JSONObject pathObj = pathsObj.getJSONObject("/repo/v1/complex-pet/cat/{name}");
+		JSONObject deleteObj = pathObj.getJSONObject("delete");
+		JSONObject responsesObj = deleteObj.getJSONObject("responses");
+		JSONObject okObj = responsesObj.getJSONObject("200");
+
+		// Should be a reference to the place the complex type lives in the "components" section
+		assertEquals("The request was successful, but there is no response content.", okObj.getString("description"));
+		assertThrows(JSONException.class, () -> okObj.getString("content"));
+	}
```
john-hill, line 264: add a test for the non-void missing description.

lib/lib-openapi/src/test/java/org/sagebionetworks/translator/ControllerModelDocletTest.java
```diff
@@ -247,4 +249,17 @@ public void testComplexTypesInPropertiesAreGeneratedAsReferences() {
...
+	public void testDescriptionandContentForVoidReturnMethodWithNoRedirect() {
+		JSONObject pathsObj = generatedOpenAPISpec.getJSONObject("paths");
+		JSONObject pathObj = pathsObj.getJSONObject("/repo/v1/complex-pet/cat/{name}");
+		JSONObject deleteObj = pathObj.getJSONObject("delete");
+		JSONObject responsesObj = deleteObj.getJSONObject("responses");
+		JSONObject okObj = responsesObj.getJSONObject("200");
+
+		// Should be a reference to the place the complex type lives in the "components" section
+		assertEquals("The request was successful, but there is no response content.", okObj.getString("description"));
+		assertThrows(JSONException.class, () -> okObj.getString("content"));
```
john-hill, line 263: okObject.isNull("content") instead of checking for an exception.

john-hill (approved):

PR #4951, opened 2023-10-24
PLFM-8011 and PLFM-8015: Added support for additional HTTPStatus codes and skipped methods not required for OpenAPI translation

harveyx22:


lib/lib-openapi/src/test/resources/controller/ComplexExampleController.java
```diff
@@ -144,4 +145,58 @@ public LoginResponse createNewAccount(
...
+	public @ResponseBody String getNoResponseStatus(@PathVariable String name) {}
+
+	/**
+	 * Example of a private method included in the controller
+	 *
+	 * @param wikiId
+	 * @param wikiPage
+	 */
+	private static void validateUpateArguments(String wikiId, WikiPage wikiPage) {
+	}
```
john-hill, line 194: need a static and private separate.

john-hill (approved):

PR #4961, opened 2023-11-03
PLFM-8126: Temp rollback for release-476

xschildw:


services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/file/FileHandleManagerImplTest.java
```diff
@@ -622,7 +624,6 @@ public void testGetRedirectURLForFileHandleS3() throws DatastoreException, NotFo
 		
 		when(mockFileHandleDao.get(s3FileHandle.getId())).thenReturn(s3FileHandle);
 		String expectedURL = "https://amamzon.com";
-		when(mockStackConfig.getS3Bucket()).thenReturn("devdata.sagebase.org");
```
marcomarasca: INstead of removing this, why don't we leave the if as is and just make the same call? This makes it so we won't have to track back this changes later.

marcomarasca (changes requested):

marcomarasca (approved):

PR #4966, opened 2023-11-09
PLFM-8007: Updated OpenAPI translator to reference JSON basic types instead of java.lang classes

harveyx22:


lib/lib-openapi/src/main/java/org/sagebionetworks/translator/ControllerModelsToOpenAPIModelTranslator.java
```diff
@@ -166,7 +170,13 @@ EndpointInfo getEndpointInfo(MethodModel method, String displayName, String full
 	JsonSchema getReferenceSchema(String id) {
 		ValidateArgument.required(id, "id");
 		JsonSchema schema = new JsonSchema();
-		schema.set$ref("#/components/schemas/" + id);
+
+		Optional<Type> schemaType = getJsonSchemaBasicTypeForClass(id);
```
john-hill, line 174: consider optional.ifPresentOrElse

john-hill (approved):

PR #4978, opened 2023-11-30
PLFM-7273: Add effective ARs to node snapshots

marcomarasca:


john-hill (changes requested):
It would be nice to have an integration test such as:
select * from nodesnapshots where effective_ars is not null and cardinality(effective_ars) > 0  limit 10;

john-hill (approved):

PR #4991, opened 2023-12-13
PLFM-8181: Added security requirements to OpenAPI description

harveyx22:


lib/lib-openapi/src/main/java/org/sagebionetworks/translator/ControllerToControllerModelTranslator.java
```diff
@@ -246,6 +249,12 @@ List<MethodModel> getMethods(List<? extends Element> enclosedElements, DocTrees
 				throw new IllegalStateException("Method " + methodName + " missing RequestMapping annotation.");
 			}
 
+			RequiredScopeModel requiredScope = null;
+			if (annotationToModel.containsKey(RequiredScope.class)) {
+				System.out.println(annotationToModel);
```
john-hill: this will be based on userId

lib/lib-openapi/src/test/java/org/sagebionetworks/translator/ControllerModelDocletTest.java
```diff
@@ -868,4 +868,22 @@ public void testUserId() {
...
+	public void testAuthorization() {
+		JSONObject pathsObj = generatedOpenAPISpec.getJSONObject("paths");
+		JSONObject pathObj = pathsObj.getJSONObject("/repo/v1/complex-pet/authorization");
+		JSONObject getObj = pathObj.getJSONObject("get");
+		JSONArray securityObj = getObj.getJSONArray("security");
+		JSONObject bearerObj = securityObj.getJSONObject(0);
+
+		assertEquals(new JSONArray().toString(), bearerObj.getJSONArray("bearerAuth").toString());
+
+		JSONObject componentsObj = generatedOpenAPISpec.getJSONObject("components");
```
john-hill, line 821: this is based on the case where the userId parameter is applied to the method.

lib/lib-openapi/src/test/resources/controller/ComplexExampleController.java
```diff
@@ -473,4 +475,16 @@ public void getUserId(
...
+	 * Example of an endpoint that requires authorization
+	 *
+	 * @param userId
+	 */
+	@RequiredScope({OAuthScope.view, OAuthScope.modify})
+	@RequestMapping(value = "/complex-pet/authorization", method = RequestMethod.GET)
+	public void getAuthorization(
+			@RequestParam(value = "userId") Long userId
+	) {
+	}
```
john-hill, line 491: we need two cases: one that includes the userId (requires auth), and one missing userId, no auth needed.

john-hill (changes requested):

lib/lib-openapi/src/main/java/org/sagebionetworks/openapi/datamodel/pathinfo/EndpointInfo.java
```diff
@@ -180,4 +181,28 @@ void populateTags(JSONArrayAdapter tags) throws JSONObjectAdapterException {
 			tags.put(i, this.tags.get(i));
 		}
 	}
+
+	boolean securityRequirementsAreEqual(Map<String, String[]> map1, Map<String, String[]> map2) {
```
john-hill, line 185: implemented equals on two maps?

john-hill (approved):

PR #5011, opened 2024-02-01
PLFM-8218: Expose 2fa info in user profile

marcomarasca:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/UserProfileManagerImpl.java
```diff
@@ -144,10 +150,24 @@ private void addAliasesToProfiles(List<UserProfile> userProfiles) {
 			insertAliasIntoProfile(profile, alias);
 		}
 	}
+	
+	private void addTwoFactorAuthInfo(List<UserProfile> userProfiles) {
+		Map<Long, Boolean> twoFactorStateMap = authDao.getTwoFactorAuthStateMap(userProfiles.stream()
```
SandhraSokhal, line 159: We should check if userProfile list is empty or not. Dao will throw exception on empty list

SandhraSokhal (approved):
Looks good to me

PR #5022, opened 2024-02-16
PLFM-7546: resolve issue with column maxSize not accounting for conca…

lukemoenning:
…t in definingSQL

### DIFF
--- a/lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/SQLTranslatorUtils.java
+++ b/lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/SQLTranslatorUtils.java
@@ -1366,49 +1366,65 @@ public class SQLTranslatorUtils {
 	 * @return
 	 */
 	public static ColumnModel getSchemaOfDerivedColumn(DerivedColumn derivedColumn,
-			TableAndColumnMapper tableAndColumnMapper) {
-		// the SelectColumn provides a starting name and type.
+			TableAndColumnMapper tableAndColumnMapper) 
+	{
+		// The SelectColumn provides a starting name and type.
 		SelectColumn selectColumn = getSelectColumns(derivedColumn, tableAndColumnMapper);
-		Long maximumSize = null;
-		Long maxListLength = null;
-		List<String> enumValues = null;
 		// The data type is correctly inferred by the #getSelectColumns call
 		ColumnType columnType = selectColumn.getColumnType();
-		String defaultValue = null;
-		FacetType facetType = null;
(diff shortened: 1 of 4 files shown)


lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/SQLTranslatorUtils.java
```diff
@@ -1366,49 +1366,65 @@ public static List<ColumnModel> getSchemaOfSelect(SelectList selectList,
 	 * @return
 	 */
 	public static ColumnModel getSchemaOfDerivedColumn(DerivedColumn derivedColumn,
-			TableAndColumnMapper tableAndColumnMapper) {
-		// the SelectColumn provides a starting name and type.
+			TableAndColumnMapper tableAndColumnMapper) 
+	{
```
john-hill, line 1370: This '{' must be on the same line as the method signature.

lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/SQLTranslatorUtils.java
```diff
@@ -1366,49 +1366,65 @@ public static List<ColumnModel> getSchemaOfSelect(SelectList selectList,
...
+			attrs.maxListLength = cm.getMaximumListLength();
+			attrs.defaultValue = cm.getDefaultValue();
+			attrs.facetType = cm.getFacetType();
+			attrs.enumValues = cm.getEnumValues();
+			attrs.jsonSubColumns = cm.getJsonSubColumns();
+		} else {
+			derivedColumn.createIterable(ValueExpressionPrimary.class).forEach(vep -> {
+				switch(vep.getChild().getClass().getName()) {
+				
+					case "org.sagebionetworks.table.query.model.UnsignedValueSpecification":
```
john-hill, line 1400: In the version of Java we are using the switch only works with a few basic types.  It looks like they will be improving this in Java 17+.  Until then we need to work with what we have.  The problem with matching on the string of a class name, is that this will break if the class is moved or renamed.
Also a switch with only one case plus a default is identical to a simple if/else statement.
However, in this case you should have three possible matches on the child:
1. UnsignedValueSpecification
2. ColumnReference
3. Any other case
For the second case you do not need to call vep.getFirstElementOfType() because the child already be a ColumnReference.
Here is a structure I would accept:
if(vep.getChild() instanceof UnsignedValueSpecification){
...
} else if (vep.getChild() instanceof ColumnReference){
    tableAndColumnMapper.lookupColumnReference((ColumnReference)vep.getChild())...
}

john-hill (changes requested):

PR #5031, opened 2024-03-01
PLFM-8299: Deterministic query search conditions order

marcomarasca:


lib/lib-table-cluster/src/test/java/org/sagebionetworks/table/cluster/QueryTranslatorTest.java
```diff
@@ -2451,5 +2452,144 @@ public void tesVirtualTableWithDouble() {
...
+		query = QueryTranslator.builder("select * from syn1 where bar HAS (32, 42) AND foo HAS_LIKE ('a', 'c', 'b')", userId)
+				.schemaProvider(mockSchemaProvider)
+				.sqlContext(SqlContext.query)
+				.indexDescription(indexDescription)
+				.build();
+		
+		assertEquals(expectedSql, query.getOutputSQL());
+		// We want to make sure that the order is deterministic, so we compare the toString
+		assertEquals(expectedParams.toString(), query.getParameters().toString());
+	}
```
john-hill, line 2593: Can we have tests that confirm when search conditions, boolean term and in clause with duplicates are removed.  So we can state that it is intentional an "safe".

lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/model/SQLElement.java
```diff
@@ -257,6 +259,30 @@ public <T extends Element> Optional<T> getContext(Class<T> type){
 		}
 		return this.parent.getContext(type);
 	}
+
+	@Override
```
john-hill, line 263: comment as to why we use toSql for all three.

john-hill (approved):

PR #5038, opened 2024-03-06
PLFM-8258: Add discriminator to open api specs

marcomarasca:


lib/lib-openapi/src/main/java/org/sagebionetworks/openapi/model/OpenApiJsonSchema.java
```diff
@@ -0,0 +1,43 @@
...
+	}
+	
+	public Discriminator getDiscriminator() {
+		return discriminator;
+	}
+	
+	public OpenApiJsonSchema setDiscriminator(Discriminator discriminator) {
+		this.discriminator = discriminator;
+		return this;
+	}
```
SandhraSokhal, line 42: we need to add equals and hash code method

SandhraSokhal (approved):
Looks good to me

PR #5053, opened 2024-04-02
PLFM-7692: add check for name length in NameValidation.validateName

lukemoenning:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/jdo/NameValidation.java
```diff
@@ -34,6 +36,9 @@ public static String validateName(String key) {
 		if ("".equals(key)) {
 			throw new IllegalArgumentException("Name cannot be only whitespace or empty string");
 		}
+		if (key.length() > TableConstants.MAX_COLUMN_NAME_SIZE_CHARS) {
```
john-hill: This is the wrong constant.  Add the constant to this class.  We will need tests for the other two cases.

lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/jdo/NameValidationTest.java
```diff
@@ -46,6 +47,15 @@ public void testInvalidNamesWhiteEmpty() {
 		}).getMessage();
 		assertEquals("Name cannot be only whitespace or empty string", message);
 	}
+	
+	@Test
+	public void testInvalidNamesTooLong() {
+		String message = assertThrows(IllegalArgumentException.class, () -> {
+			// These are all bad names
+			NameValidation.validateName("0".repeat(TableConstants.MAX_COLUMN_NAME_SIZE_CHARS + 1));
```
john-hill: change to use the other constant.

lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/jdo/NameValidationTest.java
```diff
@@ -46,6 +47,15 @@ public void testInvalidNamesWhiteEmpty() {
...
+		}).getMessage();
+		assertEquals(NameValidation.NAME_LENGTH_TOO_LONG, message);
+	}
```
john-hill, line 58: add one more unit test for the case where the name is exactly the max size.

john-hill (changes requested):

john-hill (approved):

PR #5061, opened 2024-04-12
PLFM- 8304

SandhraSokhal:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/TeamUtils.java
```diff
@@ -56,9 +56,32 @@ public static Team copyDboToDto(DBOTeam dbo) throws DatastoreException {
 		} else {
 			dto.setIcon(String.valueOf(dbo.getIcon()));
 		}
+		setMembershipStatus(dbo.getState(), dto);
 		return dto;
 	}
 
+	public static void setMembershipStatus(String state, Team team) throws DatastoreException{
+		if(team.getCanPublicJoin() == null || team.getCanRequestMembership() == null){
```
john-hill: we can move this logic to the enum.  We also unconditionally set the two Booleans from the state enum.
We also need to change all places in the code to use the state enum and no longer use the booleans.

john-hill (changes requested):

john-hill (approved):

PR #5062, opened 2024-04-12
PLFM-8286: Add selectFileVersionColumn for adding files to download list

marcomarasca:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/download/DownloadListManagerImpl.java
```diff
@@ -460,16 +466,21 @@ AddToDownloadListResponse addQueryResultsToDownloadList(final ProgressCallback p
 	 */
 	DownloadListItem createDownloadsListItemFromRow(final boolean useVersion, Row row) {
```
john-hill: useVersion->inlucdeVersion
can we unconditionally use the values for both rowId and rowVersion?

john-hill (changes requested):

john-hill (approved):

PR #5076, opened 2024-05-01
PLFM-8365 and PLFM-8346: User certification revocation

marcomarasca:


integration-test/src/test/java/org/sagebionetworks/IT500SynapseJavaClient.java
```diff
@@ -838,7 +840,7 @@ public void testCertifiedUserQuiz() throws Exception {
 		assertEquals(1, qrs.getResults().size());
 		assertEquals(pr.getResponseId(), qrs.getResults().iterator().next().getId());
 
-		PaginatedResults<PassingRecord> prs = adminSynapse.getCertifiedUserPassingRecords(0L, 2L, myId);
+		PaginatedResults<PassingRecord> prs = synapse.getCertifiedUserPassingRecords(0L, 2L, myId);
```
SandhraSokhal, line 844: we can use the utility to create a certified user , even for failing records and test user is removed from certified group.

SandhraSokhal (changes requested):

SandhraSokhal (approved):
LGTM

PR #5091, opened 2024-05-24
use new CsvWriter that throws IOExceptions

john-hill:


lib/models/src/main/java/org/sagebionetworks/repo/model/dao/table/RowHandler.java
```diff
@@ -13,6 +15,7 @@ public interface RowHandler {
 	/**
 	 * Called for each row of the set.
 	 * @param row
+	 * @throws IOException 
 	 */
-	public void nextRow(Row row);
+	public void nextRow(Row row) throws IOException;
```
marcomarasca, line 20: I do not think we should throw an IOException here.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/report/StorageReportManager.java
```diff
@@ -23,6 +25,7 @@ public interface StorageReportManager {
 	 * @return
 	 * @throws NotFoundException
 	 * @throws LockUnavilableException
+	 * @throws IOException 
 	 */
-	void writeStorageReport(UserInfo user, DownloadStorageReportRequest request, CSVWriterStream writer) throws NotFoundException, LockUnavilableException;
+	void writeStorageReport(UserInfo user, DownloadStorageReportRequest request, CSVWriterStream writer) throws NotFoundException, LockUnavilableException, IOException;
```
marcomarasca, line 30: Probably not needed

services/workers/src/main/java/org/sagebionetworks/table/worker/TableCSVDownloadWorker.java
```diff
@@ -119,6 +110,10 @@ public DownloadFromTableResult run(String jobId, UserInfo user, DownloadFromTabl
 			jobProgressCallback.updateProgress("Waiting for the table index to become available...", 0L, 100L);
 			// Throwing this will put the message back on the queue in 5 seconds.
 			throw new RecoverableMessageException();
+		} catch (IOException e){
+			log.error("Worker Failed. Will retry", e);
```
marcomarasca, line 114: We might not want to actually retry on IOException, e.g. what if it is unrecoverable?

marcomarasca (changes requested):

PR #5099, opened 2024-05-30
Plfm 8233.2

john-hill:


lib/models/src/main/java/org/sagebionetworks/repo/model/table/ObjectField.java
```diff
@@ -40,6 +40,7 @@ public enum ObjectField {
 	projectId				(OBJECT_REPLICATION_COL_PROJECT_ID, 		ColumnType.ENTITYID,		null,	FacetType.enumeration),
 	modifiedOn				(OBJECT_REPLICATION_COL_MODIFIED_ON,		ColumnType.DATE,			null,	FacetType.range),
 	modifiedBy				(OBJECT_REPLICATION_COL_MODIFIED_BY,		ColumnType.USERID,			null,	FacetType.enumeration),
+	_path					(OBJECT_REPLICATION_COL_PATH,				ColumnType.LARGETEXT,		null,	null),
```
marcomarasca: 1. Should we really use a LARGETEXT?
2. _path is a poor name, we should just use path if possible (we need to check if there is a matching column model).

marcomarasca (changes requested):

marcomarasca (approved):
Since now "path" is a STRING with a max size of 1000, will that interfere with any potential eisting ColumnModel?

PR #5114, opened 2024-07-12
PLFM-8508 - Do not overwrite the AccessRequirement name field if it is present

nickgros:


lib/jdomodels/src/test/java/org/sagebionetworks/repo/model/dbo/dao/AccessRequirementUtilsTest.java
```diff
@@ -74,7 +74,53 @@ public void testRoundtrip() throws Exception {
...
+		dto.setName("someName");
+		dto.setDescription("someDescription");
+
+		DBOAccessRequirement dboRequirement = new DBOAccessRequirement();
+		DBOAccessRequirementRevision dboRevision = new DBOAccessRequirementRevision();
+		AccessRequirementUtils.copyDtoToDbo(dto, dboRequirement, dboRevision);
+		assertEquals(dto.getName(), dboRequirement.getName());
+		AccessRequirement dto2 = AccessRequirementUtils.copyDboToDto(dboRequirement, dboRevision);
+		assertEquals(dto, dto2);
+	}
```
nickgros, line 121: This test verifies that the bug is fixed

john-hill (approved):

PR #5125, opened 2024-07-30
PLFM-8519 - Add `userHasDownloadPermission` to `RestrictionInformationResponse`

nickgros:


john-hill (changes requested):
Can we explore refactoring the decider logic into a static util that can be used for both the mangers?

marcomarasca (approved):
LGTM!

PR #5133, opened 2024-08-28
Plfm 8153

john-hill:


lib/lib-javadoc/pom.xml
```diff
@@ -38,6 +39,10 @@
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-web</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework</groupId>
```
marcomarasca: maybe test scope?

lib/lib-javadoc/src/main/java/org/sagebionetworks/javadoc/web/services/FilterUtils.java
```diff
@@ -1,93 +1,80 @@
...
-                    	list.add(classDoc);
-                    }
-            	}
-            }
-		}
-		return list.iterator();
+	public static Iterator<TypeElement> controllerIterator(DocletEnvironment docletEnvironment) {
+		Set<TypeElement> types = ElementFilter.typesIn(docletEnvironment.getIncludedElements());
+		types.forEach(t -> {
+			System.out.println(t.toString());
```
marcomarasca, line 36: casn get rid of debugging statmt

marcomarasca (approved):

PR #5142, opened 2024-09-12
PLFM-8606 - Fix links to reapply for profile validation

nickgros:
Identity verification approval link goes to the profile page, so the text should say so (not "your home page")

Identity verification rejection links should go to the account settings page, where users can re-apply.


marcomarasca (approved):
Lgtm

PR #5186, opened 2024-11-01
switch to a streaming JSON read

john-hill:


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/migration/MigrationTypeProviderImpl.java
```diff
@@ -87,9 +84,10 @@ public <B> Optional<List<B>> readObjects(Class<? extends B> clazz, BackupAliasTy
 	}
 
 	<B> Optional<List<B>> readJSON(Class<? extends B> clazz, BackupAliasType backupAliasType, InputStream input) {
+
 		try {
 			return Optional.of(
-					JavaJSONUtil.readFromJSON(clazz, new JSONArray(IOUtils.toString(new InputStreamReader(input)))));
+					JavaJSONUtil.readFromJSON(clazz, new InputStreamReader(input, "UTF-8")));
```
marcomarasca: Consider wrapping the InputStreamReader in a BufferedReader for efficiency.

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/migration/MigrationTypeProviderImpl.java
```diff
@@ -87,9 +84,10 @@ public <B> Optional<List<B>> readObjects(Class<? extends B> clazz, BackupAliasTy
 	}
 
 	<B> Optional<List<B>> readJSON(Class<? extends B> clazz, BackupAliasType backupAliasType, InputStream input) {
+
 		try {
 			return Optional.of(
-					JavaJSONUtil.readFromJSON(clazz, new JSONArray(IOUtils.toString(new InputStreamReader(input)))));
```
marcomarasca, line 92: Would the old readFromJSON be still in use? Can we remove it?

lib/lib-utils/pom.xml
```diff
@@ -47,7 +47,10 @@
 			<artifactId>mockito-junit-jupiter</artifactId>
 			<scope>test</scope>
 		</dependency>
-
+		<dependency>
```
marcomarasca, line 50: I do not think we need to use gson, the JSON-java should already be sufficient to read from the stream.

lib/lib-utils/src/main/java/org/sagebionetworks/util/json/JavaJSONUtil.java
```diff
@@ -84,6 +94,84 @@ public static <T> List<T> readFromJSON(Class<? extends T> clazz, JSONArray array
...
+	 * @return
+	 */
+	public static <T> List<T> readFromJSON(Class<? extends T> clazz, Reader reader) {
+		ValidateArgument.required(reader, "reader");
+		ValidateArgument.required(clazz, "clazz");
+		Map<String, Object> typeMap = Arrays.stream(clazz.getDeclaredFields())
+				.collect(Collectors.toMap(f -> f.getName(), f -> f.getType()));
+		try {
+			List<T> list = new ArrayList<>();
+			try (JsonReader stream = new JsonReader(reader)) {
```
marcomarasca: Rather than using the gson library, we can use the JSONTokener(Reader) to read the stream. See JSONTokenener.nextClean() (e.g. starts with '[') and JSONTOkener.nextValue() (to read the next JSONObject).

lib/lib-utils/src/main/java/org/sagebionetworks/util/json/JavaJSONUtil.java
```diff
@@ -84,6 +94,84 @@ public static <T> List<T> readFromJSON(Class<? extends T> clazz, JSONArray array
...
+	}
+
+	/**
+	 * Read a single JSONObject from the provided stream.
+	 * 
+	 * @param stream
+	 * @return
+	 * @throws IOException
+	 */
+	public static JSONObject readObject(Map<String, Object> typeMap, JsonReader stream) throws IOException {
```
marcomarasca: I think this would not work if there were nested objects, I would simply use the JSONTokener (See previous comment)

lib/lib-utils/src/main/java/org/sagebionetworks/util/json/JavaJSONUtil.java
```diff
@@ -84,6 +94,84 @@ public static <T> List<T> readFromJSON(Class<? extends T> clazz, JSONArray array
...
 	}
 
+	/**
+	 * Read a list of objects directly from the provided Stream
+	 * @param <T>
+	 * @param clazz
+	 * @param reader
+	 * @return
+	 */
+	public static <T> List<T> readFromJSON(Class<? extends T> clazz, Reader reader) {
```
marcomarasca: I would rename this to something like readFromJsonArrayStream

marcomarasca (changes requested):

marcomarasca (approved):

PR #5197, opened 2024-11-07
PLFM-8688: Project storage limits enforcement

marcomarasca:


services/repository-managers/src/main/java/org/sagebionetworks/repo/service/metadata/FileEntityMetadataProvider.java
```diff
@@ -42,11 +54,42 @@ public void validateEntity(FileEntity entity, EntityEvent event)
...
 			// Note: Parent ID is validated as non-null by AllTypesValidatorImpl.
 			stsManager.validateCanAddFile(event.getUserInfo(), entity.getDataFileHandleId(), entity.getParentId());
 		}
 	}
+	
+	void validateProjectStorageLocationUsageLimit(FileEntity entity, EntityEvent event) {
+		Long newFileHandleId = KeyFactory.stringToKey(entity.getDataFileHandleId());
+		Long existingFileHandleId = null;
+		
+		switch (event.getType()) {
```
john-hill, line 69: if we unconditionally check the limit on all updates then the only weird thing would be that you could not rename a file when over the limit.

john-hill (approved):
Good job!  Almost done!

PR #5217, opened 2024-12-10
 PLFM-8627

SandhraSokhal:


marcomarasca (changes requested):
The model providing the internal function names is not valuable information, I think the prompt should be modified to be more specific (See comments on https://github.com/Sage-Bionetworks/Synapse-Stack-Builder/pull/676).

PR #5225, opened 2025-01-30
Plfm 8718

john-hill:


services/workers/src/main/java/org/sagebionetworks/change/workers/RangeIterator.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.List;
+import java.util.Objects;
+
+import org.sagebionetworks.util.ValidateArgument;
+
+/**
+ * Provides an iterator over a range of lower and upper bounds given a page size
+ * and direction.
+ */
+public class RangeIterator implements Iterable<RangeIterator.Range> {
```
marcomarasca, line 15: maybe let us move this under lib-utils or ranme it (e.g. ChangesRangeIterator)

marcomarasca (approved):

PR #5229, opened 2025-03-08
PLFM-8522: Set displayName to null on get().

xschildw:


xschildw:
Passed tests on personal build.

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/UserProfileUtils.java
```diff
@@ -91,6 +91,7 @@ public static UserProfile convertDboToDto(DBOUserProfile dbo) throws DatastoreEx
 		if(dbo.getCreatedOn() != null) {
 			dto.setCreatedOn(new Date(dbo.getCreatedOn()));
 		}
+		dto.setDisplayName(null);
```
john-hill, line 94: Just because this field is deprecated does not mean an older version of a client does not use it.  Setting this to null could be a breaking API change.   I would close this issue as "will not do" explaining that it is a breaking change.

john-hill (changes requested):

PR #5231, opened 2025-03-10
PLFM-8870: Check for null or empty

xschildw:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/ProjectSettingsManagerImpl.java
```diff
@@ -361,6 +361,9 @@ public boolean isStsStorageLocationSetting(ProjectSetting projectSetting) {
 		// Short-cut: Only check the first Storage Location ID. Entities with an StsStorageLocation can't have other
 		// storage locations.
 		List<Long> storageLocationIdList = ((UploadDestinationListSetting) projectSetting).getLocations();
+		if (storageLocationIdList == null || storageLocationIdList.isEmpty()) {
```
john-hill: This fix is in the wrong place.  The actual bug is in validateUploadDestinationListSetting() should throw an exception if the uploadId is null.

john-hill (changes requested):

marcomarasca (approved):

john-hill (approved):

john-hill (approved):

PR #5233, opened 2025-03-11
PLFM-8817: Requires 2fa to be enabled (under feature flag)

marcomarasca:


integration-test/src/test/java/org/sagebionetworks/ITTwoFactorAuthTest.java
```diff
@@ -452,6 +458,44 @@ public void testReset2FaWithTokenWithPassword(SynapseAdminClient adminClient) th
 		}
 	}
 	
+	@Test
+	public void testEnable2FaEnableRequirement(SynapseAdminClient adminClient, SynapseClient synapseClient) throws SynapseException {
```
john-hill, line 462: when ITTestExtension sets up the Admin, enable 2fa for the admin.  This would allow any test to test MFA features using that admin.  This test could validate 2fa required with the other user (admin toggles the feature).

lib/models/src/main/java/org/sagebionetworks/repo/web/TwoFactorAuthEnabledRequiredException.java
```diff
@@ -0,0 +1,11 @@
+package org.sagebionetworks.repo.web;
+
+public class TwoFactorAuthEnabledRequiredException extends RuntimeException {
```
john-hill, line 3: I wold throw the same exception we throw when attempting to download a file that requires 2fa.  The UI does not need a special exception.

integration-test/src/test/java/org/sagebionetworks/ITTwoFactorAuthTest.java
john-hill, line 462: when ITTestExtension sets up the Admin, enable 2fa for the admin.  This would allow any test to test MFA features using that admin.  This test could validate 2fa required with the other user (admin toggles the feature).

lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/feature/Feature.json
```diff
@@ -33,6 +33,10 @@
 		{
 			"name": "ALLOW_AGENT_WRITES",
 			"description": "When enabled, bedrock agents are allowed to directly write to Synapse."
+		},
+		{
+			"name": "REQUIRE_TWO_FA_BYPASS",
+			"description": "When enabled, two factor authentication WON'T be required for users accessing Synapse."
```
john-hill, line 39: DISABLE_2FA_REQUIREMENT
Set this value to "false" to require 2FA for all users.

john-hill (approved):

marcomarasca:
Subsumed by #5234

PR #5243, opened 2025-04-11
PLFM-8885

brucehoff:
Allow clients to revoke refresh tokens using an application/x-www-form-urlencoded request body as per the OAuth spec' https://www.rfc-editor.org/rfc/rfc7009.html#section-2.1


services/repository-managers/src/main/java/org/sagebionetworks/repo/web/controller/JSONEntityHttpMessageConverter.java
```diff
@@ -167,17 +163,41 @@ public static void validateJSONArrayRecursive(JSONArrayAdapter parsedArray,
...
+	
+	public static String convertFormEncodedDataToJSONString(String s) {
+		JSONObject result = new JSONObject();
+		// split by &
+		StringTokenizer st=new StringTokenizer(s, "&");
+		while (st.hasMoreTokens()) {
+			String token = st.nextToken();
+			String[] keyValuePair = token.split("=");
+			if (keyValuePair.length!=2) throw new IllegalArgumentException("Expected key-value pairs separated by '= but found "+s);
+			result.put(keyValuePair[0], keyValuePair[1]);
```
marcomarasca: I think you might have to use a "url decocder" instead of taking the plain split value of the input string.

marcomarasca: Maybe even better, let us look at the standard implementation of the FormHttpMessageConverter

client/synapseJavaClient/src/main/java/org/sagebionetworks/client/SynapseClientImpl.java
```diff
@@ -4769,6 +4770,19 @@ public void revokeToken(OAuthTokenRevocationRequest revocationRequest) throws Sy
...
 
+	@Override
+	public void revokeTokenURLEncoded(String token) throws SynapseException {
+		ValidateArgument.required(token, "token");
+		Map<String, String> headers = new HashMap<String, String>();
+		headers.put(AuthorizationConstants.AUTHORIZATION_HEADER_NAME, getAuthorizationHeader());
+		headers.put(CONTENT_TYPE, "application/x-www-form-urlencoded");
+		String requestBody="token="+token;
+		
+		SimpleHttpResponse response = signAndDispatchSynapseRequest(getAuthEndpoint(),
```
marcomarasca: we do not need to "sign" requests anymore

client/synapseJavaClient/src/main/java/org/sagebionetworks/client/SynapseClientImpl.java
```diff
@@ -4769,6 +4770,19 @@ public void revokeToken(OAuthTokenRevocationRequest revocationRequest) throws Sy
 		voidPost(getAuthEndpoint(), AUTH_OAUTH_2 + REVOKE, revocationRequest, null);
 	}
 
+	@Override
+	public void revokeTokenURLEncoded(String token) throws SynapseException {
+		ValidateArgument.required(token, "token");
+		Map<String, String> headers = new HashMap<String, String>();
+		headers.put(AuthorizationConstants.AUTHORIZATION_HEADER_NAME, getAuthorizationHeader());
+		headers.put(CONTENT_TYPE, "application/x-www-form-urlencoded");
+		String requestBody="token="+token;
```
marcomarasca: The token should be encoded

marcomarasca (changes requested):

marcomarasca (approved):

PR #5247, opened 2025-04-16
PLFM-8897: Boolean mode support for table fts

marcomarasca:


lib/lib-table-query/src/main/java/org/sagebionetworks/table/query/model/TextMatchesMode.java
```diff
@@ -0,0 +1,16 @@
...
+	BOOLEAN("IN BOOLEAN MODE");
+
+	private String sql;
+	
+	TextMatchesMode(String sql) {
+		this.sql = sql;
+	}
+	
+	public String getSql() {
+		return sql;
```
john-hill, line 14: should the "in" be here?

lib/lib-table-query/src/main/resources/table-query-parser.jj
```diff
@@ -337,6 +336,12 @@ TOKEN [IGNORE_CASE]:{
 | < div: "DIV">
 }
 
+/* Text Matches Modes */
+TOKEN [IGNORE_CASE]: {
+  <text_matches_mode_boolean: "BOOLEAN MODE">
+| <text_matches_mode_natural: "NATURAL_LANGUAGE MODE">
```
john-hill: need to remove the underscore

john-hill (changes requested):

john-hill (approved):

PR #5270, opened 2025-06-13
Plfm 9050

john-hill:


lib/models/src/main/java/org/sagebionetworks/repo/model/grid/patch/LogicalTimestamp.java
```diff
@@ -25,6 +25,29 @@ public LogicalTimestamp setSequenceNumber(Long sequenceNumber) {
...
 
+	/**
+	 * Create a new LogicalTimestamp with a sequence that is incremented by the
+	 * provided span.
+	 * 
+	 * @param original
+	 * @param span
+	 * @return
+	 */
+	public static LogicalTimestamp newIncrement(LogicalTimestamp original, long span) {
```
marcomarasca, line 36: I would probably avoid the static if you can

lib/models/src/main/java/org/sagebionetworks/repo/model/grid/patch/operation/InsertValue.java
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	public InsertValue setOperationId(LogicalTimestamp operationId) {
+		this.operationId = operationId;
+		return this;
+	}
+
+	@Override
+	public long span() {
+		return 1L;
```
marcomarasca, line 48: L not needed

lib/models/src/main/java/org/sagebionetworks/repo/model/grid/patch/operation/NewArray.java
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	public NewArray setOperationId(LogicalTimestamp operationId) {
+		this.operationId = operationId;
+		return this;
+	}
+
+	@Override
+	public long span() {
+		return 1L;
```
marcomarasca, line 28: L not needed

lib/models/src/main/java/org/sagebionetworks/repo/model/grid/patch/operation/NewBinary.java
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	@Override
+	public LogicalTimestamp getOperationId() {
+		return id;
+	}
+
+	@Override
+	public long span() {
+		return 1L;
```
marcomarasca, line 23: L not needed

lib/models/src/main/java/org/sagebionetworks/repo/model/grid/patch/compact/NewConstantSerializable.java
```diff
@@ -1,38 +1,93 @@
...
-		if (array.length() > 2) {
+		if (array.length() == 3) {
 			con.setTimestamp(array.getBoolean(2));
 		}
-		return con;
+		if (array.isNull(1)) {
+			return con.setValue(new ConValue(ConType._null, null));
+		}
+
+		Object value = array.get(1);
```
marcomarasca, line 26: Maybe we do not need to care about typing here, consider treating all as a String (we store it as a json blob anyway)

lib/models/src/test/java/org/sagebionetworks/repo/model/grid/patch/LogicalTimestampCompactSerializableTest.java
```diff
@@ -0,0 +1,61 @@
...
+import org.junit.jupiter.api.Test;
+import org.sagebionetworks.repo.model.grid.patch.compact.LogicalTimestampCompactSerializable;
+
+public class LogicalTimestampCompactSerializableTest {
+
+	@Test
+	public void testDeserializeAndSeserialize() {
+		LogicalTimestamp ts = new LogicalTimestamp().setReplicaId(1l).setSequenceNumber(2L);
+		// call under test
+		JSONArray array = LogicalTimestampCompactSerializable.serialize(ts);
```
marcomarasca, line 18: Maybe we should have the assertEqual("[1, 2]", array.toString(2)); Easier to understand

marcomarasca (approved):

PR #5285, opened 2025-06-30
plfm-9093 return optional of document

SandhraSokhal:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/search/oss/SearchManagerImpl.java
```diff
@@ -78,6 +79,9 @@ public void documentChangeMessages(List<ChangeMessage> messages) throws Temporar
         } catch (OpenSearchException | IOException e) {
             log.error(e);
             throw new RecoverableMessageException(e);
+        } catch (Exception e){
```
marcomarasca, line 82: This is not needed

marcomarasca (approved):

PR #5296, opened 2025-07-10
SAGL-90: Invoke lambda using API

xschildw:
Depends on https://github.com/Sage-Bionetworks/synapse-markdown-it-lambda/pull/3 to be deployed before building (need actual function name to call).


lib/lib-markdown/src/main/java/org/sagebionetworks/markdown/MarkdownDaoImpl.java
```diff
@@ -1,27 +1,33 @@
...
+import com.amazonaws.services.lambda.model.AWSLambdaException;
+import com.amazonaws.services.lambda.model.InvokeRequest;
+import com.amazonaws.services.lambda.model.InvokeResult;
 import org.json.JSONException;
 import org.json.JSONObject;
 import org.springframework.beans.factory.annotation.Autowired;
 
+import java.nio.charset.StandardCharsets;
+
 public class MarkdownDaoImpl implements MarkdownDao{
```
marcomarasca, line 15: This should be refactored to use the @Service annotation rather than an xml bean

lib/lib-markdown/src/main/java/org/sagebionetworks/markdown/MarkdownDaoImpl.java
```diff
@@ -1,27 +1,33 @@
...
-	public String getSynapseBaseUrl() {
-		return synapseBaseUrl;
-	}
+	String synapseBaseUrl;
+	String stack;
 
 	public void setSynapseBaseUrl(String synapseBaseUrl) {
 		this.synapseBaseUrl = synapseBaseUrl;
 	}
+	public void setStack(String stack) { this.stack = stack; }
```
marcomarasca: If using the service annotation then you can inject the StackConfiguration to get these properties.

lib/lib-markdown/src/main/java/org/sagebionetworks/markdown/MarkdownDaoImpl.java
```diff
@@ -1,27 +1,33 @@
...
 public class MarkdownDaoImpl implements MarkdownDao{
 
 	public static final String MARKDOWN = "markdown";
 	public static final String OUTPUT = "output";
 	public static final String RESULT = "result";
 	public static final String BASE_URL = "baseURL";
+	public static final String FUNCTION_NAME_FMT = "%s-markdownit:prod";
 
 	@Autowired
-	MarkdownClient markdownClient;
```
marcomarasca, line 15: We can remove the markdown client as it's not used anywhere else.

lib/lib-markdown/src/main/resources/markdown-dao.spb.xml
```diff
@@ -4,7 +4,10 @@
 	xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans-3.0.xsd">
 
 	<import resource="classpath:stack-configuration.spb.xml" />
-	
+
+	<bean id="lambdaClient" class="org.sagebionetworks.aws.AwsClientFactory"
```
marcomarasca: We should get rid of the xml and just use java config classes.

lib/lib-markdown/src/test/java/org/sagebionetworks/markdown/MarkdownDaoImplTest.java
```diff
@@ -1,42 +1,81 @@
...
 
-	@Before
+	@BeforeEach
 	public void before() {
 		MockitoAnnotations.initMocks(this);
 		dao = new MarkdownDaoImpl();
-		ReflectionTestUtils.setField(dao, "markdownClient", mockMarkdownClient);
+		dao.setSynapseBaseUrl("https://www.synapse.org");
+		dao.setStack("dev");
+		ReflectionTestUtils.setField(dao, "lambdaClient", mockLambdaClient);
```
marcomarasca: No need for the utility, use junit 5 @InjectMocks instead

lib/stackConfiguration/pom.xml
```diff
@@ -203,7 +203,11 @@
 			<groupId>software.amazon.awssdk</groupId>
 			<artifactId>apache-client</artifactId>
 		</dependency>
-			
+		<dependency>
+			<groupId>com.amazonaws</groupId>
+			<artifactId>aws-java-sdk-lambda</artifactId>
```
marcomarasca: We should probably use the v2 of the sdk.

lib/stackConfiguration/src/main/java/org/sagebionetworks/aws/AwsClientFactory.java
```diff
@@ -260,5 +263,16 @@ public static AmazonApiGatewayV2 createAmazonApiGatewayClient() {
 				.withCredentials(SynapseAWSCredentialsProviderChain.getInstance())
 				.build();
 	}
+
+	/**
+	 *
+	 * @return An instance of AWSLambdaClient using the Synapse credential chain
+	 */
+	public static AWSLambda createLambdaClient() {
```
marcomarasca: I would really avoid staying on the v1 of the sdk.

lib/lib-markdown/src/test/java/org/sagebionetworks/markdown/MarkdownDaoImplIntegrationTest.java
```diff
@@ -21,15 +21,16 @@ public class MarkdownDaoImplIntegrationTest {
...
 		String rawMarkdown = "## a heading";
 		String outputType = "html";
-		String result = "<h2 toc=\"true\">a heading</h2>\n";
-		assertEquals(result, dao.convertMarkdown(rawMarkdown, outputType));
+		String expectedResult = "<h2 toc=\"true\">a heading</h2>\n";
+		String result = dao.convertMarkdown(rawMarkdown, outputType);
+		assertEquals(expectedResult, result);
 	}
 
 	@Test
```
marcomarasca, line 27: Update to junit 5

lib/lib-markdown/src/main/java/org/sagebionetworks/markdown/MarkdownDaoImpl.java
```diff
@@ -31,11 +37,32 @@ public String convertMarkdown(String rawMarkdown, String outputType) throws JSON
...
+	private String convertToMarkdownWithLambda(String request) throws MarkdownClientException {
+		try {
+			InvokeRequest invokeRequest = new InvokeRequest()
+					.withFunctionName(String.format(FUNCTION_NAME_FMT, stack))
+					.withPayload(request);
+
+			InvokeResult result = lambdaClient.invoke(invokeRequest);
+
+			if (result.getFunctionError() != null) {
+				throw new MarkdownClientException(500, "Lambda execution failed: " + result.getFunctionError());
```
marcomarasca: We should probably log more details from response (the docs says more details are in the result.getPayload(), also the result object has a getStatusCode() function that we can use instead of 500.

marcomarasca (changes requested):

marcomarasca:
I also want to point out that I'm not sure why we need any of this in the first place. As far as I can tell from the code we only use it to convert some markdown into an html email, can't we just use an html template for the email as we do in other places? I do not see this used in the webclient and the current lambda URL is not publicly accessible (but I might be missing something) either so maybe we do not need the lambda at all?

lib/lib-markdown/src/main/java/org/sagebionetworks/markdown/MarkdownConfig.java
```diff
@@ -0,0 +1,33 @@
...
+import org.sagebionetworks.aws.AwsClientFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.ImportResource;
+import org.springframework.context.annotation.Scope;
+import software.amazon.awssdk.services.lambda.LambdaClient;
+
+@Configuration
+@ImportResource("classpath:stack-configuration.spb.xml")
```
marcomarasca, line 12: Very cool, thanks!

lib/lib-markdown/src/test/java/org/sagebionetworks/markdown/MarkdownDaoImplIntegrationTest.java
```diff
@@ -1,24 +1,22 @@
...
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 
-@RunWith(SpringJUnit4ClassRunner.class)
-@ContextConfiguration(locations = {"classpath:test-context.xml"})
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration(classes = {MarkdownConfig.class})
```
marcomarasca, line 12: Nice!

services/repository-managers/src/main/resources/managers-import.xml
```diff
@@ -7,7 +7,6 @@
 	<!-- Load the single cloudwatch consumer -->
 	<import resource="classpath:cloudwatch-spb.xml" />
 	<!-- Markdown service -->
-	<import resource="classpath:markdown-dao.spb.xml" />
```
marcomarasca, line 10: How is the MarkdownConfig imported where it is used? Maybe through the ManagerConfiguration? Unfortunately, the code structure is very convoluted and while this is only used by the workers application it is currently imported by the manager (which is also included in the repo API application).

lib/lib-markdown/src/main/java/org/sagebionetworks/markdown/MarkdownConfig.java
```diff
@@ -0,0 +1,33 @@
...
+    @Scope("singleton")
+    public LambdaClient lambdaClient() {
+        LambdaClient lambdaClient = AwsClientFactory.createLambdaClient();
+        return lambdaClient;
+    }
+
+    @Bean
+    @Scope("singleton")
+    public MarkdownDao markdownDao() {
+        MarkdownDaoImpl markdownDao = new MarkdownDaoImpl(lambdaClient(), stackConfiguration.getSynapseBaseUrl(), stackConfiguration.getStack());
```
marcomarasca, line 23: The StackConfiguration instance can be a parameter of this method rather than an object property, the LambdaClient does not need to be an exported bean actually, if it needs to be "shared" I would rather have it exposed as a bean somewhere else. In other words, I think it's best if the MarkdownConfig does not add to the context beans that are not related to this module.

PR #5297, opened 2025-07-11
PLFM-9086: Multi-value column facet CI

marcomarasca:


lib/lib-table-cluster/src/main/java/org/sagebionetworks/table/cluster/SQLTranslatorUtils.java
```diff
@@ -1065,25 +1065,35 @@ private static SearchCondition createArrayHasLikeJsonSearchSearchCondition(Array
 		return new SearchCondition(orTerms);
 	}
 	
-	// Translate the expression foo HAS (1, 2, 3) -> JSON_OVERLAPS(foo, JSON_ARRAY(1, 2, 3)) IS TRUE
+	// Translate the expression foo HAS (1, 2, 3) -> JSON_OVERLAPS(lower(foo), JSON_ARRAY(lower(1), lower(2), lower(3))) IS TRUE
```
john-hill, line 1068: can lower the entire array.

john-hill (approved):

PR #5306, opened 2025-07-23
PLFM-9133 Allow Docker registry a wider range of media types in the events sent to Synapse.

brucehoff:
Allow Docker registry a wider range of media types in the events sent to Synapse.


SandhraSokhal (approved):
LGTM

PR #5311, opened 2025-07-27
PLFM-9088 - follow-up

nickgros:
Fix issue inserting empty maps, which created an invalid patch that could not be parsed by json-joy.


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/grid/PatchRowHandler.java
```diff
@@ -172,10 +172,12 @@ private NewObject getSynapseRowMetadata(Row row) {
...
-		// fill the `synapseRow` object
-		currentPatch.addNewOperation(InsertObject.class)
-				.setObjectId(synapseRowMetadata.getOperationId())
-				.setMap(synapseRowMetadataObjectMap);
+		if (!synapseRowMetadataObjectMap.isEmpty()) {
+			// fill the `synapseRow` object
+			currentPatch.addNewOperation(InsertObject.class)
+					.setObjectId(synapseRowMetadata.getOperationId())
+					.setMap(synapseRowMetadataObjectMap);
+		}
```
nickgros, line 180: Check if the map is empty before inserting.

Ideally, we would not need this check, and the `Patch` object would just treat this as a no-op, but that would be a substantial refactor.

lib/lib-grid/src/main/java/org/sagebionetworks/repo/model/grid/patch/compact/InsertObjectSerializable.java
```diff
@@ -38,6 +39,13 @@ public InsertObject deserialize(LogicalTimestamp operationId, JSONArray array) {
 
 	@Override
 	public JSONArray serialize(InsertObject operation) {
```
nickgros, line 41: Add the check to the `InsertObjectSerializable.serialize` method, because it is at this point we consider the operation to become immutable and must be valid.

john-hill (approved):

PR #5315, opened 2025-07-31
PLFM-9032 - Scalability testing & perf improvements

nickgros:
- `Patch`: cache span and return immutable operations
- Remove FK references from Grid Replica tables to index table to speed up insertion of large batches of operations
- Add disabled scalability test.
- Updated `Operation` classes
  - All operation classes are now immutable
  - All operation classes now validate that required properties are provided in the constructor
- Added OperationBuilder classes to support creating operations within a patch.


lib/lib-grid-db/src/main/resources/grid-db-spb.xml
```diff
@@ -14,7 +14,9 @@
 
 	<!-- Trun on Spring's auotproxy using AspectJ's @Aspect annotations. -->
 	<aop:aspectj-autoproxy />
-	
+
+	<bean id="gridIndexManager" class="org.sagebionetworks.grid.db.GridIndexManagerImpl" scope="singleton" />
```
john-hill: This should not be needed.

lib/lib-grid-db/src/main/resources/schema/Grid-Array-ddl.sql
```diff
@@ -12,6 +12,6 @@ CREATE TABLE IF NOT EXISTS GRID_REPLICA_ARR (
     IS_DELETED BOOLEAN DEFAULT FALSE, -- on delete, set to true, indicating that the node was deleted.
 	PRIMARY KEY (SESSION_ID, REPLICA_ID, ARR_REP, ARR_SEQ, NODE_REP, NODE_SEQ),
 	UNIQUE INDEX (SESSION_ID, REPLICA_ID, ARR_REP, ARR_SEQ, REF_REP, REF_SEQ),
-    CONSTRAINT `GRID_ARR_REF_FK` FOREIGN KEY (SESSION_ID, REPLICA_ID, ARR_REP, ARR_SEQ, REF_REP, REF_SEQ) REFERENCES GRID_REPLICA_ARR(SESSION_ID, REPLICA_ID, ARR_REP, ARR_SEQ, NODE_REP, NODE_SEQ) ON DELETE CASCADE,
-	CONSTRAINT `GRID_ARR_ARR_INDEX_FK` FOREIGN KEY (SESSION_ID, REPLICA_ID, ARR_REP, ARR_SEQ) REFERENCES GRID_REPLICA_INDEX(SESSION_ID, REPLICA_ID, NODE_REP, NODE_SEQ) ON DELETE CASCADE
+    CONSTRAINT `GRID_ARR_REPLICA_FK` FOREIGN KEY (SESSION_ID, REPLICA_ID) REFERENCES GRID_REPLICA(SESSION_ID, REPLICA_ID) ON DELETE CASCADE,
+    CONSTRAINT `GRID_ARR_REF_FK` FOREIGN KEY (SESSION_ID, REPLICA_ID, ARR_REP, ARR_SEQ, REF_REP, REF_SEQ) REFERENCES GRID_REPLICA_ARR(SESSION_ID, REPLICA_ID, ARR_REP, ARR_SEQ, NODE_REP, NODE_SEQ) ON DELETE CASCADE
```
john-hill: Remove this second FK constraint?

lib/lib-grid/src/main/java/org/sagebionetworks/repo/model/grid/patch/Patch.java
```diff
@@ -37,28 +40,51 @@ public List<Operation<?>> getOperations() {
...
 			if (operations == null) {
 				operations = new ArrayList<>();
 			}
 			operation.setOperationId(LogicalTimestamp.newIncrement(patchId, getSpan()));
 			operations.add(operation);
-			return operation;
-		} catch (InstantiationException | IllegalAccessException | IllegalArgumentException | InvocationTargetException
-				| NoSuchMethodException | SecurityException e) {
+			span += operation.getSpan();
+			return ImmutableOperation.of(operation);
```
john-hill: Constructor would work here.

lib/lib-grid/src/main/java/org/sagebionetworks/repo/model/grid/patch/Patch.java
```diff
@@ -37,28 +40,51 @@ public List<Operation<?>> getOperations() {
...
-	 * 
+	 *
 	 * @param <T>
 	 * @param clazz
-	 * @return
+	 * @return the operation, wrapped to ensure immutability. Note that any changes to the operation after this call may
+	 *   break the patch
 	 */
-	public <T extends Operation<T>> T addNewOperation(Class<? extends T> clazz) {
+	public <T extends Operation<T>> ImmutableOperation<T> addNewOperation(Class<? extends T> clazz) {
```
john-hill: remove this one.

john-hill (changes requested):
There are still ways to mutate the span after and add, this could lead to  some difficult to debug problems if exploited.
Can we find a way to do this where it is impossible to mutate the span?

lib/lib-grid-db/src/main/resources/schema/Grid-Array-ddl.sql
```diff
@@ -12,6 +12,5 @@ CREATE TABLE IF NOT EXISTS GRID_REPLICA_ARR (
     IS_DELETED BOOLEAN DEFAULT FALSE, -- on delete, set to true, indicating that the node was deleted.
 	PRIMARY KEY (SESSION_ID, REPLICA_ID, ARR_REP, ARR_SEQ, NODE_REP, NODE_SEQ),
 	UNIQUE INDEX (SESSION_ID, REPLICA_ID, ARR_REP, ARR_SEQ, REF_REP, REF_SEQ),
-    CONSTRAINT `GRID_ARR_REF_FK` FOREIGN KEY (SESSION_ID, REPLICA_ID, ARR_REP, ARR_SEQ, REF_REP, REF_SEQ) REFERENCES GRID_REPLICA_ARR(SESSION_ID, REPLICA_ID, ARR_REP, ARR_SEQ, NODE_REP, NODE_SEQ) ON DELETE CASCADE,
```
nickgros, line 15: Removed self-referential FK constraint

lib/lib-grid-db/src/main/resources/schema/Grid-Array-ddl.sql
```diff
@@ -12,6 +12,5 @@ CREATE TABLE IF NOT EXISTS GRID_REPLICA_ARR (
...
-	CONSTRAINT `GRID_ARR_ARR_INDEX_FK` FOREIGN KEY (SESSION_ID, REPLICA_ID, ARR_REP, ARR_SEQ) REFERENCES GRID_REPLICA_INDEX(SESSION_ID, REPLICA_ID, NODE_REP, NODE_SEQ) ON DELETE CASCADE
```
nickgros, line 16: In all tables, replace FK reference to INDEX table with reference to GRID_REPLICA table (supports ON DELETE CASCADE without exponential complexity for insert operations)

lib/lib-grid-db/src/test/java/org/sagebionetworks/grid/db/handler/InsertArrayHandlerTest.java
```diff
@@ -44,14 +44,20 @@ public void before() {
...
 				// one
-				new InsertArray().setArrayId(ids.get(0))
-						.setOperationId(new LogicalTimestamp().setReplicaId(99L).setSequenceNumber(1L))
-						.setReferenceId(ids.get(0)).setElementIds(List.of(ids.get(1), ids.get(2))),
+				new InsertArray(
+						new LogicalTimestamp().setReplicaId(99L).setSequenceNumber(1L),
+						ids.get(0),
+						ids.get(0),
+						List.of(ids.get(1), ids.get(2))
+				),
```
nickgros, line 53: Operation instances are now immutable, so all required properties must be provided in the constructor.

lib/lib-grid-db/src/test/java/org/sagebionetworks/grid/db/GridIndexManagerAutowiredTest.java
```diff
@@ -0,0 +1,121 @@
...
+                log.info("Applying patch {} of {}: {}", i + 1, patches.size(), p.getPatchId());
+            }
+            gridIndexManager.applyPatch(sessionId, replicaId, p);
+        }
+    }
+
+
+    @Test
+    @Timeout(value = 5, unit = TimeUnit.MINUTES)
+    @Disabled
```
nickgros, line 70: Note that this test is disabled

lib/lib-grid/src/main/java/org/sagebionetworks/repo/model/grid/patch/operation/builder/InsertArrayBuilder.java
```diff
@@ -0,0 +1,34 @@
+package org.sagebionetworks.repo.model.grid.patch.operation.builder;
+
+import org.sagebionetworks.repo.model.grid.patch.LogicalTimestamp;
+import org.sagebionetworks.repo.model.grid.patch.operation.InsertArray;
+
+import java.util.List;
+
+public final class InsertArrayBuilder extends OperationBuilder<InsertArray, InsertArrayBuilder> {
```
nickgros: Added builder classes for all operations. Each builder exposes methods to set all properties except for the operation ID, which is passed as an argument to the `build` method (which is intended to be called by the `Patch` class), which returns the immutable operation.

lib/lib-grid/src/main/java/org/sagebionetworks/repo/model/grid/patch/operation/builder/Operations.java
```diff
@@ -0,0 +1,49 @@
+package org.sagebionetworks.repo.model.grid.patch.operation.builder;
+
+/**
+ * Utility class for creating instances of various operations.
+ * This class provides static methods to create builders for different types of operations.
+ * Each method returns a builder that can be used to construct the operation with specific parameters.
+ */
+public final class Operations {
```
nickgros, line 8: Utility class containing static methods, providing a single entrypoint for building operations.

lib/lib-grid/src/main/java/org/sagebionetworks/repo/model/grid/patch/Patch.java
```diff
@@ -37,28 +38,29 @@ public List<Operation<?>> getOperations() {
...
-	 * operation will be issued a correct operationId.
-	 * 
-	 * @param <T>
-	 * @param clazz
-	 * @return
+	 * Factory Method that accepts any valid OperationBuilder.
+	 * It generates the ID and asks the builder to construct the final object.
 	 */
-	public <T extends Operation<T>> T addNewOperation(Class<? extends T> clazz) {
+	public <T extends Operation<T>> T addNewOperation(OperationBuilder<T, ?> builder) {
```
nickgros: You must pass an `OperationBuilder` to add an operation to the patch. The patch will return the new, immutable operation for the caller to reference.

lib/lib-grid/src/test/java/org/sagebionetworks/repo/model/grid/node/ObjectNodeTest.java
```diff
@@ -164,48 +161,16 @@ public void testAttemptInsertWithStartingNull() {
...
-
-		InsertObject change = new InsertObject().setObjectId(ids.get(0));
-		change.setMap(null);
-
-		// call under test
-		assertFalse(node.attemptInsert(change));
-		ObjectNode expected = new ObjectNode().setId(ids.get(0));
-		expected.setValue(null);
-		assertEquals(expected, node);
-	}
```
nickgros, line 180: It is no longer possible to create operations that are missing required properties, so these tests can be removed.

nickgros:
Split to #5323 and #5324

PR #5348, opened 2025-08-25
PLFM-9161: Add support for RGA delete (arr only)

marcomarasca:


lib/lib-grid-db/src/test/java/org/sagebionetworks/grid/db/GridIndexDaoImplTest.java
```diff
@@ -68,7 +69,7 @@ public void before() {
 
 	@AfterEach
 	public void after() {
-		gridIndexDao.truncateAll();
+		//gridIndexDao.truncateAll();
```
john-hill, line 72: would you make me remove this?

lib/lib-grid-db/src/test/java/org/sagebionetworks/grid/db/GridIndexDaoImplTest.java
```diff
@@ -864,6 +865,82 @@ public void testGetRootObjectWithRootNotAnObject() {
...
+		);
+		
+		valuesOne.forEach(a -> {
+			assertEquals(Optional.of(a.getReferenceNodeId()), gridIndexDao.findArrayInsertLocation(sessionIdOne, replicaIdOne, a));
+			gridIndexDao.insertIntoArray(sessionIdOne, replicaIdOne, a);
+		});
+		
+		assertEquals(valuesOne, gridIndexDao.getArrayNodesInOrder(sessionIdOne, replicaIdOne, arrOneId, limit, offset));
+		
+	}
```
john-hill, line 943: we might need a test for two session.
case where only one is deleted then value inserted at its location.

john-hill (approved):

PR #5331, opened 2025-08-18
PLFM-8871: ACLs for OAuth Clients

brucehoff:
Today our collaborators can't share maintenance responsibility for the OAuth Clients they create in Synapse, a serious shortcoming.  We add ACLs (sharing settings) in 3 phases:
- Add the object and CRUD operations;
- Backfill ACLs for existing OAuth Clients (of which there are only ~200);
- Change the existing OAuth Client services to make authorization decisions based on the ACLs.

This PR contains the first two phases.

PR #5351, opened 2025-08-26
PLFM-8871

brucehoff:
The previous PR introduced ACLs for OAuth clients and a service to backfill ACLs for existing clients.
This PR changes the authorization logic for OAuth client services to make use of the ACLs.


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/auth/OAuthClientDaoImpl.java
```diff
@@ -60,8 +74,25 @@ public class OAuthClientDaoImpl implements OAuthClientDao {
...
+			TABLE_RESOURCE_ACCESS+" ra, "+
+			TABLE_RESOURCE_ACCESS_TYPE+" rat "+
+			"WHERE "+
+			" oc."+COL_OAUTH_CLIENT_ID+"=acl."+COL_ACL_OWNER_ID+" and "+
+			" acl."+COL_ACL_OWNER_TYPE+"='OAUTH_CLIENT'"+
+			" and ra."+COL_RESOURCE_ACCESS_OWNER+"=acl."+COL_ACL_ID+
+			" and ra."+COL_RESOURCE_ACCESS_GROUP_ID+" in (:"+PRINCIPAL_IDS_BIND_VAR+")"+
+			" and ra."+COL_RESOURCE_ACCESS_ID + " = rat." + COL_RESOURCE_ACCESS_TYPE_ID+
+			" and rat."+COL_RESOURCE_ACCESS_TYPE_ELEMENT+"=:"+ACCESS_TYPE_BIND_VAR+
+			" LIMIT :"+LIMIT_PARAM_NAME+" OFFSET :"+OFFSET_PARAM_NAME+" ";
```
marcomarasca, line 96: Always better to be explicit about the order by when using limit and offset (in this case most likely mysql will order by the oc primary key, but it is not a guarantee). For example, oc.id or something else (e.g. creation date?)

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/oauth/OAuthClientManagerImpl.java
```diff
@@ -231,21 +228,23 @@ public OAuthClient createOpenIDConnectClient(UserInfo userInfo, OAuthClient oaut
 	@Override
 	public OAuthClient getOpenIDConnectClient(UserInfo userInfo, String id) {
 		OAuthClient result = oauthClientDao.getOAuthClient(id);
-		if (!canAdministrate(userInfo, result.getCreatedBy())) {
+		if (!authManager.canAccess(userInfo, id, ObjectType.OAUTH_CLIENT, ACCESS_TYPE.UPDATE).isAuthorized()) {
```
marcomarasca, line 231: We should update the documentation in the OpenIDConnectController to explain the access control model for the oauth clients. I understand that using the ACCESS_TYPE.UPDATE makes it so that we do not have to change the UI code (e.g. there is no need to check if one has permissions to "update" a client from the list because the returned list only contains the clients that the user can update) but it is important to document this somewhat unusual API detail (e.g. usually we list things that the user has read permission for).

marcomarasca (changes requested):

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/oauth/OAuthClientManagerImpl.java
brucehoff: I agree and this was part of the previous [PR](https://github.com/Sage-Bionetworks/Synapse-Repository-Services/pull/5331/files#diff-f6b8d6433b36fcbfc98df739b245a23812d2f5785bd6e97d39301d3c08fff331R228).  Please have a look and say if you feel more is needed.

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/auth/OAuthClientDaoImpl.java
brucehoff: Good catch!

marcomarasca (approved):

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/oauth/OAuthClientManagerImpl.java
marcomarasca: I was specifically talking about the "list" operations on the client (https://rest-docs.synapse.org/rest/GET/oauth2/client.html), the current documentation says:

"List the OAuth 2.0 clients created by the current user."

I would add that the user will see all the clients for which they have the UPDATE permissions.

Similarly, the update https://rest-docs.synapse.org/rest/PUT/oauth2/client/id.html states that "Note: Only the creator of a client can update it."

PR #5352, opened 2025-08-27
plfm-9173 send less search request with exponential wait time

SandhraSokhal:


services/workers/src/test/java/org/sagebionetworks/search/workers/sqs/search/SearchIndexWorkerIntegrationTest.java
```diff
@@ -37,7 +37,7 @@
 @ExtendWith(SpringExtension.class)
 @ContextConfiguration(locations = { "classpath:test-context.xml" })
 public class SearchIndexWorkerIntegrationTest {
-    private static final long MAX_WAIT = 2 * 60*1000; // 2 minutes
+    private static final long MAX_WAIT = 60*1000; // 1 minute
```
marcomarasca: If the process timed out in a 2 minute window, how can it work in half the time?

services/workers/src/test/java/org/sagebionetworks/search/workers/sqs/search/SearchIndexWorkerIntegrationTest.java
```diff
@@ -141,9 +140,22 @@ private static SearchQuery searchQueryByTerm(String term) {
 
     public void waitForQuery(UserInfo user, String term, long expected) throws Exception {
         SearchQuery searchQuery = new SearchQuery().setQueryTerm(Arrays.asList(term));
-        TimeUtils.waitFor(MAX_WAIT, CHECK_TIME, () -> {
-            System.out.println("Waiting for search query: "+searchQuery);
-            return Pair.create(searchManager.search(user, searchQuery).getFound() == expected, null);
-        });
+        long startTime = System.currentTimeMillis();
```
marcomarasca: TimeUtils already has a waitForExponential method, no need to reimplement this. But I do not understand how exponential backoff solves the timeout? As far as I understand either the process of ingesting the documents and indexing them (plus propagation to all the nodes) in opensearch takes more than the timeout (2 mins) or there is some issue with the search logic or the condition check (e.g. are we sure the check ...getFound() == expected is the right thing to do?).

marcomarasca (changes requested):
See comments.

PR #5354, opened 2025-08-29
PLFM-9194: Add RecordSet entity

marcomarasca:


lib/lib-shared-models/src/main/java/org/sagebionetworks/repo/model/table/ViewTypeMask.java
```diff
@@ -1,10 +1,11 @@
 package org.sagebionetworks.repo.model.table;
 
 import org.sagebionetworks.repo.model.EntityType;
+import org.sagebionetworks.repo.model.EntityTypeUtils;
 
 public enum ViewTypeMask {
 
-	File(0x01, EntityType.file),
+	File(0x01, EntityTypeUtils.getFileTypes().toArray(new EntityType[0])),
```
john-hill: Think we are going to need a new type for views (can include only RecordSets or exclude them).

services/repository/src/test/java/org/sagebionetworks/repo/web/service/EntityServiceImplAutowiredTest.java
```diff
@@ -177,11 +180,11 @@ public void after(){
 		if(toDelete != null){
 			for(String id: toDelete){
 				try {
-					entityService.deleteEntity(adminUserId, id);
+					//entityService.deleteEntity(adminUserId, id);
```
john-hill: put it back?

john-hill (changes requested):
Think we need a new type for views.

john-hill (approved):

PR #5355, opened 2025-08-29
PLFM-9207

brucehoff:
Aridhia uses Azure B2C as an OAuth client and found that the client fails when the 'modulus' part of the Synapse public key is padding with a leading zero.  This PR corrects the problem by encoding the BigInteger as an 'unsigned' integer, which ensures it's always the correct 256 bytes.


SandhraSokhal (approved):
LGTM

PR #5387, opened 2025-09-25
PLFM-9198: Emit patches for the upsert changes

marcomarasca:


lib/jdomodels/src/main/resources/schema/grid/Grid-Connection-ddl.sql
```diff
@@ -5,7 +5,7 @@ CREATE TABLE IF NOT EXISTS `GRID_CONNECTION` (
   `REPLICA_ID` BIGINT NOT NULL,
   `CREATED_BY` BIGINT NOT NULL,
   `CREATED_ON` TIMESTAMP(3) NOT NULL,
-  `SOURCE` ENUM ('WEBSOCKET','INTERNAL','VALIDATION','AGENT'),
+  `SOURCE` ENUM ('WEBSOCKET','INTERNAL','VALIDATION','AGENT','USER_SUPPORT'),
```
john-hill, line 8: if we merge this to staging you will need to patch the DB.

lib/lib-grid-db/src/main/java/org/sagebionetworks/grid/db/GridIndexDao.java
```diff
@@ -201,6 +201,17 @@ public interface GridIndexDao extends ConstantProvider {
...
+
+	/**
+	 * 
+	 * @param sessionIdString
+	 * @param replicaId
+	 * @param arrayId
+	 * 
+	 * @return The last node in the given array, empty if the array has no nodes.
+	 */
+	Optional<ArrayNode> getArrayLastNode(String sessionIdString, Long replicaId, LogicalTimestamp arrayId);
```
john-hill, line 213: If there is no node then the array is the node.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/grid/PatchUtils.java
```diff
@@ -30,5 +39,30 @@ public static int calculateRowsPerPatch(Long maxRowSizeBytes) {
 	public static Long plusTenPerent(Long value) {
 		return value + Double.valueOf(value * 0.1).longValue();
 	}
+	
+	public static ConType getConType(Object value) {
```
john-hill, line 43: can this move to the ConType Enum?

services/workers/src/test/java/org/sagebionetworks/grid/workers/GridEventBrokerWorkerIntegrationTest.java
```diff
@@ -753,7 +753,31 @@ public void testGridWithRecordSet() throws Exception {
...
+			}
+			
+			// Also wait for the validation results to be set, the last row should be valid
+			return Pair.create(
+				new ValidationResults().setIsValid(true).equals(page.get(5).getRowValidationResults()), 
+				page
+			);
+		});
+		
+		assertEquals(
```
john-hill, line 770: if this is in the wait, then there is no chance that there are 6 rows that are not exactly what you are looking for.

john-hill (approved):

PR #5390, opened 2025-09-27
PLFM-9092

brucehoff:
send status to GitHub; 
cancel CP when Wkflow is cancelled; 
allow passing EXTRA_ARGS; 
set bucket lifecycle


.github/workflows/build_and_test.yml
```diff
@@ -31,6 +35,7 @@ jobs:
       statuses: write
       actions: write
       checks: write
+      pull-requests: write
```
marcomarasca: Maybe this is not needed?

.github/workflows/build_and_test.yml
```diff
@@ -9,6 +9,10 @@
 # - STACK_BUILDER_REPO_OWNER: defaults to Sage-Bionetworks
 # - STACK_BUILDER_BRANCH: defaults to develop
 # - FULL_BUILD: if true, will do a full build, vs. an abbreviated "feature build"
+# - EXTRA_ARGS: pass extra arguments to maven build
+#
+# To post the build status to a PR, add this repository secret:
+# - WRITE_STATUS_TOKEN: A GitHub token with "repo:status" and "public_repo" permissions
```
marcomarasca, line 15: Maybe we can use some sort of application token from github (https://docs.github.com/en/apps/creating-github-apps/about-creating-github-apps/about-creating-github-apps) so that the user doesn't have to setup their own PAT?

configuration/build/buildspec.yml
```diff
@@ -56,4 +56,30 @@ phases:
...
+      function post_github_status() {
+        echo GITHUB_TOKEN length ${#GITHUB_TOKEN}
+        echo GIT_COMMIT $GIT_COMMIT
+        if [[ $GITHUB_TOKEN && $GIT_COMMIT ]]; then
+          echo "About to set status"
+          curl "https://api.github.com/repos/Sage-Bionetworks/Synapse-Repository-Services/statuses/$GIT_COMMIT" \
+            -H "Authorization: token $GITHUB_TOKEN" \
+            -H "Content-Type: application/json" \
+            -X POST \
+            -d "{\"state\": \"$1\", \"description\": \"$2\"}"
```
marcomarasca: What about the build url? Can we generate the URL to the build (maybe action, maybe code pipeline)?

configuration/build/buildspec.yml
```diff
@@ -56,4 +56,30 @@ phases:
...
+        fi
+      }
+
+      post_github_status pending "Build in Progress"
+
       ./docker_build.sh
+
+      if [[ $? -eq 0 ]]; then
```
marcomarasca: If we cancel the pipeline this won't execute. We should probably move this into the github workflow.

marcomarasca (changes requested):
See comments

.github/workflows/build_and_test.yml
brucehoff: removed

configuration/build/buildspec.yml
brucehoff: Done!

configuration/build/buildspec.yml
brucehoff: Done!

.github/workflows/build_and_test.yml
brucehoff: I wrote a follow on issue for this idea: https://sagebionetworks.jira.com/browse/PLFM-9259

marcomarasca (approved):
Looks great!

PR #5393, opened 2025-09-30
This change resolves a race condition where messages from the same re…

john-hill:
…plica were processed out of order.

    Sequential Processing: The root cause was the use of a standard SQS queue. This has been replaced with a FIFO queue, using the connection-id as the MessageGroupId. This guarantees that all messages from a single connection are processed in order, which also eliminates the need for a semaphore lock when writing to the replica.

    Performance Regression: The switch to sequential processing introduced a new problem. Duplicate clock-synchronization calls, which were previously resolved concurrently, would now run sequentially, causing the hub to waste time reapplying patches that had already been processed.

    Duplicate Prevention & Safety: To fix this, we now block new clock-synchronization chains from starting if one is already active. To prevent a lost completion message from blocking a replica indefinitely, these chain records in the database now have an expiration time. The expiration is refreshed each time a new patch is applied, ensuring stale locks are eventually cleared.


lib/lib-grid-db/src/main/java/org/sagebionetworks/grid/db/GridIndexDaoImpl.java
```diff
@@ -623,17 +624,20 @@ public Integer createNextMessageId(String sessionIdString, Long replicaId, int m
...
 		ValidateArgument.required(chain.getMethod(), "chain.method");
+		ValidateArgument.required(expires, "chain.expires");
 		Long sessionId = validateReplica(chain.getSessionId(), chain.getReplicaId());
 		jdbcTemplate.update(
-				"INSERT INTO GRID_REPLICA_MESSAGE (SESSION_ID, REPLICA_ID, MESSAGE_ID, METHOD_NAME, CREATED_ON)"
-						+ " VALUES (?,?,?,?,NOW()) ON DUPLICATE KEY UPDATE METHOD_NAME = ?, CREATED_ON = NOW()",
-				sessionId, chain.getReplicaId(), chain.getId(), chain.getMethod(), chain.getMethod());
+				"INSERT INTO GRID_REPLICA_MESSAGE (SESSION_ID, REPLICA_ID, MESSAGE_ID, METHOD_NAME, CREATED_ON, EXPIRES_On)"
+						+ " VALUES (?,?,?,?,NOW(),NOW() + INTERVAL ? SECOND) ON DUPLICATE KEY"
+						+ " UPDATE METHOD_NAME = ?, CREATED_ON = NOW(), EXPIRES_ON = NOW() + INTERVAL ? SECOND",
```
marcomarasca, line 638: We have to add the column to the staging DB before deploying this.

lib/lib-grid-db/src/main/java/org/sagebionetworks/grid/db/GridIndexDaoImpl.java
```diff
@@ -650,6 +654,31 @@ public Optional<MessageChain> getMessageChain(String sessionIdString, Long repli
...
+	@Override
+	public Optional<MessageChain> getNonExpiredMessageChain(String sessionIdString, Long replicaId, String method) {
+		Long sessionId = validateReplica(sessionIdString, replicaId);
+		ValidateArgument.required(method, "method");
+		try {
+			return jdbcTemplate
+					.query("SELECT * FROM GRID_REPLICA_MESSAGE WHERE SESSION_ID = ? AND REPLICA_ID = ? AND METHOD_NAME = ?"
+							+ " AND EXPIRES_ON > NOW() LIMIT 1", MESSAGE_CHAIN_MAPPER, sessionId, replicaId, method)
+					.stream().findFirst();
+		} catch (EmptyResultDataAccessException e) {
```
marcomarasca, line 677: THere is no Empty result thrown

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/grid/internal/replica/change/PatchBuilderPublisherImpl.java
```diff
@@ -27,13 +27,12 @@ public void sendChangesToPatchBuilder(IntendedChangeSet changeSet) {
...
-		 * conflicting patches. Messages are grouped by 'sessionId-replicaId' to ensure
+		 * conflicting patches. Messages are grouped by 'connectionId' to ensure
 		 * the FIFO queue triggers the creation of all patches for the same replica
 		 * sequentially.
 		 */
-		String groupId = String.format("%s-%d", changeSet.getSessionId(), changeSet.getReplicaId());
-		sqsClient.sendMessage(SendMessageRequest.builder().queueUrl(queueUrl)
-				.messageDeduplicationId(DigestUtils.md5Hex(body)).messageGroupId(groupId).messageBody(body).build());
+		sqsClient.sendMessage(SendMessageRequest.builder().queueUrl(queueUrl).messageGroupId(changeSet.getConnectionId())
+				.messageDeduplicationId(DigestUtils.md5Hex(body)).messageBody(body).build());
```
marcomarasca, line 35: Maybe use the UUID to be more efficient? Since this is going to be very unique anyway

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/grid/internal/replica/GridReplicaManagerImpl.java
```diff
@@ -66,29 +68,17 @@ void sendClockMessage(Integer methodId, String connectionId, List<LogicalTimesta
 
 	@Override
 	public void onResponseComplete(ProgressCallback callback, GridConnectionInfo connection, Integer methodId) {
```
marcomarasca, line 70: callback is unused

marcomarasca (approved):

PR #5394, opened 2025-09-30
PLFM-9253

brucehoff:
Replicates the Jenkins jobs, `repo-develop-build` and `repo-release-build`:

For builds running in `Sage-Bionetworks`, maven "deploy" is used.  If the branch is "develop" then the artifact versions are set to date+short hash.  If the branch is "release-*" then the artifact versions are replaced with the latest tag on the repo'.


configuration/build/buildspec.yml
```diff
@@ -56,4 +66,16 @@ phases:
         unset org_sagebionetworks_google_cloud_key
       fi
 
+      if [[ "${PUBLISH_ARTIFACTS}" ]]; then
+        export build_deploy=true
+        export artifactory_username=$(echo $ARTIFACTORY_CREDS | jq -r '.artifactory_username')
+        export artifactory_password=$(echo $ARTIFACTORY_CREDS | jq -r '.artifactory_password')
+        if [[ "${PUBLISH_ARTIFACTS}" == "develop" ]]; then
+          SHORT_HASH=${GIT_COMMIT:0:7}
```
marcomarasca: I'm not sure we need to pass the short hash to this pomSnapshotToGitHash.sh script:

```
buildNumber=$1
if [ -z "$buildNumber" ] ;
  then buildNumber="??"
fi
echo $buildNumber
abbreviatedCommitHash="<version>"$(date +%Y-%m-%d)"-"$buildNumber"-"`git log -n 1 --pretty=format:%h`"</version>"
```

The "buildNumber" is the first argument of the script, in jenkins we use the ${BUILD_NUMBER} variable which is the job build number. Note that the script already appends the short git hash above (`git log -n 1 --pretty=format:%h`).

configuration/build/buildspec.yml
```diff
@@ -56,4 +66,16 @@ phases:
...
+          ./pomSnapshotToGitHash.sh ${SHORT_HASH}
+        elif [[ "${PUBLISH_ARTIFACTS}" == "release" ]]; then
+          ./pomSnapshotToLastGitTag.sh ${LATEST_TAG}
```
marcomarasca: I see that you extracted the logic to get the latest tag from the script and moved it to the github workflow, is there a specific reason for it? I also wonder if it might be better to move the logic to decide if the pipeline should publish the artifact to antifactory (along with which version to publish) in the github workflow itself.

marcomarasca (changes requested):

configuration/build/buildspec.yml
brucehoff: Whoops, I did not see that 'git' command.  We need to modify the script to work without git commands (just accept the short hash passed in).

marcomarasca (changes requested):
Looks great! In order to get a build, you have to merge from develop because some of the queues in the stack develop branch were recently renamed.

marcomarasca (approved):

PR #5401, opened 2025-10-02
PLFM-9269: Allow grid creation from another user record set

marcomarasca:
Allows the creation of a grid from any record set that the user has download access to (since we need to read the content of the record set in order to instantiate the grid).


nickgros (approved):
Lgtm! Ok for this to go into either release-562 or develop (I stumbled upon this when trying to validate other changes, but I don't think this blocks demos)

PR #5404, opened 2025-10-02
when the agent requests to filter rows based on the user's selection,…

john-hill:
… the correct replicaID is used in the filter.  Also, when the agent provides invalid JSON requests, we now include the provided value in our error response to the agent.  This change seems to result in the agent correcting the problem after a single failure.

### DIFF
--- a/services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/agent/handler/ReturnControlEvent.java
+++ b/services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/agent/handler/ReturnControlEvent.java
@@ -79,21 +79,26 @@ public class ReturnControlEvent {
 		JSONObject object = new JSONObject();
 		try {
 			requestBodyParameters.forEach(p -> {
-				if ("object".equals(p.getType())) {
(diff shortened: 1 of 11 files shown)


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/agent/handler/ReturnControlEvent.java
```diff
@@ -79,21 +79,26 @@ String extractRequestBodyFromParameters() {
...
+				} catch (JSONException e) {
+					throw new IllegalArgumentException("Failed to parse the JSON value for parameter '" 
+						+ p.getName() + "' with value: " + p.getValue() + ". Error: " + e.getMessage(), e);
 				}
 			});
 			return object.toString();
-		} catch (JSONException e) {
-			throw new IllegalArgumentException("Failed to parse the JSON request body: " + e.getMessage(), e);
+		} catch (IllegalArgumentException e) {
+			throw e;
```
nickgros, line 101: Could probably remove this catch

nickgros (approved):

PR #5408, opened 2025-10-03
PLFM-9274: Cleanup temp tables after import

marcomarasca:
Temporary tables are tied to a session, but they need to be cleaned up since we use a connection pool where a connection might be reused. I also tied the name of the temporary tables to the grid session for good measure.


john-hill (approved):
Seems reasonable to me.

PR #5410, opened 2025-10-09
PLFM-9187: Job to get count/size before download list add

marcomarasca:


marcomarasca:
Depends on https://github.com/Sage-Bionetworks/Synapse-Stack-Builder/pull/781

lib/jdomodels/src/main/resources/sql/GetDatasetFileStats.sql
```diff
@@ -0,0 +1,21 @@
+WITH ALL_FILES AS (
+    SELECT D_FILES.OWNER_NODE_ID AS ID, FILES.CONTENT_SIZE AS SIZE
+    FROM NODE_REVISION AS D
+    -- Unpack the JSON array of file references from each dataset
+    JOIN JSON_TABLE(
+        D.items,
+        '$[*]' COLUMNS (
+            id VARCHAR(255) PATH '$.entityId',
```
nickgros, line 8: This could probably be shorter, 25? (18446744073709551615 is the max long value, + 3 for 'syn'). Only if this is a memory concern on the database machine

nickgros (approved):

PR #5411, opened 2025-10-10
PLFM-9253

brucehoff:
Fixes the deployment build by changing the home directory to `root.

Removes some repeated lines of code.

Since we lose the Jenkins 'build number' in Code Pipeline (that Jenkins appends the date stamp) extends the date stamp to include the time (hours, minutes).

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -81,7 +81,7 @@ jobs:
           fi
           # for builds in Sage-Bionetworks repo', publish build artifacts
           if [[ $GITHUB_REPOSITORY_OWNER == "brucehoff" ]]; then
-            if [[ $branch == "develop" ]]; then
+            if [[ $branch == "develop" || "true" ]]; then
               GIT_COMMIT=${{ github.sha }}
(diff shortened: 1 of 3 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -83,7 +83,7 @@ jobs:
           if [[ $GITHUB_REPOSITORY_OWNER == "Sage-Bionetworks" ]]; then
             if [[ $branch == "develop" ]]; then
               GIT_COMMIT=${{ github.sha }}
-              ARTIFACT_VERSION=$(date +%Y-%m-%d)"-"${GIT_COMMIT:0:7}
+              ARTIFACT_VERSION=$(date +%Y-%m-%d_%H-%M)"_"${GIT_COMMIT:0:7}
```
marcomarasca, line 86: Maybe `github.run_number` can be a good substitute (https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context)

marcomarasca (approved):
Loogs good, we could use the gitub.run_number in place of the jenkins build number.

PR #5413, opened 2025-10-10
PLFM-9217

brucehoff:
Allow a user to specify which of their registered email addresses a password reset email should go to.

### DIFF
--- a/services/repository-managers/src/main/java/org/sagebionetworks/repo/service/auth/AuthenticationServiceImpl.java
+++ b/services/repository-managers/src/main/java/org/sagebionetworks/repo/service/auth/AuthenticationServiceImpl.java
@@ -142,7 +142,7 @@ public class AuthenticationServiceImpl implements AuthenticationService {
 		try {
 			PrincipalAlias principalAlias = userManager.lookupUserByUsernameOrEmail(usernameOrEmail);
 			PasswordResetSignedToken passwordRestToken = authManager.createPasswordResetToken(principalAlias.getPrincipalId());
-			messageManager.sendNewPasswordResetEmail(passwordResetUrlPrefix, passwordRestToken, principalAlias);
+			messageManager.sendNewPasswordResetEmail(passwordResetUrlPrefix, passwordRestToken, principalAlias, usernameOrEmail);
 		} catch (NotFoundException e) {
 			// should not indicate that a email/user could not be found
 		}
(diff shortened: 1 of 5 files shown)


services/repository-managers/src/main/java/org/sagebionetworks/repo/service/auth/AuthenticationServiceImpl.java
```diff
@@ -142,7 +142,7 @@ public void sendPasswordResetEmail(String passwordResetUrlPrefix, String usernam
 		try {
 			PrincipalAlias principalAlias = userManager.lookupUserByUsernameOrEmail(usernameOrEmail);
 			PasswordResetSignedToken passwordRestToken = authManager.createPasswordResetToken(principalAlias.getPrincipalId());
-			messageManager.sendNewPasswordResetEmail(passwordResetUrlPrefix, passwordRestToken, principalAlias);
+			messageManager.sendNewPasswordResetEmail(passwordResetUrlPrefix, passwordRestToken, principalAlias, usernameOrEmail);
```
marcomarasca, line 145: I do not think this change is necessary if I understand how the current implementation works:

1. The service looks up a PrincipalAlias that matches the username or email, notice that the as far as I can tell the usernameOrEmail in input can be any email associated with the account.
2. The MessageManagerImpl gets the actual email given the PrincipalAlias, See https://github.com/Sage-Bionetworks/Synapse-Repository-Services/blob/81f474b17a1e881d03bd11de26a7285dee17d9a5/services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/MessageManagerImpl.java#L586
**The email IS the alias value** if the PrincipalAlias type is USER_EMAIL (See https://github.com/Sage-Bionetworks/Synapse-Repository-Services/blob/81f474b17a1e881d03bd11de26a7285dee17d9a5/services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/MessageManagerImpl.java#L699)
3. Sends the email if is not quarantineed

marcomarasca (changes requested):
I believe [PLFM-9217](https://sagebionetworks.jira.com/browse/PLFM-9217) is already supported and nothing needs to be changed. I just tried resetting my password in synapse using a secondary email and it works?

services/repository-managers/src/main/java/org/sagebionetworks/repo/service/auth/AuthenticationServiceImpl.java
brucehoff: Yes, you are right!

PR #5414, opened 2025-10-13
PLFM-9283

brucehoff:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/oauth/ArcusBioProvider.java
```diff
@@ -0,0 +1,73 @@
...
+import org.sagebionetworks.util.ValidateArgument;
+import org.scribe.exceptions.OAuthException;
+import org.scribe.model.OAuthConfig;
+import org.scribe.model.Verifier;
+
+/**
+ * OAuthProvider for Arcus Bio's OpenID Connect server
+ * 
+ */
+public class ArcusBioProvider implements OAuthProviderBinding {
```
marcomarasca, line 14: Could you add a unit test for the provider?

marcomarasca (approved):
Approved, but the build is failing because the new secret names in the dev account need the "dev." prefix:

org.sagebionetworks.oauth2.arcusbio.client.secret -> dev.org.sagebionetworks.oauth2.arcusbio.client.secret

I added the new dev...client.id and dev..cllient.secret in the dev account (as my builds were failing too), so you should be able to run a build now.

PR #5415, opened 2025-10-14
PLFM-9253

brucehoff:
This PR changes the build behavior:
- to build and push to Artifactory when a 'develop' or 'release-*' branch is pushed to the `Sage-Bionetworks` repo';
- to add an explicit step "Display Artifact Version" which shows the version, for convenience;
- to fix the problem with authenticating to Artifactory by (1) removing the unneeded Docker `--user` override for the CodePipeline build and, (2) correctly setting the home directory in the Docker container so that the `settings.xml` file is found. (Previously the dir was not being set.)

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -15,11 +15,7 @@ name: main
 
 on:
   push:
-    branches:
-      - '*'
-    tags:
-      - '[0-9]+'
-      - '[0-9]+\.[0-9]'
+    branches: ['*']
 
 # let only one copy of the workflow run at a time
 concurrency:
@@ -76,9 +72,9 @@ jobs:
             if [[ $branch == "develop" ]]; then
               GIT_COMMIT=${{ github.sha }}
               ARTIFACT_VERSION=$(date +%Y-%m-%d_%H-%M)"_"${GIT_COMMIT:0:7}
-            fi
-            if [[ ${GITHUB_REF:0:10} == "refs/tags/" ]]; then
-              ARTIFACT_VERSION="${GITHUB_REF#refs/tags/}"
+            #elif [[ $branch =~ "release-" ]]; then
+            elif [[ $branch =~ "PLFM-9253" ]]; then
+              ARTIFACT_VERSION=$(git describe --tags)
             fi
           fi
           if [[ $ARTIFACT_VERSION ]]; then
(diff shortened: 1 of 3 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -15,7 +15,11 @@ name: main
 
 on:
   push:
-    branches: ['*']
+    branches:
+      - '*'
+    tags:
```
xschildw: I think we should just build on pushing the branch.
Note: I create the new branch with 'git push origin release-$3 --tags', does it
mean that the build could be triggered before the xxx.0 tag is created?

.github/workflows/build_and_test.yml
```diff
@@ -78,15 +75,22 @@ jobs:
             if [[ $branch == "develop" ]]; then
               GIT_COMMIT=${{ github.sha }}
               ARTIFACT_VERSION=$(date +%Y-%m-%d_%H-%M)"_"${GIT_COMMIT:0:7}
-            elif [[ $branch =~ "release-" ]]; then
-              ARTIFACT_VERSION=${{ steps.find_tag.outputs.previous_version }}
+            fi
+            if [[ ${GITHUB_REF:0:10} == "refs/tags/" ]]; then
```
xschildw: I'd go back to the previous version and look for 'release-*' branch. What was the problem with that implementation?

.github/workflows/build_and_test.yml
brucehoff: I did some testing.  What I learned is, when you `git push origin <branch> --tags`

- if there are no code changes then only the tag-push event is registered and only one run starts;
- if there is a code change then both the tag-push and branch-push events are registered and two runs start.

So if we want to guarantee only one run kicks off then a different approach is needed.  We should remove the `on tags` trigger.

.github/workflows/build_and_test.yml
brucehoff: The previous approach failed to find the right tag.  We'll need to fix it so it correctly finds the tag.

.github/workflows/build_and_test.yml
brucehoff: @xschildw It's now fixed to work as you prefer.  Please re-review.

.github/workflows/build_and_test.yml
brucehoff: @xschildw It's now fixed to work as you prefer.  Please re-review.

xschildw (approved):
Looks good!

PR #5426, opened 2025-10-28
added tests to reproduce the error and fixed the issue

john-hill:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/grid/internal/replica/view/query/filter/CellValueFilterElement.java
```diff
@@ -26,41 +26,67 @@ public CellValueFilterElement(Filter filter) {
...
+		return this;
+	}
 
 	@Override
 	public void toSql(StringBuilder sqlBuilder, Map<String, Object> params, Context context) {
 		validateInputs(sqlBuilder, params, context);
-		
+
 		String bind = "val" + params.size();
 		Integer columnIndex = context.getColumnIndexForName(columnName);
```
marcomarasca, line 76: I believe this to be the wrong index. The "VALS" array includes all the columns ordered according to the defined column order. I checked the getColumnIndexForName and it returns the actual value of the "vectorIndex" instead that might be different if the columns are re-ordered.

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/grid/internal/replica/view/query/filter/CellValueFilterElementTest.java
```diff
@@ -0,0 +1,61 @@
...
+import org.sagebionetworks.repo.model.grid.query.CellValueOperator;
+
+public class CellValueFilterElementTest {
+
+	/**
+	 * This is case that failed for PLFM-9309.
+	 */
+	@Test
+	public void testTransalteWithNullValue() {
+		CellValueFilter toClone = new CellValueFilter().setColumnName("c1").setOperator(CellValueOperator.IS_NOT_NULL);
```
marcomarasca, line 19: I think the operator IS_NOT_NULL might be confusing. Maybe it is better to have a dedicated filter instead? Or support c1 EQUALS null?

marcomarasca (approved):

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/grid/internal/replica/view/query/filter/CellValueFilterElement.java
nickgros: @marcomarasca to create a new Jira

nickgros (approved):

PR #5427, opened 2025-10-28
PLFM-9297: Add ability for the agent to select a single column by name

marcomarasca:


marcomarasca:
Depends on https://github.com/Sage-Bionetworks/Synapse-Stack-Builder/pull/793

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/grid/internal/replica/view/query/QueryElement.java
```diff
@@ -28,27 +28,47 @@ public QueryElement() {
...
-		for (SelectItemElement item : select) {
-			if (!isFirstSelect) {
-				sqlBuilder.append(" ,");
+		
+		if (!isAggregate()) {
+			// For non-aggregate queries, always include all metadata columns
+			sqlBuilder.append(" AN_REP, AN_SEQ, `INDEX`, RO_REP, RO_SEQ, VEC_REP, VEC_SEQ,");
+			sqlBuilder.append(" MO_REP, MO_SEQ, SRC_REP, SRC_SEQ, RVC_REP, RVC_SEQ, VAL_RES, SYN_ROW,");
+			
+			// Build a SELECTED_VALS array from the selected columns extracted directly from the CTE VALS array
```
john-hill: If the aggregate cases is treated the same (JSON_ARRAY(count(*))) then the only difference here is that we include more columns for the non-aggregate case.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/grid/internal/replica/view/query/Context.java
```diff
@@ -15,10 +16,31 @@ public GridHeader getHeader() {
...
 
-	public Integer getColumnIndexForName(String columnName) {
+	/**
+	 * 
+	 * @param columnName
+	 * @return The {@link Column#getVectorIndex() vector index} for the given column name.
+	 * @throws IllegalArgumentException if the column name is not found.
+	 */
+	public Integer getColumnVectorIndexForName(String columnName) {
 		return header.getOrderedColumns().stream().filter(c -> c.getName().equals(columnName))
```
john-hill: we may not have a use case for this anymore?

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/grid/internal/replica/view/query/select/SelectColumnElement.java
```diff
@@ -0,0 +1,39 @@
...
+	public boolean isAggregate() {
+		return false;
+	}
+
+	@Override
+	public void setSelect(Context context, Long index, List<org.sagebionetworks.repo.model.grid.query.result.SelectColumn> selectColumns) {
+		Integer vectorIndex = context.getColumnVectorIndexForName(columnName);		
+		selectColumns.add(new SelectColumn()
+			.setColumnName(columnName)
+			.setColumnIndex(Long.valueOf(vectorIndex))
```
john-hill: it seems the index is redundant as it will always match the position of its SelectColumns in the resulting List.

john-hill (changes requested):

nickgros (approved):

PR #5421, opened 2025-10-22
PLFM-9300 - Grid validation handles optional values

nickgros:
- Update constant value DB and DTO representations to discern between `null` and `undefined`
- Extend grid-index-view-template.sql to generate a JSON object document for each row in the grid, which properly includes `null` values and omits `undefined` values. JSON document is used for validation
- Update PatchRowHandler to consider required columns when creating empty values. If a column is `required`, then cells without data will be instantiated with `null` values. Otherwise, the cells will be instantiated with an `undefined` value, which is the appropriate value for an optional property.

PR #5436, opened 2025-11-04
PLFM-9273: Fix LWW insertion to avoid circular references

marcomarasca:


lib/lib-grid/src/test/java/org/sagebionetworks/repo/model/grid/node/VectorNodeTest.java
```diff
@@ -127,6 +127,18 @@ public void testAttemptInsertWithNewIndex() {
 				.setValueFromJson("{\"c0\":{\"v\":111,\"i\":[3,4]},\"c1\":{\"v\":222,\"i\":[4,4]}}");
 		assertEquals(expected, vec);
 	}
+	
+	@Test
+	public void testAttemptInsertWithNewNodeIdLesserThanContainerId() {
+		VectorNode vec = new VectorNode().setId(id).setValueFromJson("{\"c0\":{\"v\":111,\"i\":[3,4]}}");
```
nickgros, line 133: These JSON strings will need to change with #5421 (depending on which goes in first)

nickgros (approved):

PR #5438, opened 2025-11-07
PLFM-9342

nickgros:
Always update the validation result, even if it did not change. The clock of the validation result node can be used to determine if the validation result is up-to-date compared to the nodes in a row's data vector.


john-hill (approved):
Looks good to me.

PR #5439, opened 2025-11-07
PLFM-9283

brucehoff:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/oauth/ArcusBioProvider.java
```diff
@@ -50,12 +63,18 @@ public ProvidedUserInfo validateUserWithProvider(String authorizationCode, Strin
...
-			OAuth2Service service = (new OAuth2Api(authUrl, tokenUrl)).
+			OAuth2Service service = (new OAuth2Api(authUrl, tokenUrl, userInfoUrl)).
 					createService(new OAuthConfig(apiKey, apiSecret, redirectUrl, null, null, null));
 			
 			AccessTokenResponse accessTokenResponse = service.getAccessToken(null, new Verifier(authorizationCode));
 						
-			return accessTokenResponse.parseIdToken();
+			ProvidedUserInfo providedUserInfo =  accessTokenResponse.parseIdToken();
+			// if the IdP fails to provide the required claims in the ID Token, then
+			// a request is made to the /userinfo endooint
```
jay-hodgson: sp (endpoint)

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/oauth/OAuth2Api.java
```diff
@@ -88,6 +96,35 @@ public OAuth2Service createService(OAuthConfig config) {
...
+				info.setFirstName(json.getString(GIVEN_NAME));
+			}
+			if (json.has(SUB)) {
+				info.setSubject(json.getString(SUB));
+			}
+			if (json.has(EMAIL)) {
+				info.setEmail(json.getString(EMAIL));
+			}
+			if (json.has(EMAIL_VERIFIED)) {
+				info.setEmailVerified(json.getString(EMAIL_VERIFIED));
```
jay-hodgson, line 117: Should ProvidedUserInfo.emailVerified be a boolean, and use `json.getBoolean(EMAIL_VERIFIED)` here?

brucehoff: No, because i'm not sure what Okta puts into the `email_verified` string and I want to capture the raw data even if it isn't literally `true` or `false`.  If it's an unexpected value we will return it in the error message added in the previous PR which in turn will help our debugging.

jay-hodgson: Ok.  So if the values is not a true or false, then the `json.getBoolean(EMAIL_VERIFIED)` on line 119 will cause a JSONException (which will be wrapped in an UnauthorizedException)

jay-hodgson (approved):

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/oauth/OAuth2Api.java
brucehoff: > Ok. So if the values is not a true or false, then the `json.getBoolean(EMAIL_VERIFIED)` on line 119 will causes a JSONException (which will be wrapped in an UnauthorizedException)

Good catch.  I updated the code and added a test.  Please re-review.

jay-hodgson (approved):

PR #5454, opened 2025-11-18
PLFM-9355 - log agent invocation events & failures

nickgros:
In PLFM-9355, we discovered that when invoking functions, the agent submits JSON, but deeply nested objects are mangled by Bedrock such that they become unparseable.

As a workaround, we will ask the agent to encode these nested objects as strings, which Bedrock will not transform, and our JSON parsing library incidentally permits and transforms to the un-escaped string object.

With this change, we can validate the effectiveness of our workaround by tracking frequency of this type of exception.


john-hill (approved):
Looks good to me.

PR #5455, opened 2025-11-18
PLFM-9355 - Modify GridUpdateRequest to have a single top-level `update` object property.

nickgros:
Additionally, update the example generation code to escape requests in a way that works around Bedrock's limitations.


nickgros:
Passing build with required stack builder changes here: https://build-system-synapse.dev.sagebase.org/job/devpnickgros5/79/

john-hill (approved):
Looks good to me!

PR #5474, opened 2026-01-08
PLFM-9323

brucehoff:
Add a Realm object which has an id, a name and a list of identity providers (IdPs).
An IdP is either Synapse or it is an OAuthProvider.
No two realms can share an identity provider (so when you authenticate via an IdP you are unambiguously authenticating to a realm).
Add a realm id field to UserGroup (FK to realm).
Add a PrincipalRealm relation for the default principals for each realm.  Each record has a realm ID, a principal ID, and a type (anonymous, authenticated, pubic, admin).

Admin services for creating and deleting realms. (Note, there is no update service at this time.)
Non-admin services to list the realm IDs, get a realm by its ID and get a realm's default principals by its ID.


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/auth/RealmDaoImpl.java
```diff
@@ -0,0 +1,321 @@
...
+					break;
+				default:
+					throw new IllegalStateException("Unexpected type" +principal.getPrincipalType());
+			}
+		}
+	}
+	
+	List<DBORealmPrincipal> copyRealmPrincipalsToDBOList(RealmPrincipal realmPrincipal) {
+		List<DBORealmPrincipal> result = new ArrayList<DBORealmPrincipal>();
+		if (null != realmPrincipal.getAnonymousUser()) {
```
SandhraSokhal: consider if null check is necessary

services/repository/src/main/java/org/sagebionetworks/repo/web/controller/UserGroupController.java
```diff
@@ -52,4 +57,39 @@ PaginatedResults<UserGroup> getUserGroups(HttpServletRequest request,
...
+	
+	/**
+	 * List the IDs of the current realms.
+	 * @return list of the IDs of the existing realms
+	 */
+	@RequiredScope({view})
+	@ResponseStatus(HttpStatus.OK)
+	@RequestMapping(value = UrlHelpers.REALM_LIST, method = RequestMethod.GET)
+	@ResponseBody
+	public RealmIdList listRealmIds() {
```
SandhraSokhal: Should we move Realm API's in RealmController?

SandhraSokhal (changes requested):

services/repository/src/main/java/org/sagebionetworks/repo/web/controller/RealmController.java
```diff
@@ -0,0 +1,64 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ControllerInfo(displayName="User & Group Services", path="repo/v1")
```
SandhraSokhal, line 22: "Realm services" instead of "User & Group Services"

SandhraSokhal (approved):
Looks good to me

services/repository/src/main/java/org/sagebionetworks/repo/web/controller/RealmController.java
brucehoff: Good catch, thanks.  I will fix in the next PR.

PR #5476, opened 2026-01-14
PLFM-9324

brucehoff:
This PR:

Adds a DAO-level query to get the realm ID for a given identity provider.

In UserManagerImpl.createUser(), sets the user's realm according to the identity provider used to authenticate.

Adds realmId to the UserInfo constructor and cleans up the various constructors by (1) removing one which is superfluous and (2) marking as 'deprecated' several others which would be laborious to remove since they are used in many, many tests.

In UserManagerImpl.getUserInfo(), sets PUBLIC, ANONYMOUS, AUTHENTICATED_USERS IDs according to the user's realm.

In AuthenticationServiceImpl.bindExternalID, disallows the binding if the user's existing realm is not associated with the identity provider which is the source of new identity attempting to be bound to the user.

Bonus:

Adds a service to get the realm principals based on user id. (The user id will come from the access token in the request.) This allows the client to be able to get the realm principals without needing to know the realm id explicitly.

### DIFF
--- a/client/synapseJavaClient/src/main/java/org/sagebionetworks/client/SynapseClient.java
+++ b/client/synapseJavaClient/src/main/java/org/sagebionetworks/client/SynapseClient.java
@@ -4632,10 +4632,10 @@ public interface SynapseClient extends BaseClient {
     
     RealmIdList listRealmIds() throws SynapseException ;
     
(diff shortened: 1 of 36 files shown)


SandhraSokhal (approved):
LGTM

PR #5481, opened 2026-01-20
PLFM-9398 - Node model refactor, CBOR utility

nickgros:
- Refactor node model; ArrayNode represents a list/container of (new) RGANodes
- Refactor VectorNode to use integer map keys instead of string to avoid parsing later on
- Add CBOR utility to encode/decode ConValue


lib/lib-grid/src/main/java/org/sagebionetworks/repo/model/grid/encoding/CBORUtils.java
```diff
@@ -0,0 +1,251 @@
...
+            }
+            pb.unread(firstByte); // Put it back for Jackson to read
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to read from stream", e);
+        }
+
+        // Use Jackson CBOR to decode the value
+        JsonNode jsonNode = parseJsonNode(pb);
+
+        if (jsonNode.isNull()) {
```
john-hill, line 152: Consider moving this to its own method.

lib/lib-grid/src/main/java/org/sagebionetworks/repo/model/grid/encoding/CBORUtils.java
```diff
@@ -0,0 +1,251 @@
...
+import com.fasterxml.jackson.dataformat.cbor.databind.CBORMapper;
+
+public class CBORUtils {
+
+    private static final CBORMapper CBOR_MAPPER = new CBORMapper();
+    private static final CBORFactory CBOR_FACTORY = CBOR_MAPPER.getFactory();
+
+    private static final int UNDEFINED_BYTE = 0xF7;
+
+    public static CBORMapper getCBORMapper() {
```
john-hill, line 29: need to confirm that this is thread safe and designed to be used as a static.

lib/lib-grid/src/test/java/org/sagebionetworks/repo/model/grid/encoding/CBORUtilsTest.java
```diff
@@ -0,0 +1,68 @@
...
+        LONG_NEGATIVE(ConType.LONG, -123L),
+        DOUBLE(ConType.DOUBLE, 1.25),
+        DOUBLE_ZERO(ConType.DOUBLE, 0.0),
+        BOOLEAN_TRUE(ConType.BOOLEAN, true),
+        BOOLEAN_FALSE(ConType.BOOLEAN, false),
+        ARRAY(ConType.JSON_ARRAY, new JSONArray("[1,2,3]")),
+        ARRAY_EMPTY(ConType.JSON_ARRAY, new JSONArray("[]")),
+        OBJECT(ConType.JSON_OBJECT, new JSONObject("{\"key\":99}")),
+        OBJECT_EMPTY(ConType.JSON_OBJECT, new JSONObject("{}")),
+        OBJECT_VARIOUS_INTERNAL_TYPES(ConType.JSON_OBJECT, new JSONObject("{\"str\": \"foo\", \"int\": 42, \"float\": 1.5, \"bool\": true, \"null\": null, \"arr\": [1,2], \"obj\": {\"a\": \"b\"}}"));
```
john-hill, line 40: nested for the array (array of objects).

john-hill (approved):

lib/lib-grid/src/main/java/org/sagebionetworks/repo/model/grid/encoding/CBORUtils.java
nickgros: These are thread-safe and intended to be created/configured once, since instantiating these is expensive.

The generator is not thread-safe, but is created/destroyed within a single operation, so this is being used correctly as far as I can tell (cross referenced with LLM agent).

john-hill (approved):

PR #5482, opened 2026-01-21
PLFM-9323

brucehoff:
three fixes:
- allow upper case letters in realm name
- set created on date in default realm principals
- set realm of admin' group correctly


SandhraSokhal (approved):
LGTM

PR #5484, opened 2026-01-22
PLFM-9326

brucehoff:
/login2 (log in with password) is only allowed if the authenticated user is in a realm that allows Synapse as the IdP
/oauth2/session2 is only allowed if the authenticated user is in a realm that allows the given oauth provider.

Also, this PR removes the realm backfill migration logic, introduced when the non-nullable realm id was added to the USER_GROUP table.


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/authentication/AuthenticationManagerImpl.java
```diff
@@ -215,13 +215,23 @@ public LoginResponse login(LoginRequest request, String tokenIssuer){
 			}
 		});
 		
+		UserInfo user = userManager.getUserInfo(userId);
+		// user must be in default Synapse reealm
+		if (!AuthorizationConstants.DEFAULT_REALM_ID.equals(user.getRealmId())) {
+			throw new UnauthorizedException("Cannot log in using a password.  Use the designated identity provider instead.");
+		}
+		
 		return loginWithNoPasswordCheck(userId, tokenIssuer);
```
SandhraSokhal, line 224: we should pass userInfo instead of userId

SandhraSokhal (approved):
LGTM

PR #5488, opened 2026-01-27
PLFM-9398 - Full model encoder and decoder

nickgros:


lib/lib-grid/src/main/java/org/sagebionetworks/repo/model/grid/encoding/IndexedModelEncoder.java
```diff
@@ -0,0 +1,133 @@
...
+
+	/**
+	 * Close the encoder, writing the complete CBOR map.
+	 * This must be called to produce valid output.
+	 *
+	 * @throws IOException if an I/O error occurs
+	 */
+	@Override
+	public void close() throws IOException {
+		if (!closed) {
```
john-hill, line 110: can remove the boolean and we need to close out.

lib/lib-grid/src/main/java/org/sagebionetworks/repo/model/grid/encoding/IndexedModelEncoder.java
```diff
@@ -0,0 +1,133 @@
...
+
+		// Update the clock table with the node.
+		clockTable.processNode(node);
+
+		// Generate the node key
+		String nodeKey = clockTable.encodeNodeKey(node.getId());
+
+		// Encode the node to binary
+		// Use a Piped stream to avoid buffering the entire node in memory
+		final PipedInputStream in = new PipedInputStream();
```
john-hill: streaming should give you free memory management ensuring only the latest nodes will be i memory.

lib/lib-grid/src/main/java/org/sagebionetworks/repo/model/grid/encoding/IndexedModelDecoder.java
```diff
@@ -0,0 +1,360 @@
...
+		} catch (IOException e) {
+			throw new UncheckedIOException("Failed to initialize parser for node streaming", e);
+		}
+
+		return new Iterator<>() {
+            private Node next = null;
+            private boolean done = false;
+
+            @Override
+            public boolean hasNext() {
```
john-hill, line 268: check /simplify

lib/lib-grid/src/main/java/org/sagebionetworks/repo/model/grid/encoding/IndexedModelDecoder.java
```diff
@@ -0,0 +1,360 @@
...
+        };
+	}
+
+	/**
+	 * Close the decoder.
+	 *
+	 * @throws IOException if an I/O error occurs
+	 */
+	@Override
+	public void close() throws IOException {
```
john-hill, line 312: ensure the parsers will close the input stream, and do you still need the boolean?

lib/lib-grid/src/main/java/org/sagebionetworks/repo/model/grid/ClockTable.java
```diff
@@ -32,6 +38,50 @@ public ClockTable(List<LogicalTimestamp> clocks) {
...
+    /**
+     * Process a node to update the clock table as needed.
+     */
+    public void processNode(Node node) {
+        ValidateArgument.required(node, "node");
+
+        updateClockTable(node.getId());
+
+        // For nodes that reference other nodes, also process those references
+        if (node instanceof ArrayNode) {
```
john-hill: can we change nodes such that you can iterate through the children with ids then this method just iterate, where each node type can handle the node specific logic?

lib/lib-grid/src/test/java/org/sagebionetworks/repo/model/grid/encoding/IndexedModelDecoderTest.java
```diff
@@ -0,0 +1,286 @@
...
+			assertEquals(expected, nodes.get(0));
+		}
+	}
+
+	@Test
+	public void testDecodeMultipleReplicas() throws IOException {
+		try (IndexedModelDecoder decoder = new IndexedModelDecoder(loadResource("/indexed-model/multiple-replicas.cbor"))) {
+			// Verify clock table
+			ClockTable clockTable = decoder.getClockTable();
+			assertNotNull(clockTable);
```
john-hill, line 72: was expecting two clock entries with multiple replicas

lib/lib-grid/src/test/java/org/sagebionetworks/repo/model/grid/encoding/IndexedModelEncoderTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+		try (IndexedModelEncoder encoder = new IndexedModelEncoder(out, rootNodeId)) {
+			for (Node node : nodes) {
+				encoder.writeNode(node);
+			}
+		}
+
+		byte[] result = out.toByteArray();
+		assertNotNull(result);
+		assertTrue(result.length > 0);
```
john-hill, line 99: can we test with the CBOR reader that this at least contains the expected elements?

lib/lib-grid/src/test/java/org/sagebionetworks/repo/model/grid/encoding/IndexedModelEncoderTest.java
```diff
@@ -0,0 +1,108 @@
...
+		ByteArrayOutputStream out = new ByteArrayOutputStream();
+		IndexedModelEncoder encoder = new IndexedModelEncoder(out, rootNodeId);
+		encoder.close();
+
+		ConstantNode node = new ConstantNode()
+			.setId(new LogicalTimestamp().setReplicaId(100L).setSequenceNumber(2L))
+			.setValue(new ConValue(ConType.LONG, 42L));
+
+		assertThrows(IllegalStateException.class, () -> {
+			encoder.writeNode(node);
```
john-hill, line 173: Is there a way we can ensure that both the passed input and output streams are closed?

lib/lib-grid/src/test/java/org/sagebionetworks/repo/model/grid/encoding/IndexedModelEncodingRoundTripTest.java
```diff
@@ -0,0 +1,132 @@
...
+                )));
+        nodes.add(new ConstantNode()
+                .setId(new LogicalTimestamp().setReplicaId(100L).setSequenceNumber(12L))
+                .setValue(new ConValue(ConType.BOOLEAN, true)));
+        nodes.add(new ConstantNode()
+                .setId(new LogicalTimestamp().setReplicaId(200L).setSequenceNumber(13L))
+                .setValue(new ConValue(ConType.NULL, null)));
+
+        // Write to test resources directory
+        Path outputFile = Files.createTempFile("indexed-model-", ".cbor");
```
john-hill: I would use in memory byte[]

john-hill (changes requested):
Looking good!

lib/lib-grid/src/main/java/org/sagebionetworks/repo/model/grid/encoding/IndexedModelEncoder.java
nickgros: The `close` method writes CBOR data to the stream. If we do not keep the boolean, additional calls to `close` will throw an error.

Put another way, the boolean allows `close` to be idempotent.

lib/lib-grid/src/main/java/org/sagebionetworks/repo/model/grid/encoding/IndexedModelDecoder.java
nickgros: This cannot be simplified on its own, but if we added guava as a dependency to lib-grid we could use their `AbstractIterator` class instead:

```java
return new AbstractIterator<Node>() {
      @Override
      protected Node computeNext() {
          try {
              Node node = readNode();
              return node != null ? node : endOfData();
          } catch (IOException e) {
              throw new UncheckedIOException("Failed to read node", e);
          }
      }
  };

```

lib/lib-grid/src/main/java/org/sagebionetworks/repo/model/grid/encoding/IndexedModelDecoder.java
nickgros: Boolean can be removed, Jackson's parser will automatically close the stream unless configured not to

I chose not to test that the stream is closed because
- The InputStream API does not provide a way to check if a stream has been closed
- It would be complicated to mock an input stream and verify that the CBORGenerator closes it, because the mock would need to return the correct bytes

lib/lib-grid/src/test/java/org/sagebionetworks/repo/model/grid/encoding/IndexedModelEncoderTest.java
nickgros: Added `testOutputStreamIsClosedOnClose`

lib/lib-grid/src/test/java/org/sagebionetworks/repo/model/grid/encoding/IndexedModelEncoderTest.java
nickgros: Extended tests to parse the output with the Jackson CBOR parser. Now verifying that we see the expected fields in the CBOR map

lib/lib-grid/src/test/java/org/sagebionetworks/repo/model/grid/encoding/IndexedModelDecoderTest.java
nickgros: Updated encoded file and test

lib/lib-grid/src/main/java/org/sagebionetworks/repo/model/grid/ClockTable.java
nickgros: Added to Node interface:

```java
	/**
	 * Returns a stream of all LogicalTimestamps referenced by this node, including
	 * the node's own ID and any timestamps contained within the node's structure.
	 * <p>
	 * The stream includes the node's own ID first, followed by any referenced timestamps.
	 * Null values are filtered out.
	 *
	 * @return a stream of all referenced LogicalTimestamps
	 */
	Stream<LogicalTimestamp> streamReferencedTimestamps();

```

lib/lib-grid/src/main/java/org/sagebionetworks/repo/model/grid/encoding/IndexedModelEncoder.java
nickgros: Updated so `nodeCodec.encode` writes to an in-memory `ByteArrayOutputStream`, and the contents can be passed to the CBOR generator with `ByteArrayOutputStream.toByteArray`

john-hill (approved):

PR #5490, opened 2026-01-28
PLFM-9353

brucehoff:
Update references to default principals (anonymous, authenticated users, and public) to instead refer to the default principals of the user's realm.

To review:
- Start with UserInfo and UserManager.getUserInfo() to see how we get the default realm principals for the user.
- Note the new method RealmDao.getRealmForAnonymousPrincipal() which checks whether a principal is anonymous and, if so, returns the realm id.
- AuthorizationUtils.isUserAnonymous() is preserved as AuthorizationUtils.isDefaultRealmAnonymousId() for the few remaining cases where we actually have to compare against that particular principal ID.
- There are a few places where we have to reject the anonymous user, e.g., when someone tries to put it in an ACL.  We update the check to reject the anonymous user for the user's realm.  We further need to reject *all* anonymous users, for *all* realms, but we delegate that check to the forthcoming updates to ACL management, which will ensure that all ACL entries are for the same realm.
- In many places, we update `AuthorizationUtils.isUserAnonymous(userId)` to `userInfo.isUserAnonymous()` and the surrounding code is updated to pass along the `userInfo` object instead of just the id.


lib/models/src/main/java/org/sagebionetworks/repo/model/UserInfo.java
```diff
@@ -47,6 +50,34 @@ public UserInfo(boolean isAdmin, Long id, String realmId){
 		this.groups.add(this.id);
 		this.realmId=realmId;
 	}
+	
+	public boolean isUserAnonymous() {
+		return id==null || id.equals(realmAnonymousUserId);
```
SandhraSokhal, line 55: can we make this non null

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/UserManagerImpl.java
SandhraSokhal, line 221: can we use one database call to get all information

services/repository/src/main/java/org/sagebionetworks/repo/web/filter/throttle/RequestThrottleFilter.java
```diff
@@ -34,7 +34,7 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
 		HttpRequestIdentifier httpRequestIdentifier = HttpRequestIdentifierUtils.getRequestIdentifier(request);
 
 		//TODO: remove exception for anonymous users once java client has a way to get session id from cookies
-		if ( isMigrationAdmin(httpRequestIdentifier.getUserId()) || AuthorizationUtils.isUserAnonymous(httpRequestIdentifier.getUserId())) { //do not throttle the admin responsible for migration.
+		if (isMigrationAdmin(httpRequestIdentifier.getUserId()) || AuthorizationUtils.isDefaultRealmAnonymousId(httpRequestIdentifier.getUserId())) { //do not throttle the admin responsible for migration.
```
SandhraSokhal, line 37: can we check for all the Anonymous user from all realms in one database all using caching.

services/repository/src/main/java/org/sagebionetworks/repo/web/OAuthScopeInterceptor.java
```diff
@@ -55,11 +59,13 @@ public static boolean hasUserIdParameterOrAccessTokenHeader(HandlerMethod handle
...
+	public static boolean isUnAuthenticated(HttpServletRequest request) {
+		// if a request is unauthenticated then AuthenticationFilter will fill in
+		// user-id with the id of the default anonymous user
 		String userIdRequestParameter = request.getParameter(AuthorizationConstants.USER_ID_PARAM);
-		return userIdRequestParameter == null ||
-				AuthorizationConstants.BOOTSTRAP_PRINCIPAL.ANONYMOUS_USER.getPrincipalId()
-					.equals(Long.parseLong(userIdRequestParameter));
+
+		return userIdRequestParameter==null ||
+			AuthorizationUtils.isDefaultRealmAnonymousId(Long.parseLong(userIdRequestParameter));
```
SandhraSokhal, line 68: Can we add parameter(boolean) for Anonymous user from upstream so we have all the information.

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/auth/RealmDaoImpl.java
```diff
@@ -81,6 +83,11 @@ public class RealmDaoImpl implements RealmDao {
 	
 	private static final String DELETE_REALM_SQL = "DELETE FROM "+TABLE_REALM+" WHERE "+COL_REALM_ID+" = ?";
 	
+	private static final String SELECT_REALM_FOR_ANONYNOUS_PRINCIPAL_SQL = 
+			"SELECT "+COL_REALM_PRINCIPAL_REALM_ID+" FROM "+TABLE_REALM_PRINCIPAL+
+			" WHERE "+
+			COL_REALM_PRINCIPAL_PRINCIPAL_ID+" = "+"? AND "+
```
SandhraSokhal, line 89: Add a method to get all principals of all realms at once

services/repository/src/main/java/org/sagebionetworks/auth/filter/AcceptTermsOfUseFilter.java
SandhraSokhal, line 50: can we add another request parameter for Anonymous user

services/repository/src/main/java/org/sagebionetworks/auth/filter/TwoFactorAuthRequiredFilter.java
```diff
@@ -41,7 +45,8 @@ protected void doFilterInternal(HttpServletRequest httpRequest, HttpServletRespo
 		Long userId = Long.parseLong(userIdParam);
 		
 		// The anonymous user is not a conventional user and cannot enable 2fa
-		if (BOOTSTRAP_PRINCIPAL.ANONYMOUS_USER.getPrincipalId().equals(userId)) {
+		// we check to see if the given user is the anonymous user in any realm
+		if (realmDao.getRealmForAnonymousPrincipal(userId.toString()).isPresent()) {
```
SandhraSokhal, line 49: we can pass here boolean for Anonymous

SandhraSokhal (approved):
LGTM

PR #5499, opened 2026-02-06
PLFM-9447: Change inactive days from 180 to 370

xschildw:
This PR changes the limit for defining an inactive account from 180 to 370 days.


john-hill (approved):
Feel free to merge when you are ready.

PR #5501, opened 2026-02-07
PLFM-9353

brucehoff:
- Add an 'anonymous' request parameter in the AuthenticationFilter which downstream filters can use, rather than issuing their own DB query to find out if the user is anonymous;
- Remove AuthorizationUtils.isDefaultRealmAnonymousId(), no longer needed;
- Although we discussed making a (cached) query to get anonymous principals for all realms, this did not end up being necessary

### DIFF
--- a/services/repository/src/main/java/org/sagebionetworks/auth/HttpAuthUtil.java
+++ b/services/repository/src/main/java/org/sagebionetworks/auth/HttpAuthUtil.java
@@ -198,5 +198,10 @@ public class HttpAuthUtil {
 		}
 		reject(resp, er, status);
 	}
+	
+	public static boolean isAnonymous(HttpServletRequest httpRequest) {
+		String param = httpRequest.getParameter(AuthorizationConstants.ANONYMOUS_PARAM);
+		return Boolean.parseBoolean(param);
+	}
 
 }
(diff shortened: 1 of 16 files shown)


services/repository/src/test/java/org/sagebionetworks/auth/HttpAuthUtilTest.java
```diff
@@ -274,5 +274,16 @@ void testRejectUnauthorized() throws Exception {
...
+	@Test
+	void testIsAnonymousTrue() throws Exception {
+		when(httpRequest.getParameter("anonymous")).thenReturn("true");
+		assertTrue(HttpAuthUtil.isAnonymous(httpRequest));
+	}	
+	@Test
+	void testIsAnonymousFalse() throws Exception {
+		when(httpRequest.getParameter("anonymous")).thenReturn("false");
+		assertFalse(HttpAuthUtil.isAnonymous(httpRequest));
+	}
```
john-hill, line 287: test for null.

services/repository/src/main/java/org/sagebionetworks/auth/HttpAuthUtil.java
```diff
@@ -198,5 +198,10 @@ public static void rejectWithOAuthError(HttpServletResponse resp, OAuthErrorCode
 		}
 		reject(resp, er, status);
 	}
+	
+	public static boolean isAnonymous(HttpServletRequest httpRequest) {
```
john-hill, line 202: will throw if not there!

john-hill (approved):

PR #5514, opened 2026-02-16
PLFM-9471 - RFC-7662 introspection service

nickgros:


lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/oauth/OAuthTokenIntrospectionResponse.json
```diff
@@ -0,0 +1,47 @@
...
+			"type": "boolean",
+			"description": "Whether the token is active"
+		},
+		"scope": {
+			"type": "array",
+			"items": {
+				"$ref": "org.sagebionetworks.repo.model.oauth.OAuthScope"
+			},
+			"description": "Scopes granted by the token"
+		},
```
nickgros, line 11: Not following spec (return an array of scopes rather than space-separated string), is that ok?

lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/oauth/OAuthTokenIntrospectionResponse.json
```diff
@@ -0,0 +1,47 @@
...
+		"aud": {
+			"type": "string",
+			"description": "The audience (realm or OAuth Client) the token was issued for"
+		},
+		"sub": {
+			"type": "string",
+			"description": "Subject identifier (Pairwise Pseudonymous Identifier)"
```
nickgros: Since this is only for accessToken/PAT, is it true that this is always the user ID, not a PPID?

lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/oauth/OAuthTokenIntrospectionRequest.json
```diff
@@ -0,0 +1,13 @@
+{
+    "description": "Request to introspect an OAuth 2.0 access token, loosely based on <a href=\"https://tools.ietf.org/html/rfc7662\">RFC 7662</a>.",
```
brucehoff: I recall the spec' suggesting a token type hint.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/oauth/OpenIDConnectManagerImpl.java
```diff
@@ -692,4 +692,93 @@ public void revokeUserAccess(Long userId) {
...
+	@Override
+	public OAuthTokenIntrospectionResponse introspectToken(String token, Long maxAge) {
+		ValidateArgument.required(token, "token");
+
+		// Parse the JWT. If parsing fails (expired, invalid signature, etc.), the token is inactive.
+		Jwt<JwsHeader, Claims> jwt;
+		try {
+			jwt = oidcTokenManager.parseJWT(token);
+		} catch (Exception e) {
+			return inactiveIntrospectionResponse();
```
brucehoff, line 709: does this correctly differentiate an invalid from a valid but expired token?

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/oauth/OpenIDConnectManagerImpl.java
```diff
@@ -692,4 +692,93 @@ public void revokeUserAccess(Long userId) {
...
+		}
+
+		Claims claims = jwt.getBody();
+
+		// Check token type — only OIDC_ACCESS_TOKEN and PERSONAL_ACCESS_TOKEN are valid
+		TokenType tokenType;
+		try {
+			tokenType = TokenType.valueOf(claims.get(OIDCClaimName.token_type.name(), String.class));
+		} catch (Exception e) {
+			return inactiveIntrospectionResponse();
```
brucehoff, line 719: should we differentiate between invalid and inactive tokens? Maybe have different responses for different kinds of Exception?

brucehoff:
You might consider adding the `introspection_endpoint` to the OIDC Configuration/Discovery document.  See https://github.com/Sage-Bionetworks/Synapse-Repository-Services/blob/0eb38b1352b7e11366a3b71b44224a2a8d4ebf80/services/repository-managers/src/main/java/org/sagebionetworks/repo/service/auth/OpenIDConnectServiceImpl.java#L147

lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/oauth/OAuthTokenIntrospectionResponse.json
brucehoff: I would suggest following the spec'.  The reason I say this is because we have other circumstances where a tiny deviation caused an off-the-shelf client to break.  Two examples:
- We added a leading zero to the numbers in our public key (publishing signed rather than unsigned integers). This broke a microsoft OIDC client;
- We returned a 201 status when the spec' said to return a 200 status, which caused another off-the-shelf client to fail.

lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/oauth/OAuthTokenIntrospectionResponse.json
brucehoff: > Since this is only for accessToken/PAT

An access token can come from an OIDC client, which receieves a PPID 'sub' claim from Synapse.  I assume that this 'sub' is the same as the 'sub' in the introspected token and hence is obfuscated here just as it is in the token.

client/synapseJavaClient/src/main/java/org/sagebionetworks/client/SynapseClientImpl.java
```diff
@@ -4822,6 +4825,12 @@ public void revokeTokenURLEncoded(String token) throws SynapseException, Unsuppo
 		ClientUtils.checkStatusCodeAndThrowException(response);
 	}
 
+	@Override
+	public OAuthTokenIntrospectionResponse introspectToken(OAuthTokenIntrospectionRequest request) throws SynapseException {
```
brucehoff, line 4829: Please test that our service works according to the spec':

> The protected resource calls the introspection endpoint using an HTTP
   POST [RFC7231] request with parameters sent as
   "application/x-www-form-urlencoded" data

Note that we already have a Spring HHttpMessageConverter which may do the work with no additional coding needed: https://github.com/Sage-Bionetworks/Synapse-Repository-Services/blob/0eb38b1352b7e11366a3b71b44224a2a8d4ebf80/services/repository/src/main/java/org/sagebionetworks/repo/web/controller/FormEncodedMessageConverter.java#L26

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/oauth/OpenIDConnectManagerImpl.java
```diff
@@ -692,4 +692,93 @@ public void revokeUserAccess(Long userId) {
...
+		}
+		if (claims.getIssuedAt() != null) {
+			response.setIat(claims.getIssuedAt().getTime() / 1000L);
+		}
+		if (authTimeSeconds != null) {
+			response.setAuth_time(authTimeSeconds);
+		}
+
+		// Extract scopes
+		response.setScope(ClaimsJsonUtil.getScopeFromClaims(claims));
```
brucehoff: We could also add the claims from the access token, if any.  These claims are the fine-grained list of attibutes that the access token allow the client to retrieve from the `user_info` endpoint.

brucehoff (changes requested):

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/oauth/OpenIDConnectManagerImpl.java
nickgros: The specification suggests that we do _not_ distinguish between the two.

> Note that a properly formed and authorized query for an inactive or
>   otherwise invalid token (or a token the protected resource is not
>   allowed to know about) is not considered an error response by this
>   specification.  In these cases, the authorization server MUST instead
>   respond with an introspection response with the "active" field set to
>   "false" as described in [Section 2.2](https://datatracker.ietf.org/doc/html/rfc7662#section-2.2).

Since the access token is technically opaque per the specification, we should not distinguish between these cases, even though we technically can.

nickgros: Note that we will throw a 400 if the token is `null`, so we will not consider that scenario to be a "properly formed query"

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/oauth/OpenIDConnectManagerImpl.java
nickgros: See response above

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/oauth/OpenIDConnectManagerImpl.java
nickgros: I am opting against including this to keep the response object simple. It is optional per the spec, and not particularly useful since this information is already embedded in the token itself.

brucehoff (approved):
We may have to add 'claims' later but I don't min deferring that for now.

PR #5523, opened 2026-02-23
PLFM-9477

nickgros:
- Fix issue where CSV importer would fail to insert rows if no insert exists

Changed GridIndexDao#getRgaLastNode to GridIndexDao#getArrayLastNodeId since the method was only ever used to find an ID suitable for appending rows. Updated the implementation to return the array ID when no RGA nodes are found. This is consistent with the JSON CRDT RGA implementation where an RGA node may reference the array ID.


john-hill (approved):
Good Job!!!!

PR #5528, opened 2026-02-24
PLFM-9475 - Ignore grid data with null upsert key during upsert merge

nickgros:
Users may have entered blank rows or rows missing upsert key information in the grid before attempting an upsert. Records with null upsert key values cannot be added to the temporary tables for upsert. Instead of throwing an error when we encounter a null upsert key value, which effectively blocks upsert functionality, we can skip these rows. The skipped rows will remain in the grid, and the upsert process can proceed.


john-hill (approved):

nickgros:
Passing in code pipeline even though failure is still reported

PR #5538, opened 2026-02-27
PLFM-9493

brucehoff:
- freeze IDs for the realm-principal rows for the default realm to be 1,2,3,4
- remove row deletion for secondary realm tables which are unneeded and can break bootstrapping
- use `insert ignore` for bootstrapping but NOT when creating later realms, since we want DB constraints to be enforced


lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/auth/RealmDaoImpl.java
```diff
@@ -174,61 +177,90 @@ List<DBORealmPrincipal> copyRealmPrincipalsToDBOList(RealmPrincipal realmPrincip
...
+		dbo.setId(principalIdToRealmPrincipalDboId.getOrDefault(dbo.getPrincipalId().toString(),
+				idGenerator.generateNewId(IdType.REALM_PRINCIPAL)));
 		dbo.setPrincipalType(REALM_PRINCIPAL_TYPE_ADMINISTRATORS);
 		result.add(dbo);
 		return result;
 	}
 
-	private DBORealm createPrivate(Realm dto) {
+	private static String sqlInsertCmd(boolean idempotent) {
+		return idempotent ? "INSERT IGNORE" : "INSERT";
```
john-hill, line 210: Why not use insert ignore for all cases?

john-hill (approved):

PR #5550, opened 2026-03-04
PLFM-9520 - fix case where single patch would be stringified

nickgros:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/grid/GridManagerImpl.java
```diff
@@ -399,7 +399,7 @@ JSONArray createMaxSizedPatchArray(String sessionId, List<PatchInfo> candidatePa
 			if (candidate.getSizeBytes() == null) {
 				// Null size means pre-existing record — fallback to single-patch behavior
 				if (arrayOfPatches.length() == 0) {
-					Optional<String> body = getPatchBody(sessionId, candidate);
+					Optional<JSONArray> body = getPatchBody(sessionId, candidate);
 					if (body.isPresent()) {
 						arrayOfPatches.put(body.get());
```
nickgros, line 404: Bug was here where we did not map the string body to a JSON array. Fix by mapping to JSONArray in `getPatchBody`

john-hill (approved):
LGTM

PR #5558, opened 2026-03-10
Claude generated CLAUDE.md files

john-hill:


.claude/settings.local.json
```diff
@@ -0,0 +1,13 @@
+{
+  "permissions": {
+    "allow": [
```
BryanFauble: I think that we would commit a very limited number of permissions (if any) to source control. My concern is by opening up * commands, and committing it to Git, is that anyone who clones this codebase would also run with those permissions, likely without them really knowing that permissions are different from their settings.

CLAUDE.md
```diff
@@ -0,0 +1,199 @@
...
+- Primary tables need an etag column (NOT NULL) for change detection; secondary tables need a foreign key to their owner's backup ID
+- Key test: `services/repository/src/test/java/org/sagebionetworks/repo/web/migration/MigrationIntegrationAutowireTest.java` — extend this when adding new migratable types
+
+### Moving Data Between Tables (cross-stack safe)
+
+Use a two-stack rollout:
+1. **Stack N**: Add data mirroring (write to both old and new table) + backfilling via `MigrationTypeListener` registered in `managers-spb.xml`
+2. **Stack N+1**: Remove mirroring, switch reads to new table as source of truth
+
+## Curation Grid (Curator)
```
BryanFauble, line 136: The Curation Grid / CRDT section is very deep. Consider whether some of that detail could live only in the subdirectory-level files and be discovered on demand rather than loaded at session start.

.claude/settings.local.json
john-hill: I removed this file and added .claude to the gitignore.

CLAUDE.md
john-hill: I agree.  This was the area where Claude indicated the most context was needed (also migration).  Patterns like Controller/manager/dao are common in the industry so Claude only need to add PLFM flavor to those areas.  
Here was the priority list that Claude came up with:
 That completes the full set of CLAUDE.md files across the codebase. Here's the summary of everything created:

  ┌──────────┬───────────────────────────────┬─────────────────────────────────────────────┐
  │ Priority │            Module             │                   Status                    │
  ├──────────┼───────────────────────────────┼─────────────────────────────────────────────┤
  │ Root     │ CLAUDE.md                     │ Updated (databases, change messaging, grid) │
  ├──────────┼───────────────────────────────┼─────────────────────────────────────────────┤
  │ High     │ lib/lib-grid/                 │ Created                                     │
  ├──────────┼───────────────────────────────┼─────────────────────────────────────────────┤
  │ High     │ lib/lib-grid-db/              │ Created                                     │
  ├──────────┼───────────────────────────────┼─────────────────────────────────────────────┤
  │ High     │ lib/jdomodels/                │ Created                                     │
  ├──────────┼───────────────────────────────┼─────────────────────────────────────────────┤
  │ High     │ services/workers/             │ Created                                     │
  ├──────────┼───────────────────────────────┼─────────────────────────────────────────────┤
  │ Moderate │ lib/lib-auto-generated/       │ Created                                     │
  ├──────────┼───────────────────────────────┼─────────────────────────────────────────────┤
  │ Moderate │ services/repository-managers/ │ Created                                     │
  ├──────────┼───────────────────────────────┼─────────────────────────────────────────────┤
  │ Moderate │ lib/lib-table-query/          │ Created                                     │
  ├──────────┼───────────────────────────────┼─────────────────────────────────────────────┤
  │ Low      │ services/repository/          │ Created                                     │
  ├──────────┼───────────────────────────────┼─────────────────────────────────────────────┤
  │ Low      │ lib/lib-utils/                │ Created                                     │
  ├──────────┼───────────────────────────────┼─────────────────────────────────────────────┤
  │ Low      │ lib/models/                   │ Created                                     │
  └──────────┴───────────────────────────────┴─────────────────────────────────────────────┘

.gitignore
```diff
@@ -8,3 +8,4 @@ target/
 .DS_Store
 *.iml
 *.pyc
+.claude/
```
nickgros, line 11: Just ignore local settings and we can refine the CLAUDE.md files by moving more specific information to skills
```suggestion
.claude/settings.local.json
```

CLAUDE.md
```diff
@@ -0,0 +1,199 @@
...
+- Do NOT edit generated classes — edit the JSON schema, then rebuild
+
+## Testing
+
+- Unit tests: `*Test.java` — JUnit 5 + Mockito 2.27
+  - `@ExtendWith(MockitoExtension.class)`, `@Mock`, `@InjectMocks`
+- Integration tests: `IT*.java` (in integration-test module)
+- Mockito 2.27 — no `mockStatic` or Mockito 4/5 APIs
+
+## Deployment & Migration
```
nickgros, line 87: Shorten the migration section and move details to a new skill

CLAUDE.md
nickgros: We could also consider moving some of this information to a committed skill

lib/models/CLAUDE.md
```diff
@@ -0,0 +1,46 @@
+# lib/models
+
+DAO interfaces and shared exception types. This module defines the persistence contracts that `lib/jdomodels` implements. Compiled to **Java 8** for SWC compatibility.
```
nickgros, line 3: Might be able to upgrade to Java 11

nickgros (approved):

PR #5560, opened 2026-03-11
PLFM-9465 - export snapshot from replica index

nickgros:


lib/lib-grid-db/src/main/java/org/sagebionetworks/grid/db/GridIndexManagerImpl.java
```diff
@@ -153,6 +150,98 @@ public void applySnapshot(String sessionId, Long replicaId, Path snapshotFile) {
...
+		ValidateArgument.required(snapshotFile, "snapshotFile");
+
+		// Get the root object to determine the root node ID
+		ObjectNode rootObject = dao.getRootObject(sessionId, replicaId)
+				.orElseThrow(() -> new IllegalStateException(
+						"No root object found for session: " + sessionId + " replica: " + replicaId));
+
+		LogicalTimestamp rootNodeId = rootObject.getId();
+
+		try (OutputStream out = Files.newOutputStream(snapshotFile, StandardOpenOption.CREATE,
```
nickgros: Should use fileProvider

lib/lib-grid-db/src/main/java/org/sagebionetworks/grid/db/GridIndexDao.java
```diff
@@ -349,4 +349,57 @@ ArrayNode getArrayNode(String sessionIdString, Long replicaId, LogicalTimestamp
 	 */
 	Long getClockSequenceMaximum(String gridSessionId, Long replicaId);
 
+	/**
+	 * Stream all constant nodes for a session/replica (paginated).
```
nickgros: Is there an existing utility that would convert the pagination to iterable/stream?

lib/lib-grid-db/src/main/java/org/sagebionetworks/grid/db/GridIndexManagerImpl.java
```diff
@@ -153,6 +150,98 @@ public void applySnapshot(String sessionId, Long replicaId, Path snapshotFile) {
...
+				StandardOpenOption.TRUNCATE_EXISTING);
+				IndexedModelEncoder encoder = new IndexedModelEncoder(out, rootNodeId)) {
+
+			// Stream and write all constants (paginated)
+			exportConstants(sessionId, replicaId, encoder);
```
nickgros, line 184: If we walk the tree instead of streaming straight from the node tables, we could remove orphaned nodes. More complex, probably slower, but this might be the best opportunity to prune.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/grid/GridSnapshotCompactionManagerImpl.java
```diff
@@ -0,0 +1,132 @@
...
+				replicaId);
+		if (!currentClock.isPresent()) {
+			log.debug("Session {} replica {} is not fully synchronized. Skipping.", sessionId, replicaId);
+			return false;
+		}
+
+		// Create a temp file for the snapshot
+		Path tempFile = null;
+		try {
+			tempFile = Files.createTempFile("grid-snapshot-", ".cbor");
```
nickgros: Should use FileProvider

lib/lib-grid-db/src/main/java/org/sagebionetworks/grid/db/GridIndexManagerImpl.java
```diff
@@ -153,6 +150,98 @@ public void applySnapshot(String sessionId, Long replicaId, Path snapshotFile) {
 		dao.setClocks(sessionId, replicaId, snapshotClockTable.getClocks());
 	}
 
+	@Override
+	public ClockTable exportSnapshot(String sessionId, Long replicaId, Path snapshotFile) {
```
nickgros, line 168: `@GridTransaction` by default uses READ_COMMITTED, but I think we may need a stronger consistency guarantee for this method

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/grid/GridSnapshotCompactionManagerImpl.java
```diff
@@ -0,0 +1,132 @@
...
+
+	@Override
+	public int compactSessions(ProgressCallback callback) {
+		ValidateArgument.required(callback, "callback");
+
+		Duration maxAge = Duration.ofDays(stackConfiguration.getGridSnapshotMaxAgeDays());
+		int maxPatchCount = stackConfiguration.getGridSnapshotMaxPatchCount();
+		int batchSize = stackConfiguration.getGridSnapshotCompactionBatchSize();
+
+		List<String> sessionIds = gridDao.listSessionsNeedingCompaction(maxAge, maxPatchCount, batchSize);
```
nickgros: We should probably decouple the scanner from the compactor. The scanner could push onto a queue. The scanner could have a much larger batch size, and the compactor would just have a batch size of 1.

lib/lib-grid-db/src/main/java/org/sagebionetworks/grid/db/GridTransaction.java
```diff
@@ -34,4 +34,10 @@
 	 */
 	@AliasFor(annotation = Transactional.class)
 	boolean readOnly() default false;
+
+	/**
+	 * Alias for {@link Transactional#isolation}. Defaults to {@link Isolation#READ_COMMITTED}.
+	 */
+	@AliasFor(annotation = Transactional.class)
+	Isolation isolation() default Isolation.READ_COMMITTED;
```
nickgros: Allow overriding Tx isolation level

lib/lib-grid-db/src/main/java/org/sagebionetworks/grid/db/GridIndexManagerImpl.java
```diff
@@ -150,7 +150,125 @@ public void applySnapshot(String sessionId, Long replicaId, Path snapshotFile) {
...
+		// 'next available' sequence number. Increment each clock entry before storing in the database
+		ClockTable dbClockTable = new ClockTable(new ArrayList<>());
+		for (LogicalTimestamp snapshotEntry : snapshotClockTable.getClocks()) {
+			LogicalTimestamp dbEntry = new LogicalTimestamp()
+					.setReplicaId(snapshotEntry.getReplicaId())
+					.setSequenceNumber(snapshotEntry.getSequenceNumber() + 1L);
+			dbClockTable.updateClockTable(dbEntry);
+		}
+		dao.setClocks(sessionId, replicaId, dbClockTable.getClocks());
+	}
```
nickgros: Fix snapshot clock database representation to be the same as the patch clock database representation.

lib/lib-grid-db/src/main/java/org/sagebionetworks/grid/db/GridIndexManagerImpl.java
```diff
@@ -150,7 +150,125 @@ public void applySnapshot(String sessionId, Long replicaId, Path snapshotFile) {
...
+
+	@Override
+	@GridTransaction(readOnly = true, isolation = Isolation.REPEATABLE_READ)
```
nickgros: Use REPEATABLE_READ because phantom reads are not acceptable when serializing a snapshot.

Consider the following sequence:

1. `exportSnapshot` serializes constants, but has not serialized other rows
2. Replica 123 applies a patch in another transaction. In order, the patch add a constant (seq=4) then adds an array (seq=5).
3. `exportSnapshot` serializes arrays (capturing the committed array 123.5)

The snapshot includes the array 123.5 from the patch, but not the constant 123.4.

The snapshot clock would be 123.5, which incorrectly indicates that the entire patch is applied, even though the constant 123.4 is missing.

lib/stackConfiguration/src/main/resources/stack.properties
```diff
@@ -411,4 +411,9 @@ org.sagebionetworks.cross.account.bedrock.role.arn=arn:aws:iam::050451359079:rol
 org.sagebionetworks.cross.account.bedrock.hello.world.agent.id=0O3IDUIR36
 
 # Default project storage limit for the synapse bucket
-org.sagebionetworks.project.storage.default.limit=107374182400
\ No newline at end of file
+org.sagebionetworks.project.storage.default.limit=107374182400
+
+# Grid snapshot compaction settings
+org.sagebionetworks.grid.snapshot.max.age.days=30
+org.sagebionetworks.grid.snapshot.max.patch.count=1000
```
nickgros: Look at prod and investigate patches per session to determine a reasonable number

lib/lib-grid-db/src/main/java/org/sagebionetworks/grid/db/GridIndexManagerImpl.java
nickgros: Evaluate REPEATABLE_READ vs 
- semaphore write lock (also apply to `onApplyPatches` and `onApplySnapshot`, or
- trigger snapshots using the existing FIFO queue (effectively SERIALIZABLE)

lib/jdomodels/src/test/java/org/sagebionetworks/evaluation/dbo/grid/GridDaoImplTest.java
```diff
@@ -805,4 +805,242 @@ public void testGetLatestSnapshotWithNullSessionId() {
...
+				new LogicalTimestamp().setReplicaId(1L).setSequenceNumber(10L)));
+		dao.saveSnapshot(session.getSessionId(), clockTable, "test-key", adminUserId);
+
+		// Wait for the snapshot to exceed the max age
+		Thread.sleep(2000);
+
+		// With maxAge=1 second, the snapshot is old enough even though there are no new patches
+		// call under test
+		List<String> result = dao.listSessionsNeedingCompaction(Duration.ofSeconds(1), 1000, 10);
+		assertTrue(result.contains(session.getSessionId()));
```
nickgros: Shouldn't this be empty?

lib/lib-grid-db/src/main/resources/schema/Grid-Constant-ddl.sql
```diff
@@ -5,7 +5,7 @@ CREATE TABLE IF NOT EXISTS GRID_REPLICA_CON (
 	CON_SEQ BIGINT NOT NULL,
     CON_VAL JSON NOT NULL,
     CON_VAL_HASH INT UNSIGNED AS (CRC32(JSON_EXTRACT(CON_VAL, '$'))) STORED,
-	PRIMARY KEY (SESSION_ID, REPLICA_ID, CON_SEQ, CON_REP),
+	PRIMARY KEY (SESSION_ID, REPLICA_ID, CON_REP, CON_SEQ),
 	INDEX IDX_GRID_REPLICA_CON_HASH (CON_VAL_HASH),
     CONSTRAINT `GRID_CON_REPLICA_FK` FOREIGN KEY (SESSION_ID, REPLICA_ID) REFERENCES GRID_REPLICA(SESSION_ID, REPLICA_ID) ON DELETE CASCADE
 );
```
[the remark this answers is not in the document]
nickgros: Not necessary, table is rebuilt each deployment

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/grid/GridSessionSnapshotPublisherImpl.java
```diff
@@ -0,0 +1,85 @@
...
+		this.stackConfiguration = stackConfiguration;
+		this.sqsClient = sqsClient;
+		this.sqsQueueUrl = sqsClient
+				.getQueueUrl(GetQueueUrlRequest.builder()
+						.queueName(stackConfiguration.getQueueName(INTERNAL_EVENT_QUEUE_NAME)).build())
+				.queueUrl();
+	}
+
+	@Override
+	public List<String> scanAndPublishSessionsNeedingSnapshot() {
```
nickgros: On the first deployment of this change to staging, this will create snapshot files that will never be used by production, and be cleaned up on each migration. Do we want to prevent this case?

lib/lib-grid-db/src/main/resources/schema/Grid-Constant-ddl.sql
```diff
@@ -5,7 +5,7 @@ CREATE TABLE IF NOT EXISTS GRID_REPLICA_CON (
 	CON_SEQ BIGINT NOT NULL,
     CON_VAL JSON NOT NULL,
     CON_VAL_HASH INT UNSIGNED AS (CRC32(JSON_EXTRACT(CON_VAL, '$'))) STORED,
-	PRIMARY KEY (SESSION_ID, REPLICA_ID, CON_SEQ, CON_REP),
+	PRIMARY KEY (SESSION_ID, REPLICA_ID, CON_REP, CON_SEQ),
```
nickgros, line 8: All other node tables put REP before SEQ, so just update this index to match.

The queries to stream nodes are sensitive to this index across pages

lib/lib-grid-db/src/main/java/org/sagebionetworks/grid/db/GridIndexDaoImpl.java
```diff
@@ -796,5 +796,96 @@ public Long getClockSequenceMaximum(String sessionIdString, Long replicaId) {
...
+					CONSTANT_NODE_MAPPER, sessionId, replicaId, limit);
+		} else {
+			return jdbcTemplate.query(
+					"SELECT CON_REP, CON_SEQ, CON_VAL FROM GRID_REPLICA_CON"
+							+ " WHERE SESSION_ID = ? AND REPLICA_ID = ?"
+							+ " AND ((CON_REP = ? AND CON_SEQ > ?) OR CON_REP > ?)"
+							+ " ORDER BY CON_REP, CON_SEQ LIMIT ?",
+					CONSTANT_NODE_MAPPER, sessionId, replicaId,
+					lastSeen.getReplicaId(), lastSeen.getSequenceNumber(), lastSeen.getReplicaId(), limit);
+		}
```
nickgros, line 831: `lastSeen` allows us to use the index across pages and ensure O(n) streaming instead of \~O(n^2) (affected by page size)

lib/lib-grid-db/CLAUDE.md
```diff
@@ -77,6 +77,15 @@ Snapshots are imported without loading the entire file into memory:
 3. Nodes are batched (default 1000) and persisted via `GridIndexDao`
 4. Import order: constants → objects → values → vectors (with constant lookups) → arrays + RGA elements
 
+### Snapshot Export Flow
+
+`GridIndexManager.exportSnapshot()` streams a replica's current state out to a CBOR file without loading the full document into memory:
+1. Has `REPEATABLE_READ` transaction level for a consistent read across paginated queries
```
nickgros: wrong

lib/lib-grid-db/CLAUDE.md
```diff
@@ -87,7 +96,7 @@ The `GRID_REPLICA_RGA` table implements the Replicated Growable Array with prede
 
 ### `@GridTransaction` Annotation
 
-Custom transaction annotation that binds to `gridTransactionManager` (NOT the main app's primary transaction manager). Required because the grid uses a **separate database** from the main Synapse database. Always uses `READ_COMMITTED` isolation and `REQUIRED` propagation.
+Custom transaction annotation that binds to `gridTransactionManager` (NOT the main app's primary transaction manager). Required because the grid uses a **separate database** from the main Synapse database. Defaults to `READ_COMMITTED` isolation and `REQUIRED` propagation. Supports an `isolation` attribute to override per-method — snapshot export uses `REPEATABLE_READ` to ensure a consistent view across paginated reads.
```
nickgros: wrong

lib/lib-grid/src/main/java/org/sagebionetworks/repo/model/grid/encoding/IndexedModelEncoder.java
```diff
@@ -90,13 +91,12 @@ public void writeNode(Node node) throws IOException {
 		String nodeKey = clockTable.encodeNodeKey(node.getId());
 
 		// Encode the node to binary
-		try (ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
-			nodeCodec.encode(node, clockTable, baos);
+		nodeBuffer.reset();
```
nickgros, line 94: minor optimization; reuse the same buffer. 1 alloc instead of _n_

lib/stackConfiguration/src/main/resources/stack.properties
nickgros: No session created since Oct 2025 has >1000 patches

```sql
SELECT
	avg(count),
	MIN(count),
	MAX(count)
FROM
	(
	SELECT
		gs.ID ,
		gs.OWNER_ID,
--		pa.ALIAS_UNIQUE,
		gs.CREATED_ON,
		COUNT(*) AS Count,
		MAX(gp.CREATED_ON) as latest_patch
	FROM
		GRID_SESSION gs
	LEFT JOIN GRID_PATCH gp on
		gs.ID = FROM_BASE64(gp.SESSION_ID)
--	LEFT JOIN PRINCIPAL_ALIAS pa ON
--		gs.OWNER_ID = pa.PRINCIPAL_ID AND pa.`TYPE` ='USER_NAME'
	GROUP BY
		gp.SESSION_ID
	ORDER BY
		Count DESC
    ) as counts;
```

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/grid/GridSessionSnapshotPublisherImpl.java
```diff
@@ -0,0 +1,85 @@
...
+import software.amazon.awssdk.services.sqs.SqsClient;
+import software.amazon.awssdk.services.sqs.model.GetQueueUrlRequest;
+import software.amazon.awssdk.services.sqs.model.MessageAttributeValue;
+import software.amazon.awssdk.services.sqs.model.SendMessageRequest;
+
+@Service
+public class GridSessionSnapshotPublisherImpl implements GridSessionSnapshotPublisher {
+
+	static final String INTERNAL_EVENT_QUEUE_NAME = "GRID_INTERNAL_EVENT.fifo";
+	static final String NEW_SNAPSHOT_NOTIFICATION = "[8,\"new-snapshot\"]";
```
nickgros: consider adding IT test where a replica sends this message

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/grid/GridDao.java
```diff
@@ -219,4 +219,27 @@ public interface GridDao {
...
+	 *   - Its latest snapshot is older than maxSnapshotAge, OR
+	 *   - It has more than maxPatchCount patches since the latest snapshot
+	 * Only sessions with an INTERNAL connection are returned.
+	 *
+	 * @param maxSnapshotAge Maximum age of the latest snapshot
+	 * @param maxPatchCount  Maximum number of patches since the latest snapshot
+	 * @param limit          Maximum number of sessions to return
+	 * @return List of session IDs needing a new snapshot
+	 */
+	List<String> listSessionsNeedingSnapshot(Duration maxSnapshotAge, int maxPatchCount, int limit);
```
nickgros: If we keep all patches, and mark them for deletion after creating a snapshot, we can avoid a race condition in downloading patches by changing server logic to supply a snapshot if patches were deleted before the client could receive them.

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/grid/GridDaoImpl.java
```diff
@@ -498,4 +502,18 @@ public Optional<GridSource> getSessionSource(String sessionId) {
 		}
 	}
 
+	@Override
+	public int countPatchesSinceLatestSnapshot(String sessionId) {
+		ValidateArgument.required(sessionId, "sessionId");
```
john-hill: If we delete patches after creating a snapshot, then we just need to count the patches for that session.

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/grid/GridDao.java
john-hill: The simplest we can make this would be to make no attempt to delete old patches.  We simply keep them.  Instead, we mark each snapshot with the last patch ID included in that snapshot (null for the first snapshot).  This patch_id is then used to determine when we create new snapshots.
Trigger a job to check if a snapshot should be taken when we save a new patch (on the same FIFO queue).

nickgros: Not sure if patches are inserted into the index in the same order as the IDs are generated. Clock table is safer but more complex. I will determine if we can use the simpler patch ID

lib/lib-grid-db/src/main/java/org/sagebionetworks/grid/db/GridIndexManagerImpl.java
```diff
@@ -150,7 +149,134 @@ public void applySnapshot(String sessionId, Long replicaId, Path snapshotFile) {
...
+			/*
+			 * Patches may have incremented the clock without creating corresponding nodes. This is expected; not all
+			 * patch operations create nodes).
+			 *
+			 * Ensure these operations are accounted for by using the replica's database clock rather than the encoder's
+			 * node-derived clock (which is intended to only be used for newly-instantiated grids).
+			 */
+			ClockTable dbClock = new ClockTable(dao.getClock(sessionId, replicaId));
+			ClockTable clockTable = encoder.getClockTable();
+			for (LogicalTimestamp dbEntry : dbClock.getClocks()) {
```
john-hill: This seems common, can add a static to LogicalTimestamp?

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/grid/GridDaoImpl.java
```diff
@@ -435,9 +433,27 @@ public List<PatchInfo> listMissingPatchInfoForClock(String sessionId, List<Logic
...
 
+	@Override
+	public int countMissingPatchesForClock(String sessionId, List<LogicalTimestamp> clock) {
+		ValidateArgument.required(sessionId, "sessionId");
+		ValidateArgument.required(clock, "clock");
+		if (clock.isEmpty()) {
+			clock = List.of(new LogicalTimestamp().setReplicaId(0L).setSequenceNumber(0L));
+		}
+		StringJoiner rows = new StringJoiner(",");
+		clock.forEach(id -> {
```
john-hill, line 448: refactor to remove duplication.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/grid/internal/replica/GridReplicaManagerImpl.java
```diff
@@ -49,19 +47,22 @@ public class GridReplicaManagerImpl implements GridReplicaManager {
...
 	private final SnsClient snsClient;
 	private final String topicArn;
 	private final HttpClient httpClient;
 
-	public GridReplicaManagerImpl(GridIndexManager gridIndexManager, InternalReplicaToHubEventPublisher publisher,
-			SnsClient snsClient, String gridReplicaChangeTopicArn, HttpClient httpClient) {
+	public GridReplicaManagerImpl(GridIndexManager gridIndexManager,
+			GridReplicaSnapshotManager snapshotManager, InternalReplicaToHubEventPublisher publisher,
+			SnsClient snsClient, String gridReplicaChangeTopicArn,
+			HttpClient httpClient) {
```
john-hill, line 60: will remove this.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/grid/SnapshotRowHandler.java
```diff
@@ -467,7 +468,14 @@ public void close() throws IOException {
...
         try {
             finalizeEncoding();
-            snapshotStore.saveSnapshot(this.sessionId, this.encoder.getClockTable(), createdByUserId, snapshotFile);
+            /*
+             * A JSON CRDT snapshot's clock table is stored as the maximum sequence number seen across
+             * node IDs ("last-seen"), but clock representations stored in the patch and snapshot tables
+             * use the next available sequence number (patchId.sequenceNumber + span), which
+             * is one past the last used ID. Increment each entry by 1 to align with this convention.
+             */
+            ClockTable clockTable = this.encoder.getClockTable().incrementClocks();
```
john-hill: Since we seem to be off by one, what does the client send when sending a synch-clock?

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/grid/internal/replica/GridReplicaManagerImplTest.java
```diff
@@ -68,6 +68,8 @@ public class GridReplicaManagerImplTest {
 	HttpClient mockHttpClient;
 	@Mock
 	private HttpResponse<Path> mockHttpResponse;
+	@Mock
+	private GridReplicaSnapshotManager mockSnapshotManager;
```
john-hill: not needed

lib/lib-grid-db/src/main/java/org/sagebionetworks/grid/db/GridIndexManagerImpl.java
```diff
@@ -150,11 +150,8 @@ public void applySnapshot(String sessionId, Long replicaId, Path snapshotFile) {
 				.setValue(index.getRootNodeId()))
 		);
 
-		// Update the replica clock
-		// The snapshot encodes the clocks as the 'last used' sequence number, but our clock table stores the
-		// 'next available' sequence number. Increment each clock entry before storing in the database
-		ClockTable dbClockTable = snapshotClockTable.copy().incrementClocks();
```
nickgros, line 156: I removed this increment/decrement clock logic. The JSON Joy client implementation and Synapse DB (both GRID_REPLICA_CLOCK and GRID_SNAPSHOT) store the clock using the same convention -- the last used sequence number.

The patch database stores the 'first-used' sequence number in the patch, which we use to identify outstanding patches using a replica's clock.

I originally added this logic because at one point I was using a given clock to filter rows in both the patch table and the snapshot table, which follow different conventions in that context. 

That code has since been removed. We never use a clock to filter rows in the snapshot table. Therefore, we never need to consider these different conventions that could cause an off-by-one error.

services/workers/src/main/java/org/sagebionetworks/grid/workers/GridSessionIndexWorker.java
```diff
@@ -50,10 +54,11 @@ public void run(ProgressCallback progressCallback, ChangeMessage message) throws
 			return;
 		}
 		log.info("Firing new-patch for grid session: {}", sessionId);
-		//
 		publisher.publishEventResponses(
 				List.of(new EventContext(EventType.MESSAGE, EventSource.INTERNAL,
 						connection.get().getConnectionId())),
 				JsonRxMessageType.Notification, "new-patch");
+		// Additionally, see if a new snapshot should be created.
+		snapshotExportRequester.requestSnapshotExportIfNeeded(sessionId);
```
nickgros, line 62: In the same place where we send the `new-patch` notification, we also trigger the check to see if we should issue a `new-snapshot` notification.

john-hill: This event is triggered on staging to ensure that replicas are built on staging.

john-hill (approved):

PR #5561, opened 2026-03-11
send a CHANGE message when saving a grid patch.  This message will mi…

john-hill:
…grate and be triggered on the destination stack ensuring the grid is rebuilt on the destination


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/grid/GridManagerImpl.java
```diff
@@ -548,4 +560,23 @@ public void saveSnapshot(String sessionId, ClockTable clockTable, Long createdBy
 		gridDao.saveSnapshot(sessionId, clockTable, uploadResult.getKey(), createdBy);
 	}
 
+	@WriteTransaction
+	@Override
+	public long backfillGridSessionChanges() {
```
nickgros, line 565:
```suggestion
	@TemporaryCode
	public long backfillGridSessionChanges() {
```

nickgros (approved):

PR #5563, opened 2026-03-13
a new API to list the replicas associated with a grid session

john-hill:


services/workers/src/test/java/org/sagebionetworks/grid/workers/GridEventListenerTest.java
```diff
@@ -154,9 +166,33 @@ public void testOnConnectionWithNullContext() throws JSONObjectAdapterException
 	public void testOnDisconnected(EventType type) {
 		context = new EventContext(type, eventSource, connectionId);
 		disconnectMessage = new DisconnectedMessage(context, null, null);
+		GridConnectionInfo thisConnection = new GridConnectionInfo().setConnectionId(connectionId)
+				.setSource(EventSource.WEBSOCKET);
+		when(mockManager.getConnectionInfoOptional(connectionId)).thenReturn(Optional.of(thisConnection));
+		List<GridConnectionInfo> activeCons = List.of(thisConnection,
+				new GridConnectionInfo().setConnectionId("con999").setSource(EventSource.INTERNAL));
```
nickgros, line 173: `activeCons` mock should include the current connectionId (which would be filtered out in the method under test).

Consider refactoring the lambda to reuse it and have a shared way to get 'other' connections

nickgros (approved):

PR #5592, opened 2026-04-12
PLFM-9573: Update mysqlconnector/J client

xschildw:
This PR updates the mySQL client dependency to 8.4.0 (backward compatible with 8.0x).


lib/database-semaphore/pom.xml
```diff
@@ -47,8 +47,8 @@
 			<scope>test</scope>
 		</dependency>
 		<dependency>
-			<groupId>mysql</groupId>
-			<artifactId>mysql-connector-java</artifactId>
+			<groupId>com.mysql</groupId>
+			<artifactId>mysql-connector-j</artifactId>
 			<version>${mysql.mysql-connector-java.version}</version>
```
john-hill: remove the version here for each, the version should only be in the root pom.xml

john-hill (changes requested):

lib/database-semaphore/pom.xml
xschildw: Done

john-hill (approved):
LGTM.

PR #5612, opened 2026-04-29
Plfm 9596

SandhraSokhal:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/team/TeamManagerImpl.java
```diff
@@ -231,13 +231,8 @@ public static void addToACL(AccessControlList acl, String principalId, Set<ACCES
...
 	
-	public static void removeFromACL(AccessControlList acl, String principalId) {
-		Set<ResourceAccess> origRA = acl.getResourceAccess();
-		Set<ResourceAccess> newRA = new HashSet<ResourceAccess>();
-		for (ResourceAccess ra: origRA) {
-			if (!principalId.equals(ra.getPrincipalId().toString())) newRA.add(ra);
-		}
-		acl.setResourceAccess(newRA);
+	public static boolean removeFromACL(AccessControlList acl, String principalId) {
+		return acl.getResourceAccess().removeIf(ra -> principalId.equals(ra.getPrincipalId().toString()));
```
brucehoff, line 235: Do we know for sure that the Set `acl.getResourceAccess()` is mutable?

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/team/TeamManagerImpl.java
```diff
@@ -728,13 +723,15 @@ public void removeMember(UserInfo userInfo, String teamId,
...
+			boolean removed = removeFromACL(acl, principalId);
 			if (!userInfo.isAdmin() && !aclHasTeamAdmin(acl)) {
 				throw new InvalidModelException(MSG_TEAM_MUST_HAVE_AT_LEAST_ONE_TEAM_MANAGER);
 			}
 			groupMembersDAO.removeMembers(teamId, Collections.singletonList(principalId));
-			Team team = get(teamId);
-			aclManager.update(userInfo, acl, ObjectType.TEAM, getTeamOwner(team));
+			if (removed) {
+				Team team = get(teamId);
+				aclManager.update(userInfo, acl, ObjectType.TEAM, getTeamOwner(team));
```
brucehoff: call DAO directly.  Add comment:

// ACLManager.update generally disallows removing self from ACL.
// Since we need to remove the user from the ACL here,
// we bypass the manager and call the DAO directly

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/team/TeamManagerImpl.java
```diff
@@ -728,13 +723,17 @@ public void removeMember(UserInfo userInfo, String teamId,
...
+			boolean removed = removeFromACL(acl, principalId);
 			if (!userInfo.isAdmin() && !aclHasTeamAdmin(acl)) {
 				throw new InvalidModelException(MSG_TEAM_MUST_HAVE_AT_LEAST_ONE_TEAM_MANAGER);
 			}
 			groupMembersDAO.removeMembers(teamId, Collections.singletonList(principalId));
-			Team team = get(teamId);
-			aclManager.update(userInfo, acl, ObjectType.TEAM, getTeamOwner(team));
+			if (removed) {
+				// Bypass aclManager.update because validateACLContent enforces "caller must retain ACL edit
+				// permission" — that's an ACL-edit invariant, not a team-leave invariant. We need to remove user form ACL here.
```
brucehoff: a small typo: "remove user form" should be "remove user from"

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/team/TeamManagerImpl.java
```diff
@@ -816,16 +815,18 @@ public void setPermissions(UserInfo userInfo, String teamId,
...
 		// now, if isAdmin is false, the team membership is enough to give the user basic permissions
 		if (isAdmin) {
 			// if isAdmin is true, then we add the specified admin permissions
 			addToACL(acl, principalId, ModelConstants.TEAM_ADMIN_PERMISSIONS);
 		}
 		if (!userInfo.isAdmin() && !aclHasTeamAdmin(acl)) throw new InvalidModelException(MSG_TEAM_MUST_HAVE_AT_LEAST_ONE_TEAM_MANAGER);
 		// finally, update the ACL
-		Team team = get(teamId);
-		aclManager.update(userInfo, acl, ObjectType.TEAM, getTeamOwner(team));
+		if (removed || isAdmin) {
```
brucehoff: A small request, can you rename "isAdmin" to "isManager" or "isTeamManager", to make it clear that the boolean does not mean the member is a Synapse admin?

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/team/TeamManagerImpl.java
```diff
@@ -816,16 +815,18 @@ public void setPermissions(UserInfo userInfo, String teamId,
...
+			Team team = get(teamId);
+			aclManager.update(userInfo, acl, ObjectType.TEAM, getTeamOwner(team));
```
brucehoff: doesn't this suffer from the same problem that occurs when trying to leave a team? that is, if a user downgrades their permissions from team manager to a regular team member, then would AclManager.update() reject the ACL change, saying that the user is removing their own team management permissions?

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/team/TeamManagerImplTest.java
```diff
@@ -1282,7 +1364,24 @@ public void testSetPermissions() {
 		}
 		assertFalse(foundRA);
 	}
-	
+
+	@Test
+	public void testSetPermissionsWithNoChange() {
```
brucehoff, line 1382: I think we need another test:
- userInfo is NOT team owner;
- userInfo IS a team manager;
- userInfo downgrades themselves from team manager to non-manager;

To see why this fails we have to call AclManager.update(), i.e.:

aclManager.update(userInfo, modifiedAcl, ObjectType.TEAM, owner-who-is-distinct-from-userInfo);

where modifiedAcl does not have userInfo in it.

The test may actually have to go in the unit test for AclManager or in an integration test for TeamManagerImpl since we need to exercise AclManager.update() (not just mock it) to see the issue.

Once the issue is seen, I think the fix will be to call aclDAO.update() in setPermissions().

brucehoff (changes requested):

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/team/TeamManagerImplTest.java
brucehoff: I think we need a 'verify' here, checking that aclDao.update() is called with the correctly modified ACL, don't we? That is, the check shown here should be on the captured argument to update() in the Mock ACLDao.

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/team/TeamManagerImplTest.java
```diff
@@ -1282,13 +1362,29 @@ public void testSetPermissions() {
 		}
 		assertFalse(foundRA);
```
brucehoff: Same as above: I think we need a 'verify' here, checking that aclDao.update() is called with the correctly modified ACL, don't we? That is, the check shown here should be on the captured argument to update() in the Mock ACLDao.

brucehoff (changes requested):

brucehoff (approved):

PR #5642, opened 2026-05-25
PLFM-9648

brucehoff:
This PRs adds the DocuSign client and uses it in a service that lists available DocuSign templates.

The non-shaded approach to incorporating the Docsign client is not viable.  The `com.docusign.esign.client.auth.OAuth` class references jakarta.ws.rs directly, causing compile-time errors when we take the non-shaded approach. This is because the DocuSign SDK's published POM declares zero compile dependencies:

In the published POM on Maven Central, the only <dependencies> entries are:
  - junit (test scope)
  - junit-jupiter-api (test scope)
  
  That's it. Despite the compiled classes referencing jakarta.ws.rs, org.glassfish.jersey, com.fasterxml.jackson, org.bouncycastle, com.auth0.jwt, etc., none of those are declared as compile-scope dependencies in the POM.

This is because DocuSign's build uses maven-shade-plugin with shadedArtifactAttached=true — it produces the shaded JAR as an attached artifact (with the shaded classifier), but the shade plugin also strips the dependencies from the main POM since they're "already bundled" in the shaded output. The plain JAR is published alongside it, but its POM doesn't tell Maven to fetch the libraries it actually needs.
  
In short: DocuSign's build process was designed around the shaded JAR as the primary deliverable. The plain JAR is a build artifact that got published to Maven Central, but it's not self-sufficient — its POM doesn't declare the runtime dependencies that the compiled bytecode requires.

The shaded classifier is the only supported consumption path.

__________________

Should cachedAccessToken and cachedAccessTokenExpiryMillis be "volitile"?                                                                                     

⏺ No — they don't need to be volatile because all reads and writes are inside synchronized methods. The synchronized keyword provides both mutual exclusion and memory visibility (the Java Memory Model guarantees that changes made inside a synchronized block are visible to any thread that subsequently enters a synchronized block on the same monitor).

You'd need volatile if you were doing the double-checked locking pattern (read outside synchronized, then lock to write) — which is what the original DocuSignClient code was doing. But since DocuSignAccessTokenProvider synchronizes on every access, volatile would be redundant.


integration-test/src/test/java/org/sagebionetworks/ITEDucTemplates.java
```diff
@@ -0,0 +1,34 @@
...
+		this.config = config;
+	}
+
+	@Test
+	public void testListEDucTemplates() throws SynapseException {
+		assumeTrue(config.getDocuSignEnabled(),
+				"DocuSign integration is disabled — skipping eDUC template IT test.");
+
+		// call under test
+		EDucTemplatePage page = adminSynapse.listEDucTemplates(null);
```
john-hill: non-admin?

services/repository/src/main/java/org/sagebionetworks/repo/web/controller/DataAccessController.java
```diff
@@ -316,6 +318,25 @@ public void deleteSubmission(
...
+	 * @return a page of eDUC template metadata
+	 */
+	@RequiredScope({view})
+	@ResponseStatus(HttpStatus.OK)
+	@RequestMapping(value = UrlHelpers.EDUC_TEMPLATE, method = RequestMethod.GET)
+	public @ResponseBody EDucTemplatePage listEDucTemplates(
+			@RequestParam(value = AuthorizationConstants.USER_ID_PARAM) Long userId,
+			@RequestParam(value = UrlHelpers.NEXT_PAGE_TOKEN_PARAM, required = false) String nextPageToken)
+			throws Exception {
+		return serviceProvider.getEDucService().listTemplates(userId, nextPageToken);
```
john-hill: Can we use a Request object (with a POST) the request would include the nextPageToken and any future params (sorting...).

client/synapseJavaClient/src/main/java/org/sagebionetworks/client/SynapseClientImpl.java
```diff
@@ -5786,6 +5787,17 @@ public OpenSubmissionPage getOpenSubmissions(String nextPageToken) throws Synaps
 		return getJSONEntity(getRepoEndpoint(), url, OpenSubmissionPage.class);
 	}
 
+	private static final String EDUC_TEMPLATE = "/eDuc/template";
+
+	@Override
```
john-hill, line 5794: Will not need to modify the URL once we have a request.

lib/lib-docusign/src/main/java/org/sagebionetworks/docusign/DocuSignClient.java
```diff
@@ -0,0 +1,191 @@
...
+public class DocuSignClient {
+
+	static final long JWT_EXPIRES_IN_SECONDS = 3600L;
+	static final long TOKEN_EXPIRY_BUFFER_MILLIS = 60_000L;
+	static final List<String> SCOPES = Arrays.asList("signature", "impersonation");
+
+	private final DocuSignClientConfig config;
+	private final TemplatesApiFactory templatesApiFactory;
+
+	private volatile String cachedAccessToken;
```
john-hill: Can we move these three members to their own class.  The class would control access to the token and expiration and can ensure the token is always access from a synchronized block?

lib/lib-docusign/src/main/java/org/sagebionetworks/docusign/DocuSignClient.java
```diff
@@ -0,0 +1,191 @@
...
+		int code = e.getCode();
+		switch (code) {
+			case 401:
+				return new UnauthorizedException("DocuSign rejected the access token.", e);
+			case 403:
+				return new UnauthorizedException("DocuSign denied access for the configured user.", e);
+			case 404:
+				return new NotFoundException("DocuSign resource not found.", e);
+			default:
+				return new ServiceUnavailableException(
```
john-hill: Let's assume an error from Docusign is non-recoverable (500) and not 503.

lib/lib-docusign/src/main/java/org/sagebionetworks/docusign/DocuSignClient.java
```diff
@@ -0,0 +1,191 @@
...
+			EnvelopeTemplateResults results = templatesApi.listTemplates(config.getAccountId(), options);
+			return toEDucTemplatePage(results);
+		} catch (ApiException e) {
+			throw convertApiException(e);
+		}
+	}
+
+	private String getAccessToken() throws ServiceUnavailableException {
+		long now = System.currentTimeMillis();
+		String token = cachedAccessToken;
```
john-hill: here is read outside of the synchronized block.

lib/lib-docusign/src/main/java/org/sagebionetworks/docusign/DocuSignClient.java
```diff
@@ -0,0 +1,191 @@
...
+			return null;
+		}
+		return Date.from(Instant.parse(iso8601));
+	}
+
+	static Exception convertApiException(ApiException e) throws ServiceUnavailableException {
+		int code = e.getCode();
+		switch (code) {
+			case 401:
+				return new UnauthorizedException("DocuSign rejected the access token.", e);
```
john-hill: Normally the Unauthorized is handled by a client but they client cannot do anything about this (maybe another 500)?

lib/lib-docusign/pom.xml
```diff
@@ -0,0 +1,154 @@
...
+			      conflicts with Synapse's 2.20.0 (missing StreamReadConstraints
+			      constructor breaks OpenSearch).
+			    We keep jackson/jakarta/** and jackson/module/** because DocuSign's
+			    bundled Jersey (jakarta-based) needs them and Synapse doesn't provide
+			    jakarta xmlbind annotation support.
+			-->
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-shade-plugin</artifactId>
+				<version>3.5.0</version>
```
john-hill: Can we try the non-shadded version?

services/repository-managers/src/main/resources/managers-import.xml
```diff
@@ -9,5 +9,6 @@
 	<!-- Markdown service -->
 	<import resource="classpath:markdown-dao.spb.xml" />
 	<import resource="classpath:doi-spb.xml" />
+	<import resource="classpath:docusign-spb.xml" />
```
john-hill: Is this needed if we are using Spring annotations instead of XML?

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/asynch/AsynchJobQueuePublisherImplTest.java
```diff
@@ -37,6 +38,7 @@ public void before(){
 		asynchJobQueuePublisher.emptyAllQueues();
 	}
 	
+	@Disabled
```
john-hill: Re-able and try again?

john-hill (changes requested):

integration-test/src/test/java/org/sagebionetworks/ITEDucTemplates.java
brucehoff: The service is only available to admin and ACT.  It's much easier to set up an admin' user than an ACT user in an integration test and anyway we test the auth' logic in detail at the manager level. This test just makes sure the round trip works.

lib/lib-docusign/src/main/java/org/sagebionetworks/docusign/DocuSignAccessTokenProvider.java
```diff
@@ -0,0 +1,55 @@
...
+ * Thread-safe cache for a DocuSign access token. The token is lazily fetched on
+ * first access and refreshed when it expires or is explicitly invalidated.
+ */
+class DocuSignAccessTokenProvider {
+
+	private final Supplier<TokenResult> tokenSupplier;
+	private final long expiryBufferMillis;
+
+	private String cachedAccessToken;
+	private long cachedAccessTokenExpiryMillis;
```
john-hill, line 31: I was still expected these to be volatile?

lib/lib-docusign/src/main/java/org/sagebionetworks/docusign/DocuSignClient.java
```diff
@@ -39,19 +32,28 @@ public class DocuSignClient {
...
 	public DocuSignClient(DocuSignClientConfig config) {
 		this(config, new DefaultTemplatesApiFactory());
 	}
 
 	DocuSignClient(DocuSignClientConfig config, TemplatesApiFactory templatesApiFactory) {
+		this(config, templatesApiFactory, null);
+	}
+
+	DocuSignClient(DocuSignClientConfig config, TemplatesApiFactory templatesApiFactory,
+			DocuSignAccessTokenProvider accessTokenProvider) {
```
john-hill, line 28: The DocuSignAccessTokenProvider  should be a bean.  Then Mockito just works.

lib/lib-docusign/src/test/java/org/sagebionetworks/docusign/DocuSignClientTest.java
```diff
@@ -44,34 +35,37 @@ public class DocuSignClientTest {
 	private DocuSignClientConfig mockConfig;
 	@Mock
 	private TemplatesApiFactory mockTemplatesApiFactory;
-	@Mock
-	private TemplatesApi mockTemplatesApi;
 
 	private DocuSignClient client;
+	private DocuSignAccessTokenProvider accessTokenProvider;
```
john-hill: Will be a mock

lib/lib-docusign/src/test/java/org/sagebionetworks/docusign/DocuSignClientTest.java
```diff
@@ -110,23 +105,17 @@ public void testListTemplatesSuccess() throws Exception {
 		mapped2.setDescription("Data sharing agreement");
 		assertEquals(mapped2, page.getResults().get(1));
 
-		// Verify the pagination params reach the SDK as strings
-		ArgumentCaptor<TemplatesApi.ListTemplatesOptions> optionsCaptor =
-				ArgumentCaptor.forClass(TemplatesApi.ListTemplatesOptions.class);
-		verify(mockTemplatesApi).listTemplates(eq(ACCOUNT_ID), optionsCaptor.capture());
-		assertEquals("0", optionsCaptor.getValue().getStartPosition());
-		assertEquals("51", optionsCaptor.getValue().getCount());
+		verify(mockTemplatesApiFactory).listTemplates(BASE_PATH, ACCESS_TOKEN, ACCOUNT_ID, "0", "51");
```
john-hill: For each case the provider should either be used nor not (verifyNoiternaction)

john-hill (changes requested):
Main is the shading (is there a "canned" cache?

lib/lib-docusign/src/main/java/org/sagebionetworks/docusign/DocuSignAccessTokenProvider.java
brucehoff: Claude responds:

```                                                                            

No — they don't need to be volatile because all reads and writes 
are inside synchronized methods. The synchronized keyword 
provides both mutual exclusion and  memory visibility (the Java 
Memory Model guarantees that changes made inside a synchronized 
block are visible to any thread that subsequently enters a synchronized
 block on the same monitor).

You'd need volatile if you were doing the double-checked locking 
pattern (read outside synchronized, then lock to write) — which is 
what the original DocuSignClient code was doing. But since 
DocuSignAccessTokenProvider synchronizes on every access, 
volatile would be redundant.

```

lib/lib-docusign/src/test/java/org/sagebionetworks/docusign/DocuSignClientTest.java
brucehoff: We now have DocuSignAccessTokenProviderTest which tests that caching works.

lib/lib-docusign/src/main/java/org/sagebionetworks/docusign/DocuSignAccessTokenProvider.java
```diff
@@ -1,32 +1,47 @@
...
 	private long cachedAccessTokenExpiryMillis;
 
-	DocuSignAccessTokenProvider(Supplier<TokenResult> tokenSupplier, long expiryBufferMillis) {
-		this.tokenSupplier = tokenSupplier;
-		this.expiryBufferMillis = expiryBufferMillis;
+	DocuSignAccessTokenProvider(DocuSignClientConfig config) {
+		this.config = config;
 	}
 
 	synchronized String getAccessToken() {
```
john-hill, line 37: public for the public contract and comment about this is what is used to always* get a valid token.

john-hill (approved):

PR #5666, opened 2026-06-10
PLFM-8384: Allow dataset collection items to omit the version number to auto-default to the latest dataset version

allaway:
https://sagebionetworks.jira.com/browse/PLFM-8384

Implements the design agreed on in the ticket comments (Jan 2026): a dataset collection item that **omits** `versionNumber` always references the **latest** version of the dataset, while an item with a version stays pinned and only changes by explicit action of the collection owner. No new flag is added.

## Changes

- **`DatasetCollectionMetadataProvider`**: removed the validation requiring a non-null version number on collection items. Entity-ID, uniqueness, and datasets-only checks are unchanged. `Dataset` (file) items still require a version.
- **View index**: no change needed — `DatasetCollectionMetadataIndexProvider.getViewFilter` already passes the (now possibly null) version into `IdAndVersionFilter`, which resolves unversioned refs with `OBJECT_VERSION = CURRENT_VERSION`. New dataset versions are picked up by the existing replication / out-of-date detection, same as entity views.
- **`DownloadListManagerImpl`**: adding a collection's files to a download list joins on exact `(dataset id, version)` pairs, so unversioned items would have silently matched nothing. Added `resolveLatestVersions()` to batch-resolve unversioned items to each dataset's current version via `NodeDAO.getCurrentRevisionNumbers`. Items referencing datasets that no longer exist are skipped instead of failing the request (the deleted-version scenario from the ticket).
- **Schemas**: documented the new semantics on `EntityRef.versionNumber` and `DatasetCollection.items`.

## Notes for reviewers

- Snapshot semantics: a snapshot of a collection freezes the view data as usual; the stored item list retains null versions, so the head version of the collection keeps floating to the latest dataset versions while snapshots remain frozen. Flagging for explicit sign-off since it is user-visible.
- The stored items JSON round-trips null versions without migration impact (no DB schema change).

## Tests

- Updated `DatasetCollectionMetadataProviderTest` (null version now valid), added `DatasetCollectionMetadataIndexProviderTest#testGetViewFilterWithNullItemVersion`, added download-list resolution tests incl. the deleted-dataset case. All unit tests pass locally (134 across the three classes).
- Added `DatasetCollectionIntegrationTest#testCreateAndQueryDatasetCollectionWithItemWithoutVersion`: creates a collection with an unversioned item, verifies it shows the dataset's current version, snapshots the dataset, and verifies the collection automatically reflects the new version. Compiles locally; needs a dev stack to run, so relying on CI for execution.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


allaway:
## Moving this discussion from Slack onto the PR

@john-hill gave me detailed feedback in Slack. Summarizing it here so the context lives with the PR rather than in a DM thread, then responding point by point.

Caveat up front: I'm new to this part of the codebase, so treat the findings below as claims to check rather than conclusions. I've cited the specific methods so they're easy to verify.

---

### Summary of John's feedback

The core concern: Synapse is a mature ecosystem, and there is a deep architectural assumption that every item in a dataset collection has a version number. Relaxing that assumption ripples across multiple systems, so some due diligence is needed before this can be reviewed and merged. He laid out three areas:

**1. Product and design definitions.** Product Management and/or Design need to weigh in on how this should actually behave:
- *Snapshotting* — if a user snapshots a dataset collection, does that freeze the item versions as of the moment the snapshot is created?
- *Query results* — when viewing the current version of a collection with unversioned items, should results always dynamically show the latest version of those items?
- *Downloads* — when a collection with unversioned items is added to a download list, which version of the file actually gets downloaded?
- *Mixed collections* — can a collection hold both versioned and unversioned items? If so, how does that affect queries, downloads, and snapshots?

**2. Technical impact and testing.**
- *Blast radius* — verify that a web or CLI client encountering a null version doesn't throw an NPE or break the UI.
- *Automated tests* — once the definitions above are settled, the PR needs integration tests covering each scenario, specifically showing how the outcome differs from a strictly-versioned collection.

**3. Deployment pipeline and review.**
- AWS developer account access so I can run the full PLFM suite, and whitelisting my fork so it can trigger builds.
- A synchronous code review with a PLFM developer once there's a clean, passing build.

---

### Responses

**Query results — already supported; no new code needed.**

This one turns out to be a non-question at the query layer. `IdAndVersionFilter` was already written for optional versions. When it builds the SQL for a collection's scope, it splits the items into two groups and emits different conditions for each:

```sql
((R.OBJECT_ID, R.OBJECT_VERSION) IN (:versionedRefs))
OR (R.OBJECT_ID IN (:nonVersionedRefs) AND R.OBJECT_VERSION = R.CURRENT_VERSION)
```

In plain terms: items pinned to a version match that exact version; items with no version match whatever the current version is. `DatasetCollectionMetadataIndexProvider.getViewFilter` builds its scope through `KeyFactory.idAndVersion(entityId, versionNumber)`, which only requires the entity ID and carries the version through as an `Optional`. So a null version already means "latest" all the way down to the SQL.

**Mixed collections — same answer.**

That `OR` in the SQL above *is* the mixed case. Versioned and unversioned items in one collection resolve correctly through the existing filter without additional work. Whether mixed collections are *desirable* is still a product question, but there's no technical obstacle to find.

**Snapshotting — frozen today, but for an incidental reason. This is the part I'd most like your read on.**

Snapshots do freeze, because `createSnapshot` writes the fully-built index out to S3 and `restoreTableIndexFromS3` reads those rows back rather than recomputing them. The frozen rows are a stored artifact.

But the *intent* isn't recorded anywhere, and here's what makes me uneasy: `NodeDAOImpl.getNodeItems(Long)` builds its lookup key with the node ID and no version, so it always reads the **current** revision's item list. Every caller drops the version too (`TableViewManagerImpl.validateViewForSnapshot`, `TableIndexManagerImpl.populateViewFromEntityReplication`, `ReplicationManagerImpl`). The scope of a dataset collection is never version-aware.

Until now that didn't matter, because null versions didn't exist and snapshots restore from S3 instead of rebuilding. With nulls allowed, any path that rebuilds a snapshot's index *from its scope* would resolve to whatever is latest now, not what was latest at snapshot time. I looked and could not find such a path for versioned views, so I'm not claiming a live bug. But it's the assumption holding up "snapshots freeze," it's undocumented, and nothing tests it. I suspect this is what your snapshot question was really pointing at.

**Downloads — this is the one place the PR makes a product decision, and it should be an explicit choice rather than an implementation detail.**

Worth noting the current behavior is worse than an NPE. The download SQL joins on `D_FILES.REVISION_NUMBER = D_FILES_REF.version`; a null version matches no rows, so unversioned items would silently disappear from the download with no error at all.

The PR adds `resolveLatestVersions` in `DownloadListManagerImpl`, which resolves nulls to the current version **at the moment the item is added to the download list**. So the semantics are "latest as of when you clicked add," not "latest as of when you download." That's defensible and it makes the download list stable once built, but it's a real decision and I'd rather have it agreed than inferred from my code. Items whose dataset no longer exists are dropped.

**Blast radius — narrower than the shared-schema change makes it look.**

The `EntityRef` schema is shared between Dataset and DatasetCollection, so relaxing it there sounds alarming. But enforcement lives in the per-type validators, and `DatasetMetadataProvider.validateEntity` still hard-fails on a null version for dataset items. Only `DatasetCollectionMetadataProvider` permits null. The old contract is intact for Datasets.

The checksum/size/count path is also unaffected — `NodeDAO.getFileSummary` is only ever called from `DatasetMetadataProvider`, never for collections.

What I **cannot** verify from this repo is your actual question: whether the web portal and CLI handle a null version without crashing. That needs someone with context on those repos, or pointers on where to look.

**Testing.**

There's an integration test for the download path in the PR now, but I can't run it without the AWS dev account. Only the unit tests have run (134 passing across `DownloadListManagerImplTest`, `DatasetCollectionMetadataProviderTest`, `DatasetCollectionMetadataIndexProviderTest`). I'd rather write the remaining scenario tests *after* the product definitions land, so they encode the agreed behavior instead of just pinning whatever I happened to implement.

**Access and next steps.**

Yes please to the AWS account and fork whitelisting whenever convenient. Happy to lead or at least facilitate the product conversation on snapshot/download/mixed-collection semantics — just point me at the right people.

This branch is now merged up to current `develop` and building clean. Still marked draft; I'll flip it to ready once the definitions are settled and the integration tests reflect them.

PR #5676, opened 2026-06-16
PLFM-9649 and PLFM-9650

brucehoff:
PLFM-9649: Extend AR object, adding eDUC template id
PLFM-9650: Extend DAR object, adding new fields


SandhraSokhal (approved):
LGTM

PR #5686, opened 2026-06-23
PLFM-9732

nickgros:
- Expand RecordSetSchemaResolver to include columns defined by the JSON Schema even if they are absent from the CSV
- Refactor grid validation logic into a new GridRowValidator class
- Lightly refactor RecordSetSchemaExporter to simplify later reuse

### DIFF
--- a/services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/table/RecordSetSchemaResolver.java
+++ b/services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/table/RecordSetSchemaResolver.java
@@ -8,6 +8,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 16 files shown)


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/table/RecordSetSchemaResolver.java
```diff
@@ -90,9 +94,10 @@ public List<Integer> getRequiredColumnIndices() {
 	 */
 	public ReconciledSchema getReconciledSchema(String entityId, FileHandle fileHandle,
```
nickgros, line 95: This is currently used by utilities to create the Grid from a RecordSet and to create a column schema for the RecordSet in the table index.

For the create scenario, I think adding these extra columns is desirable (similar to the original sync case).

For the table index scenario, the JSON-Schema-only columns will be empty, but I don't think it is so harmful that we need to provide an opt-out.

john-hill (approved):
LGTM.

PR #5696, opened 2026-06-26
PLFM-9732 - Sync engine changes 

nickgros:
```mermaid
classDiagram
    class CopyItem {
        <<interface>>
        +wasChangedByUser() boolean
    }
    class SourceItem {
        <<interface>>
        +getKey() String
    }
    class SourceReader~S~ {
        <<interface>>
        +consume(key) Optional~S~
        +streamRemaining() Stream~S~
    }
    class SyncRules~C,S~ {
        <<interface>>
        +getKey(copyItem) String
        +matches(copyItem, sourceItem) boolean
        +isExcludedFromMatching(copyItem, key) boolean
        +wasDeletedByUser(sourceItem) boolean
    }
    class SyncOutcomeHandler~C,S~ {
        <<interface>>
        +onCopyAndSourceMatch(copyItem, sourceItem)
        +onCopyAndSourceConflict(copyItem, sourceItem)
        +onCopyOnlyItemAddedByUser(copyItem, key)
        +onCopyOnlyItemDeletedFromSource(copyItem)
        +onSourceOnlyItemDeletedByUserFromCopy(sourceItem)
        +onSourceOnlyItemAddedSinceLastSync(sourceItem)
    }
    class SynchronizationLogic {
        +synchronize(copyItems, source, rules, handler)
    }
    SourceReader ..> SourceItem : produces
    SyncRules ..> CopyItem : consumes
    SyncRules ..> SourceItem : consumes
    SyncOutcomeHandler ..> CopyItem : consumes
    SyncOutcomeHandler ..> SourceItem : consumes
    SynchronizationLogic ..> SourceReader : drives
    SynchronizationLogic ..> SyncRules : drives
    SynchronizationLogic ..> SyncOutcomeHandler : drives
```

PR #5705, opened 2026-06-30
PLFM-9732 - REPEATABLE_READ snapshots

nickgros:
Update the CopyHandlerProviderImpl to capture a disk-backed snapshot of the current grid rows in a REPEATABLE_READ transaction that is as brief as possible. This prevents issues where the grid data may change between queries against each page, which would cause a non-repeatable read.

This is related to but independent from #5696


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/grid/synch/GridReadSnapshotTransaction.java
```diff
@@ -0,0 +1,35 @@
...
+@Target({ ElementType.METHOD, ElementType.TYPE })
+@Retention(RetentionPolicy.RUNTIME)
+@Transactional(
+	transactionManager = "gridTransactionManager",
+	isolation = Isolation.REPEATABLE_READ,
+	propagation = Propagation.REQUIRED,
+	readOnly = true,
+	rollbackFor = Throwable.class
+)
+public @interface GridReadSnapshotTransaction {
```
john-hill: GridRepeatableRead

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/grid/synch/handler/CopyHandlerProvider.java
```diff
@@ -11,13 +13,19 @@
...
+	 * try-with-resources), which releases the snapshot temp file.
 	 *
 	 * @param session the grid session containing connection information and user
 	 *                context for accessing the copy
-	 * @return a CopyHandler instance for reading from the specified copy
+	 * @return a CopyHandler instance ready for synchronization
+	 * @throws IOException if the row snapshot cannot be captured
 	 */
-	CopyHandler createCopyHandler(GridSession session);
+	CopyHandler createCopyHandler(GridSession session) throws IOException;
```
john-hill: If the caller is expected to do something special here then we can throw it, else wrap in a Runitme()

john-hill (changes requested):

PR #5707, opened 2026-07-01
PLFM-9652

brucehoff:
Adds methods to DocusignClient to fill in a document and route it for signing.

### DIFF
--- a/lib/lib-docusign/pom.xml
+++ b/lib/lib-docusign/pom.xml
@@ -29,6 +29,10 @@
 		</dependency>
 		<dependency>
 			<groupId>org.sagebionetworks</groupId>
+			<artifactId>lib-utils</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.sagebionetworks</groupId>
 			<artifactId>lib-models</artifactId>
 		</dependency>
 		<dependency>
--- a/lib/lib-docusign/src/main/java/org/sagebionetworks/docusign/DocuSignClient.java
+++ b/lib/lib-docusign/src/main/java/org/sagebionetworks/docusign/DocuSignClient.java
@@ -5,20 +5,27 @@ import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Function;
 
 import org.sagebionetworks.repo.model.educ.EDucTemplate;
 import org.sagebionetworks.repo.model.educ.EDucTemplatePage;
 import org.sagebionetworks.util.ValidateArgument;
 import org.springframework.stereotype.Service;
 
+import org.sagebionetworks.docusign.DocuSignTemplateValidator.TabType;
+
 import com.docusign.esign.client.ApiException;
+import com.docusign.esign.model.Email;
 import com.docusign.esign.model.EnvelopeDefinition;
 import com.docusign.esign.model.EnvelopeSummary;
 import com.docusign.esign.model.EnvelopeTemplate;
 import com.docusign.esign.model.EnvelopeTemplateResults;
+import com.docusign.esign.model.FullName;
+import com.docusign.esign.model.Signer;
 import com.docusign.esign.model.Tabs;
 import com.docusign.esign.model.TemplateRole;
 import com.docusign.esign.model.Text;
+import com.docusign.esign.model.Title;
 
 /**
  * Client for the DocuSign REST API. Authenticates headlessly via the JWT Bearer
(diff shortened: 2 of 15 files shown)


lib/lib-docusign/src/test/java/org/sagebionetworks/docusign/DocuSignTemplateValidatorTest.java
```diff
@@ -0,0 +1,157 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import com.docusign.esign.model.EnvelopeTemplate;
+import com.docusign.esign.model.Recipients;
+import com.docusign.esign.model.Signer;
+import com.docusign.esign.model.Text;
+
+public class DocuSignTemplateValidatorTest {
+
```
john-hill, line 18: Can we add more test coverage?

lib/lib-docusign/src/main/java/org/sagebionetworks/docusign/DocuSignClient.java
```diff
@@ -40,13 +57,206 @@ public class DocuSignClient {
...
+	 */
+	public String createAndSendEnvelope(String templateId, Map<String, String> roleEmails,
+			Map<RoleTabKey, String> tabValues) {
+		ValidateArgument.required(templateId, "templateId");
+		ValidateArgument.required(roleEmails, "roleEmails");
+		ValidateArgument.required(tabValues, "tabValues");
+		return executeWithRetry(token -> createAndSendEnvelopeOnce(templateId, roleEmails, tabValues, token));
+	}
+
+	private <T> T executeWithRetry(Function<String, T> action) {
```
john-hill: Would it help to create a new layer that does the retry and token refresh for you.  This layer can then just be about correctly preparing requests and translating the results?

lib/lib-docusign/src/main/java/org/sagebionetworks/docusign/DocuSignClient.java
```diff
@@ -40,13 +57,206 @@ public class DocuSignClient {
...
+						t.setValue(value);
+						titleTabs.add(t);
+					}
+					case EMAIL -> {
+						Email e = new Email();
+						e.setTabLabel(tabLabel);
+						e.setValue(value);
+						emailTabs.add(e);
+					}
+					default -> {
```
john-hill: I would throw on default (ensure you are not guessing).

lib/lib-docusign/pom.xml
```diff
@@ -27,6 +27,10 @@
 			<groupId>org.sagebionetworks</groupId>
 			<artifactId>lib-auto-generated</artifactId>
 		</dependency>
+		<dependency>
```
john-hill, line 30: Valid you still need this.

john-hill (changes requested):

lib/lib-docusign/pom.xml
brucehoff: Yes — DocuSignClient still uses ValidateArgument.required() for input validation, so the lib-utils dependency is needed.

lib/lib-docusign/src/main/java/org/sagebionetworks/docusign/DocuSignClient.java
brucehoff: I refactored the code to be OO, eliminating the `switch` statements and any `default` clauses.

lib/lib-docusign/src/main/java/org/sagebionetworks/docusign/DocuSignClient.java
brucehoff: Yes, done.  The retry logic is now removed from DocuSignClient and instead in the DocuSign*ApiImpl classes.  This achieves the requested separation of business logic and authentication/retry logic.

lib/lib-docusign/src/test/java/org/sagebionetworks/docusign/DocuSignTemplateValidatorTest.java
brucehoff: I reviewed test coverage using the Eclipse Coverage tool you showed me.  It reports 78% coverage.  I reviewed each class in lib-docusign and verified that the non-covered code is limited to thin wrappers around calls to the Docusign APIs (i.e., difficult to unit test and no real business logic.).

brucehoff:
Changes since last CR:
https://github.com/Sage-Bionetworks/Synapse-Repository-Services/pull/5707/changes/b5b9245d7b3e3508c6b984945bdaebc631cf5027

- refactored 'switch' statements to be OO in new TabType class;

- refactored retry logic from DocuSignClient into lower level classes (DocuSignEnvelopesApiImpl and DocuSignTemplatesApiImpl) to separate business logic and avoid retrying unnecessarily.

- instead of DocuSignClient.buildTabTypeMap() which determine tab type of each field from a template, we can make this determination based on static information in DocuSignTemplateValidator (see DocuSignTemplateValidator.typeforRoleAndLabel).

- In DocuSign 'EMAIL' means a read-only value while 'EMAIL_ADDRESS' is a r/w value.  We want the latter so we updated our code accordingly.

lib/lib-docusign/src/test/java/org/sagebionetworks/docusign/DocuSignTemplateValidatorTest.java
```diff
@@ -154,4 +155,46 @@ public void testValidateWithEmailInEmailAddressTabs() {
 		// call under test
 		assertDoesNotThrow(() -> DocuSignTemplateValidator.validate(template));
 	}
+
+	@Test
```
john-hill, line 159: Would these lend to a parameterized test?

john-hill (approved):
LGTM!

PR #5716, opened 2026-07-11
PLFM-9654

brucehoff:
This PR implements the service which retrieves the status for a Docusign eDUC which has been sent out for signatures.

The code essentially maps the status from the DocuSign API's response to a Synapse service response.
Since an data access request (DAR) can have only one routed eDUC, the Synapse request uses the DAR's id, and looks up the Docusign envelope id to use in the Docusign API request.
Only the requester (or a Syn admin') can check the eDUC's status.
The status includes the status of each signer and also the status of the overall eDUC.
The status of each signer includes their Synapse id (if available) for the client's reference.

The PR also has a change to the service to list the available Docusign templates: The wrong fields were used to get the `createdOn` and `modifiedOn` dates; this PR fixes that. See `DocuSignClient.toEDucTemplate()`


SandhraSokhal (approved):
LGTM

PR #5719, opened 2026-07-14
PLFM-9653

brucehoff:
PLFM-9653: Service to cancel an eDUC that's been routed for signing
PLFM-9655: Service to create a file handle ID for the signed document, as. pdf file

This also updates RequestManagerImpl to do two new checks when creating or updating an access request:
- Can't add a DUC pdf attachment if there is an envelope currently being signed. (OK if there is no envelope, since that means the user isn't using the eDUC feature.)
- Can only add file attachments if the user *owns* the file being attached.

The PR also renames some files for consistency:
duc/ folder -> educ/ folder
SignatureQuota -> EDucSignatureQuota
DucSignatureStatus -> EDucSignatureStatus
DucSignerStatus -> EDucSignerStatus
DucSignerStatusEnum -> EDucSignerStatusEnum
DucStatusEnum -> EDucStatusEnum

DocuSignAccessTokenProviderIntegrationTest: Exercises the process of authenticating to Docusign.  Previously the only test that did this was an integration test, requiring the Synapse back-end to be started with Tomcat.  This test makes it easier to check credentials.  It's skipped if the environment doesn't have `docusignEnabled=true`.


lib/lib-docusign/src/test/java/org/sagebionetworks/docusign/DocuSignAccessTokenProviderIntegrationTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+	@Test
+	public void testGetAccessTokenReturnsCachedToken() {
+		String first = provider.getAccessToken();
+
+		// call under test
+		String second = provider.getAccessToken();
+
+		assertNotNull(first);
+		assertNotNull(second);
```
SandhraSokhal, line 42: we should check first and second token are same

lib/lib-docusign/src/test/java/org/sagebionetworks/docusign/DocuSignAccessTokenProviderIntegrationTest.java
```diff
@@ -0,0 +1,56 @@
...
+	public void testInvalidateAndRefresh() {
+		String first = provider.getAccessToken();
+		provider.invalidateAccessToken();
+
+		// call under test
+		String second = provider.getAccessToken();
+
+		assertNotNull(first);
+		assertNotNull(second);
+	}
```
SandhraSokhal, line 55: we can check the token are different

SandhraSokhal (approved):
LGTM

PR #5725, opened 2026-07-18
PLFM-9651

brucehoff:
New service to preview an eDUC before routing for signature.

Previously we added a service to route a document for signature which combined Docusign's steps of creating an envelope and sending it out into one step.  This PR refactors the code so the two steps can be done separately.  "Previewing" then is the step of creating (but not routing an envelope) and retrieving a pdf version for the user to see.  

If an envelope has been created already then the 'signature' service uses the envelope.

Most of the changes are in `EDucManager`:  The logic in `routeForSignature` for creating and filling the template is factored into `createDraftEDuc` but the quota checks stay in place, since it's the routing of the envelope (not just creating it) which we get charged for.  We then add a new method `previewEDuc` which calls `createDraftEDuc` (which is idempotent) and returns the envelope as a pdf.


SandhraSokhal (approved):
LGTM

PR #5726, opened 2026-07-18
PLFM-9792

brucehoff:
Service to validate a Docusign template.

Synapse considers a Docusign template to be "valid" only if it has certain required fields, and checks validity before sending out a Docusign envelope.   

This PR adds validation in two other places: (1) A template is validated before it's attached to an access requirement (create or update); (2) It adds a service that allows a Governance team member to check validity before attempting to use it with an access requirement.


SandhraSokhal (approved):
LGTM

PR #5734, opened 2026-07-23
PLFM-7275

brucehoff:
[PLFM-8198](https://sagebionetworks.jira.com/browse/PLFM-8198) added special code to 'sync up' "alias" records with OIDC records, in the case that an alias corresponded to an OIDC-based identity.  The only such cases are (1) Google OIDC, in which the user's (unique, immutable) Gmail address is their alias, and (2) ORCiD, in which the user's (unique, immutable) ORCiD is their alias.  Additional OIDC providers, like NIH RAS, have no such alias and so the sync' up is unnecessary.  Moreover, in the case of these additional IDs the code erroneously clears out the OIDC record.

This PR simply adds a condition that the 'sync up' logic only applies to Google and ORCiD OAuth providers (since they are the only OAuth providers associated with a Synapse "alias").

[PLFM-8198]: https://sagebionetworks.jira.com/browse/PLFM-8198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jay-hodgson (approved):

services/repository-managers/src/test/java/org/sagebionetworks/repo/service/auth/AuthenticationServiceImplTest.java
```diff
@@ -357,6 +352,40 @@ public void testValidateOAuthAuthenticationCodeWithNoBoundAliasAndNoAliasFound()
...
+		when(mockAuthenticationManager.loginWithNoPasswordCheck(anyLong(), any())).thenReturn(authMgrLoginResponse);
+		
+		//call under test
+		LoginResponse result = service.validateOAuthAuthenticationCodeAndLogin(request, ISSUER);
+		
+		assertEquals(authMgrLoginResponse, result);
+		
+		verify(mockOAuthManager).validateUserWithProvider(request.getProvider(), request.getAuthenticationCode(), request.getRedirectUrl());
+		verify(mockUserManager).lookupOidcBindingBySubject(request.getProvider(), info.getSubject());
+		verifyNoMoreInteractions(mockUserManager);
```
jay-hodgson, line 385: Yay, no `mockUserManager.deleteOidcBinding()`!

jay-hodgson (approved):

PR #5746, opened 2026-07-27
PLFM-9848 move dueDate from TaskStatus to CurationTaskProperty

afwillia:


lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/curation/TaskStatus.json
```diff
@@ -22,11 +22,6 @@
       "format": "date-time",
       "description": "Timestamp of when the status was last updated."
     },
-    "dueDate": {
-      "type": "string",
-      "format": "date-time",
-      "description": "Timestamp by which the task should be completed. Null if no due date has been set."
-    },
```
nickgros, line 29: If we are removing this from the model, please ensure that

- The Python client is not currently using this field
- We can update the web clients to migrate to the new field, and the web client changes can be synchronized with the release of this change.

If either cause an issue, we may want to keep this here and mark it as deprecated, and document to use the correct field

lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/curation/CurationTaskProperties.json
```diff
@@ -5,6 +5,11 @@
     "concreteType": {
       "type": "string",
       "description": "Indicates which implementation of CurationTaskProperties this object represents. Possible values include: 'org.sagebionetworks.repo.model.curation.metadata.FileBasedMetadataTaskProperties' and 'org.sagebionetworks.repo.model.curation.metadata.RecordBasedMetadataTaskProperties'."
+    },
+    "dueDate": {
+      "type": "string",
+      "format": "date-time",
+      "description": "Timestamp by which the task should be completed. Null if no due date has been set."
```
nickgros: Shouldn't this be in the `CurationTask` object?

nickgros (changes requested):

lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/curation/TaskStatus.json
afwillia: Checked and Bryan confirmed that the python client doesn't support due date for curation tasks yet. I have SRC code ready to make the switch. So it seems safe to remove here.

lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/curation/CurationTaskProperties.json
afwillia: Yes, nice catch

nickgros (commented):
LGTM, but before I approve, does this need to be in the release-600 branch (the next two BDF demos will occur on v600)?

afwillia:
@nickgros let's wait to merge this after stack 600. The date filter is not needed for our next demo and requires some SRC rework

nickgros (approved):
@afwillia great work! you can merge into `develop` when you are ready

PR #5763, opened 2026-08-07
PLFM-9658 get quota and reset quota service

SandhraSokhal:


services/repository/src/main/java/org/sagebionetworks/repo/web/controller/AccessRequirementController.java
```diff
@@ -372,4 +373,25 @@ public void deleteAccessRequirementAcl(
 			@PathVariable String requirementId) {
 		return serviceProvider.getAccessRequirementService().getPermissions(userId, requirementId);
 	}
+
```
brucehoff, line 376: I believe we have an http filter that requires URIs starting with `/admin` to come from an internal IP address.  So this service's URI should start with `/admin`.

services/repository/src/main/java/org/sagebionetworks/repo/web/UrlHelpers.java
```diff
@@ -612,6 +612,7 @@ public class UrlHelpers {
 	public static final String ACCESS_REQUIREMENT_WITH_REQUIREMENT_ID_SUBJECTS = ACCESS_REQUIREMENT_WITH_REQUIREMENT_ID+"/subjects";
 
 	public static final String ACCESS_REQUIREMENT_VERSION = ACCESS_REQUIREMENT_WITH_REQUIREMENT_ID + "/version";
+	public static final String ACCESS_REQUIREMENT_EDUC_QUOTA_RESET = ACCESS_REQUIREMENT_WITH_REQUIREMENT_ID + "/eDucQuota/reset";
```
brucehoff: URI should start with `/admin`

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/docusign/EDucManager.java
```diff
@@ -152,6 +152,60 @@ public EDucSignatureQuota routeForSignature(UserInfo userInfo, String requestId)
...
+		ValidateArgument.required(userInfo, "userInfo");
+		ValidateArgument.required(requestId, "requestId");
+
+		RequestInterface request = requestDao.get(requestId);
+
+		if (!AuthorizationUtils.isUserCreatorOrAdmin(userInfo, request.getCreatedBy())) {
+			throw new UnauthorizedException("Only the request creator or an administrator can view the signature quota.");
+		}
+
+		Long userId = userInfo.getId();
```
brucehoff: shouldn't this be request.getCreatedBy() ?

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/docusign/EDucManager.java
```diff
@@ -152,6 +152,60 @@ public EDucSignatureQuota routeForSignature(UserInfo userInfo, String requestId)
...
+		accessRequirementDao.get(accessRequirementId);
+
+		Long arId = Long.parseLong(accessRequirementId);
+
+		long nowMs = clock.currentTimeMillis();
+		long thirtyDaysAgoMs = nowMs - THIRTY_DAYS_IN_MS;
+
+		long count = eDucQuotaDao.getCount(userId, arId, thirtyDaysAgoMs, nowMs);
+
+		if (count >= MAX_ENVELOPES_PER_MONTH) {
```
brucehoff: why add this logic, rather than simply resetting the quota?  That is, an administrator might have some rationale we do not anticipate, for resetting a quota that has not yet been reached, so why add this extra logic?

brucehoff (changes requested):

services/repository-managers/src/main/java/org/sagebionetworks/repo/service/docusign/EDucService.java
```diff
@@ -44,7 +44,8 @@ public EDucSignatureQuota getSignatureQuota(Long userId, String requestId) {
 
 	public EDucSignatureQuota resetQuota(Long userId, String accessRequirementId, Long targetUserId) {
 		UserInfo userInfo = userManager.getUserInfo(userId);
-		return eDucManager.resetQuota(userInfo, accessRequirementId, targetUserId);
+		UserInfo targetUserInfo = userManager.getUserInfo(targetUserId);
+		return eDucManager.resetQuota(userInfo, accessRequirementId, targetUserInfo.getId());
```
brucehoff, line 48: I don't understand this change:  Why is `targetUserInfo.getId()` different from `targetUserId`?  Aren't they the same?

brucehoff (changes requested):

services/repository-managers/src/main/java/org/sagebionetworks/repo/service/docusign/EDucService.java
SandhraSokhal: We validate user by calling userManager.getUserInfo(Id) in service layer. To validate that target user is valid user I added this.  The caller of API is admin and admin is resetting the quota for target user. Both user should be Synapse user.

brucehoff: Ah, OK.

brucehoff (approved):
good work!

PR #5765, opened 2026-08-08
PLFM-9853 - Handle case where CSV data may not match column type

nickgros:
## Problem

RecordSet grid `PULL` sync (`RecordSetSourceHandler`) re-reads two independent CSV revisions through the schema inferred from the latest revision:

1.  the *latest* revision (source of the current schema)
2. the *previously synced baseline* revision

When the baseline holds a value that schema can't represent (e.g. an alphanumeric value in a column the latest revision inferred as `INTEGER`), strict translation threw `NumberFormatException` and aborted the whole sync.

## Fix

- **`Translator.translateLeniently(String, boolean)`** (new default method): translates as `ConType.STRING` instead of throwing when a value doesn't fit the column's type. A blank value is still normalized to no-value for types that reject blank (INTEGER/DOUBLE/DATE/JSON); types that accept any string translate blank exactly as the strict path does. The `catch` is scoped to `IllegalArgumentException | JSONException` — the only exceptions any current `Translator` implementation actually throws — so a genuine translator defect (e.g. an NPE) still propagates instead of silently becoming a text cell.
- **`RecordSetSourceHandler.createSynchRow`** now routes non-key column values through `translateLeniently` instead of `translateNullable`, so a baseline-only type mismatch is carried through as text and the sync completes.
- **`RecordSetSourceHandler` constructor** wraps `initialize()` in a try/catch that closes (and deletes) the temp file it creates before rethrowing — `initialize()` runs from the constructor, so a failure there previously leaked the temp file on every retry, since the handler is never returned to a try-with-resources.
- **`CsvDataStream`** (separate `GridCsvImportRequest` import path, same bug family): a blank cell in a non-key `INTEGER` column previously threw and failed the whole import. Non-key columns now translate leniently too; upsert-key columns stay strict because they back a typed `NOT NULL` temp-table column used to join against the grid's temp
  table.

### DIFF
--- a/services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/grid/row/translator/TranslatorTest.java
+++ b/services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/grid/row/translator/TranslatorTest.java
@@ -1,6 +1,7 @@
 package org.sagebionetworks.repo.manager.grid.row.translator;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 
 import org.junit.jupiter.api.Test;
 import org.sagebionetworks.repo.model.grid.patch.ConType;
@@ -94,7 +95,7 @@ public class TranslatorTest {
 	}
 
 	@Test
-	public void testTranslateLenientlyOneArgOverloadDefaultsToRequired() {
+	public void testTranslateLenientlyWithNoIsRequiredPropertyArgument() {
 		// call under test
 		ConValue con = new LongTranslator().translateLeniently("JH-2-009-518B9-A_1");
 		assertEquals(new ConValue(ConType.STRING, "JH-2-009-518B9-A_1"), con);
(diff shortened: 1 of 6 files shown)


services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/grid/row/translator/TranslatorTest.java
```diff
@@ -0,0 +1,141 @@
...
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import org.junit.jupiter.api.Test;
+import org.sagebionetworks.repo.model.grid.patch.ConType;
+import org.sagebionetworks.repo.model.grid.patch.ConValue;
+
+/**
+ * Tests the {@link Translator#translateLeniently} default methods, exercised
+ * through the real {@link Translator} implementations (no mocking: these are
+ * plain value objects, not system boundaries).
```
nickgros:
```suggestion
 * plain value objects).
```

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/grid/synch/handler/RecordSetSourceHandlerTest.java
```diff
@@ -656,5 +689,105 @@ public void testGetBenefactorIds() throws IOException {
...
+		String fallbackKey = UpsertKeyEncoder
+				.encode(List.of(new ConValue(ConType.STRING, "JH-2-009-518B9-A_1")));
+		assertTrue(handler.wasInSyncedBaseline(fallbackKey));
+	}
+
+	/**
+	 * A row unchanged across both revisions, including its type-incompatible
+	 * value, must still be reported as unchanged — the leniency fallback must not
+	 * perturb the row's content hash.
+	 */
```
nickgros, line 720:
```suggestion
	/**
	 * A row unchanged across both revisions, including its type-incompatible
	 * value, must still be reported as unchanged — the leniency fallback must not
	 * affect the row's content hash.
	 */
```

john-hill (approved):

PR #5766, opened 2026-08-10
PLFM-9858 - Fix grid cell misalignment for columns with no value

nickgros:
### Problem
`POST /grid/download/csv/async/start` writes CSV rows shorter than the header whenever a grid column has no value in a row (e.g. a column added to the grid after the row was created). The missing value is dropped entirely instead of written as an empty field, so every value after it shifts one column to the left. The same defect corrupts `POST /grid/session/query` results (values land under the wrong column name) and grid→grid sync (values written back into the wrong column).

Root cause: `GridReplicaViewManagerImpl.createRowViewMapper` read `SELECTED_VALS` (one JSON element per selected column, `null` where the row's sparse CRDT vector has no entry) into a `List`, skipping `null` elements instead of preserving their position. Every consumer then read a different column's value.

### Fix
Cells are now carried in a `Map<Integer, ConstantNode>` keyed by position in the query's selected columns. A column with no CRDT node is simply **absent** from the map, so each consumer decides explicitly what absence means:

| Consumer | Behavior |
|---|---|
| CSV export | empty field |
| `/grid/session/query` JSON doc | key omitted |
| Grid→grid sync | cell omitted from the copy (correct merge semantics) |
| CSV import upsert matching | treated like a NULL/UNDEFINED key (row skipped) |

Deliberate behavior change: an explicit JSON-null cell now exports as an empty CSV field instead of the literal text `"null"` (this also fixes a real corruption path in RecordSet sync, documented in the CSV exporter's javadoc).

### Changes
- `RowData`/`RowObject`/`RowView.getCells()` (`List<ConValue>`) → `getNodes()` (`Map<Integer, ConstantNode>`) + `getCell(int)`.
- `GridReplicaViewManagerImpl`: new `readSelectedValues` helper (the fix).
- `GridJsonUtils`: new map-keyed `gridRowToJsonObject`; renamed the old name-keyed overload to `gridCellsToJsonObject`.
- `GridReplicaCsvExporterImpl`, `GridDataStream`, `CopyHandlerImpl`,  `GridReplicaValidationManagerImpl`, `SnapshotRowHandler`: migrated to the map-keyed API.
- `GridIntegrationTestUtils` (workers test helper): fixed a latent bug where it looked up cells by CRDT vector index instead of header position.

### DIFF
--- a/services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/grid/internal/replica/merge/GridDataStreamTest.java
+++ b/services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/grid/internal/replica/merge/GridDataStreamTest.java
@@ -4,7 +4,6 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Map;
 
 import org.junit.Test;
 import org.sagebionetworks.repo.manager.grid.internal.replica.model.RowData;
@@ -31,11 +30,11 @@ public class GridDataStreamTest {
 				.setSequenceNumber(base.getSequenceNumber());
 		return new RowView().setRowObject(new RowObject().setData(
 			new RowData().setVectorId(LogicalTimestamp.newIncrement(vId, vecIncrement))
-				.setNodes(Map.of(
-					0, new ConstantNode().setId(LogicalTimestamp.newIncrement(vId, n1)).setValue(cell0),
-					1, new ConstantNode().setId(LogicalTimestamp.newIncrement(vId, n2)).setValue(cell1),
-					2, new ConstantNode().setId(LogicalTimestamp.newIncrement(vId, n3)).setValue(cell2)
-				))
+				.setNodes(new ConstantNode[] {
+					new ConstantNode().setId(LogicalTimestamp.newIncrement(vId, n1)).setValue(cell0),
+					new ConstantNode().setId(LogicalTimestamp.newIncrement(vId, n2)).setValue(cell1),
+					new ConstantNode().setId(LogicalTimestamp.newIncrement(vId, n3)).setValue(cell2)
+				})
 		));
 	}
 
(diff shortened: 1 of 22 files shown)


services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/grid/internal/replica/merge/GridDataStreamTest.java
```diff
@@ -47,26 +47,26 @@ public void testStream() {
...
+							2, new ConstantNode().setId(LogicalTimestamp.newIncrement(vectorId, 9)).setValue(new ConValue(ConType.LONG, 2L))
 					))
 			)),
 			new RowView().setRowObject(new RowObject().setData(
         		new RowData().setVectorId(LogicalTimestamp.newIncrement(vectorId, 3))
-        			.setNodes(Arrays.asList(
-							new ConstantNode().setId(LogicalTimestamp.newIncrement(vectorId, 10)).setValue(new ConValue(ConType.STRING, "3")),
-							new ConstantNode().setId(LogicalTimestamp.newIncrement(vectorId, 11)).setValue(new ConValue(ConType.STRING, "more3")),
-							new ConstantNode().setId(LogicalTimestamp.newIncrement(vectorId, 12)).setValue(new ConValue(ConType.LONG, 3L))
+        			.setNodes(Map.of(
```
nickgros:
```suggestion
					.setNodes(Map.of(
```

john-hill (changes requested):
Can we use ConValue[] array?

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/grid/internal/replica/merge/GridDataStreamTest.java
```diff
@@ -47,27 +46,27 @@ public void testStream() {
         List<RowView> rows = List.of(
 			new RowView().setRowObject(new RowObject().setData(
         		new RowData().setVectorId(LogicalTimestamp.newIncrement(vectorId, 1))
-        			.setNodes(Arrays.asList(
+        			.setNodes(new ConstantNode[] {
```
nickgros:
```suggestion
					.setNodes(new ConstantNode[] {
```

john-hill (approved):
LGTM!

PR #5776, opened 2026-08-14
There were two issues that caused this ticket's failure: the maxToken…

john-hill:
…s for the gridUpdateSpecialist was low (4096); and second, if a token limit was reached the supervisor was provided with a message containing anything the specialist has written before hitting the token limit.  As a result of the second issue, the supervisor concluded that the update was a success.  To address the issue we first raised the limit on the gridUpdateSpecialist (doubled it).  We also change all agents to implment a new interface.  This interfaces allows us to standardize all of the agents' chat requests and more importanly check for the token limit state, so we can provide the supervisor with a real message that it can use to break up large changes.  We also evaludated Curie's instructions and address a number of issues


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/agent/specialist/gridupdate/GridUpdateSpecialist.java
```diff
@@ -35,29 +34,16 @@ public class GridUpdateSpecialist {
 				.defaultAdvisors(MessageChatMemoryAdvisor.builder(memory).build())
 				.defaultOptions(BedrockChatOptions.builder()
 						.model(stackConfig.getModelIdClaudeSonnet())
-						.maxTokens(4096)
+						.maxTokens(8192)
```
nickgros, line 37: Does a lower `maxTokens` value limit the capabilities of a specialist? Does a larger `maxTokens` value pollute the supervisor's context window? Is there a way to separate the two? Ideally we can let the specialist generate many tokens, but limit the tokens that go into a summary that the supervisor receives.

nickgros (approved):

PR #5781, opened 2026-08-15
PLFM-9652

brucehoff:
There was a mistake in the eDUC design:  The 'institution' entered by the DAR creator is supposed to go into the 'institution' field for the *signing official*, not for the *principal investigator*.  This PR introduces two changes:
- the template validator looks for a `signing_official_institution` field instead of a `principal_investigator_institution` field in the template;
- when creating an eDUC, the `institution` entered into the DAR is now copied into the `signing_official_institution` field.


SandhraSokhal (approved):
LGTM

PR #5782, opened 2026-08-16
Added a turn limit enforced for all of the new agents.  Currently, th…

john-hill:
…e limit is set to 20.  If a supervisor attempts to make more than 20 tool calls the job will fail with a message indicating that the turn limit was reached.  This will prevent future cases where a supervisor unconditionally calls tools in an infinite loop.


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/agent/tool/TurnLimitAdvisor.java
```diff
@@ -0,0 +1,134 @@
...
+	public int getOrder() {
+		return ADVISOR_ORDER;
+	}
+
+	@Override
+	public ChatClientResponse adviseCall(ChatClientRequest chatClientRequest, CallAdvisorChain callAdvisorChain) {
+		Object counter = chatClientRequest.context().get(AgentToolContextKey.TURN_COUNT.getKey());
+		// Enforce only when a counter was seeded for this loop; used outside that path (e.g. a plain
+		// single-shot call) this advisor is a pass-through.
+		if (!(counter instanceof AtomicInteger turnCount)) {
```
xschildw, line 111: We should fail here.

xschildw (approved):

PR #5785, opened 2026-08-17
PLFM-9880 - CSV import is attributed to user

nickgros:
Fix issue where sync would revert changes made via CSV import.

This happened because sync relies on CRDT-level attribution to determine how to process each row. Since changes made via CSV import were attributed to the system, and not a user, sync would ignore/remove them.

Fixed by adding a new IMPORT EventSource that is created in the ID space allotted to user-replicas, so the changes are now correctly attributed to users. There is at most one dedicated import replica per (user, session) pair, and the replica is created at import-time.


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/grid/internal/replica/merge/GridCsvImporterImpl.java
```diff
@@ -75,11 +73,10 @@ public GridCsvImportResponse importCsv(UserInfo user, GridCsvImportRequest reque
 
 		GridHeader gridHeader = replicaSupport.getGridHeaderOrThrow(gridSession);
 		
-		// Gets the connection info for the publisher now so that we fail fast.
-		// Note: the USER_SUPPORT connection is created once per session for the session creator,
-		// and the import may be submitted by a different user (e.g. a team member). See PLFM-9571.
-		GridConnectionInfo publisherConnInfo = gridManager.getSingletonConnection(gridSession.getSessionId(), EventSource.USER_SUPPORT)
-			.orElseThrow(() -> new RecoverableMessageException("No internal connection found for session: " + gridSession.getSessionId()));
+		// Publish the imported changes under a replica owned by the importing user, so
+		// the imported cells carry user attribution (PLFM-9880)
```
nickgros:
```suggestion
		// Publish the imported changes under a replica owned by the importing user, so the
		// imported cells carry user attribution (PLFM-9880)
```

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/grid/GridDaoImpl.java
```diff
@@ -375,6 +375,17 @@ public Optional<GridConnectionInfo> getSingletonConnection(String sessionId, Eve
             return Optional.empty();
         }
     }
+
+    @Override
+    public Optional<GridConnectionInfo> getUserConnection(String sessionId, Long userId, EventSource source) {
```
john-hill, line 380: This is getLatestUserConnection() and should be limited to which event sources it can be used on.  Should we limit this to non-singletons and user source?

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/grid/GridManagerUnitTest.java
```diff
@@ -2468,6 +2468,58 @@ public void testUpdateGridWithConnectionCreationFailure() throws Exception {
...
+		ArgumentCaptor<GridConnectionInfo> connCaptor = ArgumentCaptor.forClass(GridConnectionInfo.class);
+		verify(mockGridDao).createConnection(connCaptor.capture());
+		assertEquals(EventSource.IMPORT, connCaptor.getValue().getSource());
+		assertEquals(gridSessionId, connCaptor.getValue().getSessionId());
+		assertEquals(replicaId, connCaptor.getValue().getReplicaId());
+		assertEquals(userId, connCaptor.getValue().getCreatedBy());
+	}
+
+	@ParameterizedTest
+	@EnumSource(value = EventSource.class, names = { "WEBSOCKET", "AGENT", "API", "IMPORT" }, mode = EnumSource.Mode.EXCLUDE)
```
john-hill, line 2512: add the inverse as well.

john-hill (approved):
There are some renames that we should make but we can do that later.

PR #5786, opened 2026-08-17
PLFM-9657

brucehoff:
Update an In-flight eDUC/Docusign document.

We add two new services:
- GET /dataAccessRequest/id/signature/precheck - to see if it's possible to apply the DAR to the envelope
- PUT /dataAccessRequest/id/signature - to update the in-flight envelope

Since a data access request (DAR) can be freely changed after the envelope is initially sent, this PR updates the signature *status* to return whether edits to the DAR have been applied to the envelope yet. (We add to the persisted DBORequest object a hash of the content at the time the envelope was sent, so we can tell if there are changes to apply to the envelope.)

Precheck: An envelope can't be updated only if it's in the wrong state, e.g., if it's completed or canceled.  The precheck performs this check and, if the envelope can't be updated it returns a message explaining why.

Update: After determining what changes to make, the DocsignClient does the following:
- Delete any recipients removed from the DAR (if they have not already signed);
- Update any recipients changed in the DAR (if they have not already signed);
- Add any new recipients added to the DAR;

To review the code:

The bulk of the logic changes are in DocuSignClient and EDucManager, so start with these two.  Also, in DocuSignClient the most complex logic is the logic of adding a new recipient that was originally removed from the envelope as extraneous. E.g., say the original envelope had 10 collaborators, removing 88 unneeded ones from the 98 in the template, but now we want to add one back in, and need to retrieve the 'tab' definitions for the restored signer from the template to add to the envelope.  DocuSignClient takes care of this.


integration-test/src/test/java/org/sagebionetworks/ITEDucSignatureTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+		actAR = new ACTAccessRequirement();
+		RestrictableObjectDescriptor rod = new RestrictableObjectDescriptor();
+		rod.setId(project.getId());
+		rod.setType(RestrictableObjectType.ENTITY);
+		actAR.setSubjectIds(Arrays.asList(rod));
+		actAR.setAccessType(ACCESS_TYPE.DOWNLOAD);
+		actAR = adminSynapse.createAccessRequirement(actAR);
+
+		AccessRequirementConversionRequest conversionRequest = new AccessRequirementConversionRequest();
```
brucehoff: Why use AccessRequirementConversionRequest?

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/docusign/EDucManager.java
```diff
@@ -257,18 +277,153 @@ RequestInterface createDraftEDuc(RequestInterface request) {
...
+			for (AccessorChange change : accessorChanges) {
+				if (AccessType.GAIN_ACCESS.equals(change.getType())
+						|| AccessType.RENEW_ACCESS.equals(change.getType())) {
+					builder.append("accessor=").append(change.getUserId())
+							.append(':').append(change.getType()).append('\n');
+				}
+			}
+		}
+
+		return DigestUtils.sha256Hex(builder.toString());
```
brucehoff: Note, this string encoding will be 64 bytes long

lib/lib-docusign/src/main/java/org/sagebionetworks/docusign/DocuSignClient.java
```diff
@@ -110,6 +112,161 @@ public void sendEnvelope(String envelopeId) {
...
+		int nextRecipientId = maxRecipientId;
+		for (Map.Entry<String, String> entry : roleEmails.entrySet()) {
+			String roleName = entry.getKey();
+			if (existingRoles.contains(roleName)) {
+				continue;
+			}
+			nextRecipientId++;
+			Signer signer = new Signer();
+			signer.setRecipientId(Integer.toString(nextRecipientId));
+			signer.setRoutingOrder(Integer.toString(nextRecipientId));
```
brucehoff: This is wrong: The routing order should ensure that collaborators sign before the signing official does.  But what happens if the signing official has already signed.  Will a lower/earlier routing order 'break' the envelope?

lib/lib-docusign/src/main/java/org/sagebionetworks/docusign/DocuSignClient.java
```diff
@@ -110,6 +112,161 @@ public void sendEnvelope(String envelopeId) {
...
+		// Place the envelope into the "correct" state, which pauses the signing process.
+		Envelope correcting = new Envelope();
+		correcting.setStatus("correct");
+		envelopesApi.updateEnvelope(envelopeId, correcting);
+
+		Recipients toDelete = buildRemovedRecipients(existingSigners, roleEmails);
+		Recipients toUpdate = buildUpdatedRecipients(existingSigners, roleEmails, tabValues);
+		Recipients toCreate = buildNewRecipients(existingSigners, roleEmails, tabValues);
+
+		if (hasSigners(toDelete)) {
```
brucehoff: In truth, only 'collaborators' will be added or deleted.  Principal Investigator and Signing Official can be changed but not added or deleted.  Knowing this we can ensure the correct routing order for added signers.

lib/lib-docusign/src/main/java/org/sagebionetworks/docusign/DocuSignClient.java
```diff
@@ -110,6 +112,161 @@ public void sendEnvelope(String envelopeId) {
...
+			}
+			nextRecipientId++;
+			Signer signer = new Signer();
+			signer.setRecipientId(Integer.toString(nextRecipientId));
+			signer.setRoutingOrder(Integer.toString(nextRecipientId));
+			signer.setRoleName(roleName);
+			signer.setEmail(entry.getValue());
+			Tabs tabs = new Tabs();
+			signer.setTabs(tabs);
+			String fullName = fillTabsForRole(roleName, tabs, tabValues);
```
brucehoff: If a collaborator is removed from the envelope in the initial routing and later added back in, will the values filled in here appear correctly?  That if the collaborator was removed earlier, did the envelope also lose the information about where the tabs go in the document?

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/dataaccess/DBORequestDAOImpl.java
```diff
@@ -196,6 +209,21 @@ public String getAccessRequirementId(String requestId) {
 		}
 	}
 
+	@WriteTransaction
+	@Override
+	public void setEDucContentHash(String requestId, String hash) {
+		jdbcTemplate.update(SQL_SET_CONTENT_HASH, hash, requestId);
```
brucehoff: Need to change the etag to trigger migration

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/docusign/EDucManager.java
```diff
@@ -54,6 +60,8 @@ public class EDucManager {
 	static final int MAX_ENVELOPES_PER_MONTH = 10;
 	static final long THIRTY_DAYS_IN_MS = 30L * 24 * 60 * 60 * 1000;
 	static final int MAX_GLOBAL_ENVELOPES_PER_DAY = 100;
+	// Prefix of the numbered DocuSign roles the template defines for the request's accessors.
+	static final String COLLABORATOR_ROLE_PREFIX = "collaborator_";
```
SandhraSokhal: Do we already have prefix strings somewhere

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/docusign/EDucManagerTest.java
```diff
@@ -913,6 +963,454 @@ public void testGetSignatureStatusWithNoEnvelope() {
...
+
+		// call under test
+		eDucManager.updateRoutedEnvelope(user, "req-1");
+
+		Map<String, RecipientInfo> corrected = captureCorrectedRecipients();
+		// 302 keeps collaborator_3 instead of shifting down into the role 301 vacated, so 302's
+		// signature is not reattributed and collaborator_2 is simply left to be removed
+		assertEquals("creator@example.com", corrected.get("collaborator_1").email());
+		assertEquals("c302@example.com", corrected.get("collaborator_3").email());
+		assertFalse(corrected.containsKey("collaborator_2"));
```
SandhraSokhal, line 1150: After adding max signers , remove n signer (signed/unsigned) and adding new n signers

lib/lib-docusign/src/test/java/org/sagebionetworks/docusign/DocuSignLiveTest.java
```diff
@@ -0,0 +1,585 @@
...
+	 * Which of the envelope's status and edit lock is standing in the way. Run this after a correction
+	 * fails with {@code EDIT_LOCK_NOT_LOCK_OWNER}: whether the status reached "correct" says where the
+	 * failure happened. Status "correct" means our own transition succeeded and a later recipient call
+	 * was refused, so the lock that transition created is one we do not present on subsequent calls.
+	 * Status still "sent" means the transition itself was refused, so something outside this process —
+	 * most likely the envelope open in the DocuSign web console — holds the lock.
+	 */
+	@Test
+	public void stepDiag_showEnvelopeAndLock() {
+		Envelope envelope = readEnvelopeWithTabs();
```
SandhraSokhal, line 256: should we deleted lock and delete lock test

lib/jdomodels/src/main/java/org/sagebionetworks/repo/model/dbo/dao/dataaccess/DBORequestDAOImpl.java
```diff
@@ -82,6 +83,14 @@ public class DBORequestDAOImpl implements RequestDAO {
 			+ " FROM " + TABLE_DATA_ACCESS_REQUEST
 			+ " WHERE " + COL_DATA_ACCESS_REQUEST_ID + " = ?";
 
+	public static final String SQL_GET_CONTENT_HASH = "SELECT " + COL_DATA_ACCESS_REQUEST_EDUC_CONTENT_HASH
+			+ " FROM " + TABLE_DATA_ACCESS_REQUEST
+			+ " WHERE " + COL_DATA_ACCESS_REQUEST_ID + " = ?";
+
+	public static final String SQL_SET_CONTENT_HASH = "UPDATE " + TABLE_DATA_ACCESS_REQUEST
```
SandhraSokhal, line 93: we should also update the etag

SandhraSokhal (changes requested):

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/docusign/EDucManager.java
brucehoff: Updated to remove duplicate string definitions!

lib/lib-docusign/src/test/java/org/sagebionetworks/docusign/DocuSignLiveTest.java
brucehoff: The three "stepDiag(nosis)" tests are left there intentionally. The JavaDoc says why.  So let's leave them in place:

```
 * The {@code stepDiag_*} methods are not part of the sequence. They report what is standing in the way
 * when a step fails: the envelope's status and edit lock, and what the template says its roles' tabs
 * are. {@code stepReset_voidAndClearState} abandons a run so another can start, and tolerates an
 * envelope too far along to void.
```

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/docusign/EDucManagerTest.java
brucehoff: Claude determined that the existing tests cover these edge cases, specifically:

```
In the case that the removed signer had signed → their index is retired, all 98 are now unavailable, and nextIndex walks to 99. The manager emits collaborator_99.

That case is the only thing capacity actually surfaces, and it's covered one layer down: DocuSignClientTest.testBuildNewRecipientsWithRoleMissingFromTemplate is exactly this condition in miniature — a template defining only collaborator_1 asked for collaborator_2 → IllegalArgumentException("The template does not define the role 'collaborator_2'."). A 98-role fixture would exercise the identical code path for a much higher setup cost.

Two things worth knowing while you're here:

MAX_COLLABORATORS isn't the operative cap. The real limit is however many collaborator_N roles the ACT author put in the template, which may be well under 98. That's why there's no 98-bound in EDucManager and why adding one would be the wrong check — the authoritative test is against the template, and buildNewRecipients already does it.

```

SandhraSokhal (approved):
LGTM

PR #5793, opened 2026-08-22
PLFM-9652

brucehoff:
This PR fixes two problems which today cause Docusign envelopest to fail when sent:
1) Each recipient must be filled in, with an email address:  The template has up to 100 recipients and generally not all will be used.  This PR removes the unused ones before sending out the envelope.
2) Every recipient must have a name:  Previously recipient names were only added to the envelope if available.  This PR ensures each recipient always has a name, otherwise raising an exception rather than trying to send the envelope.  Specifically, it will raise an exception of the PI name or the signing official name is missing.
The PR also ensures that collaborators always have names, by defaulting to their user name if no first name or last name is available.
The PR also forwards any error raised by Docusign when failing to send an envelope back to the Synapse client.


SandhraSokhal (approved):
LGTM

PR #5794, opened 2026-08-23
PLFM-9653

brucehoff:
- when retrieving an access request by access requirement ID, make sure the access requirement exists;
- prevent the client from arbitrarily changing the envelope ID in the access request;
- fixes a small bug in retrieving envelope status

### DIFF
--- a/lib/lib-docusign/src/main/java/org/sagebionetworks/docusign/DocuSignEnvelopesApiImpl.java
+++ b/lib/lib-docusign/src/main/java/org/sagebionetworks/docusign/DocuSignEnvelopesApiImpl.java
@@ -77,7 +77,11 @@ class DocuSignEnvelopesApiImpl implements DocuSignEnvelopesApi {
 			EnvelopesApi envelopesApi = new EnvelopesApi(apiClient);
 			EnvelopeIdsRequest request = new EnvelopeIdsRequest();
 			request.setEnvelopeIds(envelopeIds);
-			EnvelopesInformation info = envelopesApi.listStatus(config.getAccountId(), request);
+			// The status endpoint only reads the envelope ids from the request body when the
+			// envelope_ids query parameter is set to "request_body"; otherwise DocuSign returns 400.
+			EnvelopesApi.ListStatusOptions options = envelopesApi.new ListStatusOptions();
+			options.setEnvelopeIds("request_body");
+			EnvelopesInformation info = envelopesApi.listStatus(config.getAccountId(), request, options);
 			return info.getEnvelopes() != null ? info.getEnvelopes() : Collections.<Envelope>emptyList();
 		});
 	}
--- a/services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/dataaccess/RequestManagerImpl.java
+++ b/services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/dataaccess/RequestManagerImpl.java
@@ -143,6 +143,9 @@ public class RequestManagerImpl implements RequestManager{
 			throws NotFoundException {
 		ValidateArgument.required(userInfo, "userInfo");
 		ValidateArgument.required(accessRequirementId, "accessRequirementId");
+		// Verify the access requirement exists; a missing one must be a 404 rather than falling
+		// through to a blank new-request stub.
+		accessRequirementDao.get(accessRequirementId);
 		try {
 			return requestDao.getUserOwnCurrentRequest(accessRequirementId, userInfo.getId().toString());
 		} catch (NotFoundException e) {
@@ -202,7 +205,6 @@ public class RequestManagerImpl implements RequestManager{
 			throws NotFoundException, UnauthorizedException {
 		ValidateArgument.required(userInfo, "userInfo");
 		validateRequest(toUpdate);
-		validateEnvelopeCompletion(toUpdate);
 		validateFileHandleAccess(userInfo, toUpdate);
 
 		RequestInterface original = requestDao.getForUpdate(toUpdate.getId());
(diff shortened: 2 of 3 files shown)


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/dataaccess/RequestManagerImpl.java
```diff
@@ -143,6 +143,9 @@ public RequestInterface getRequestForUpdate(UserInfo userInfo, String accessRequ
 			throws NotFoundException {
 		ValidateArgument.required(userInfo, "userInfo");
 		ValidateArgument.required(accessRequirementId, "accessRequirementId");
+		// Verify the access requirement exists; a missing one must be a 404 rather than falling
+		// through to a blank new-request stub.
+		accessRequirementDao.get(accessRequirementId);
```
jay-hodgson, line 148: Makes sense.  Do other `request` paths (create, update, list) also check for AR existence?

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/dataaccess/RequestManagerImpl.java
```diff
@@ -226,6 +228,15 @@ RequestInterface update(UserInfo userInfo, RequestInterface toUpdate)
 				SubmissionState.SUBMITTED),
 				"A submission has been created. User needs to cancel the created submission or wait for an ACT member to review it before create another submission.");
 
+		// The eDUC signature envelope id is managed by the server (set when routing for signature
+		// and cleared when cancelling). Preserve the persisted value so a client editing the
+		// request cannot resurrect or change it from a stale copy. This must happen before
+		// validateEnvelopeCompletion so the envelope-completion check runs against the authoritative
+		// envelope id rather than whatever the client sent.
+		toUpdate.setEDucSignatureEnvelopeId(original.getEDucSignatureEnvelopeId());
```
jay-hodgson, line 236: Agreed, the client should not attempt to update this eDUC Signature Envelope ID (and if they do, ignore).  I'm still confused about what I'm seeing, which was a populated eDUC Signature Envelope ID field after the DELETE call.  Perhaps this is a client-side caching issue...

lib/lib-docusign/src/main/java/org/sagebionetworks/docusign/DocuSignEnvelopesApiImpl.java
```diff
@@ -77,7 +77,11 @@ public List<Envelope> listStatus(List<String> envelopeIds) {
 			EnvelopesApi envelopesApi = new EnvelopesApi(apiClient);
 			EnvelopeIdsRequest request = new EnvelopeIdsRequest();
 			request.setEnvelopeIds(envelopeIds);
-			EnvelopesInformation info = envelopesApi.listStatus(config.getAccountId(), request);
+			// The status endpoint only reads the envelope ids from the request body when the
+			// envelope_ids query parameter is set to "request_body"; otherwise DocuSign returns 400.
+			EnvelopesApi.ListStatusOptions options = envelopesApi.new ListStatusOptions();
+			options.setEnvelopeIds("request_body");
+			EnvelopesInformation info = envelopesApi.listStatus(config.getAccountId(), request, options);
```
jay-hodgson, line 84: No need to update a unit test?  Is there a unit test for this call?  And if not, should there be (this is not meant to be a leading question, I'm sincerely asking)?

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/dataaccess/RequestManagerImpl.java
brucehoff: TL;DR.  The PR is correct as is.  Details:
| Path | Keyed by AR id? | AR-existence checked? | How |
| --- | --- | --- | --- |
| create | yes (in body) | Yes | `accessRequirementDao.get(...)` + must be `ManagedACTAccessRequirement` |
| update | no (keyed by request id) | Transitively | Loads request by id (404s); AR id is immutable, so it can't point at a non-existent AR |
| list | no (no AR param) | N/A | Lists the user's own requests; no AR id to check |
| getRequestForUpdate | yes | Yes (this branch's fix) | New `accessRequirementDao.get(...)` — this was the gap, because the not-found path fell through to a blank stub instead of 404 |

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/dataaccess/RequestManagerImpl.java
brucehoff: The suspicion is that the client set the envelope ID after the deletion.  My suggestion is that we merge this PR to add the protection and then try again (e.g. on staging next week).

lib/lib-docusign/src/main/java/org/sagebionetworks/docusign/DocuSignEnvelopesApiImpl.java
brucehoff: - A plain unit test is not worth it and arguably not feasible — mocking EnvelopesApi requires mocking a constructor call, and even if we did, the test would  just assert "we set the option we set," which is tautological and doesn't verify DocuSign actually honors it. That's low value.

jay-hodgson (approved):

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/dataaccess/RequestManagerImpl.java
jay-hodgson: (updated Bruce's comment, converted table to markdown syntax so it would render properly in different viewport sizes)

PR #5795, opened 2026-08-23
PLFM-9903

brucehoff:
Add Access Requirement ID to AccessRequestSummary .


jay-hodgson (approved):
Thank you, this is what I need!  
Should accessRequirementId be marked as required (since it will always be set, right?)

PR #5797, opened 2026-08-24
PLFM-9900

brucehoff:
If a Docusign signer declines to sign, capture the reason they give and add to the signature status.

### DIFF
--- a/lib/lib-docusign/src/main/java/org/sagebionetworks/docusign/DocuSignClient.java
+++ b/lib/lib-docusign/src/main/java/org/sagebionetworks/docusign/DocuSignClient.java
@@ -166,6 +166,8 @@ public class DocuSignClient {
 				EDucSignerStatus signerStatus = new EDucSignerStatus();
 				signerStatus.setName(signer.getName());
 				signerStatus.setStatus(toEDucSignerStatusEnum(signer.getStatus()));
+				signerStatus.setDeclinedReason(signer.getDeclinedReason());
+				signerStatus.setDeclinedOn(parseDate(signer.getDeclinedDateTime()));
 				signerStatuses.add(signerStatus);
 				signerEmails.add(signer.getEmail());
 			}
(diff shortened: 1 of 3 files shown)


lib/lib-docusign/src/main/java/org/sagebionetworks/docusign/DocuSignClient.java
```diff
@@ -166,6 +166,8 @@ public EnvelopeStatusResult getEnvelopeStatus(String envelopeId) {
 				EDucSignerStatus signerStatus = new EDucSignerStatus();
 				signerStatus.setName(signer.getName());
 				signerStatus.setStatus(toEDucSignerStatusEnum(signer.getStatus()));
+				signerStatus.setDeclinedReason(signer.getDeclinedReason());
```
jay-hodgson, line 169: This declined reason is provided by a user, correct?  I'll need to remember to html sanitize it if we render it in the UI (and it could contain anything, in case there are safeguards for the backend to also consider)!

lib/lib-docusign/src/test/java/org/sagebionetworks/docusign/DocuSignClientTest.java
```diff
@@ -245,13 +252,23 @@ public void testGetEnvelopeStatusSuccess() {
...
 		assertNotNull(status.getModifiedOn());
-		assertEquals(2, status.getSignerStatus().size());
+		assertEquals(3, status.getSignerStatus().size());
 		assertEquals("Dr. Jones", status.getSignerStatus().get(0).getName());
 		assertEquals(EDucSignerStatusEnum.done, status.getSignerStatus().get(0).getStatus());
+		// A signer who has not declined has no reason or declined date.
+		assertNull(status.getSignerStatus().get(0).getDeclinedReason());
+		assertNull(status.getSignerStatus().get(0).getDeclinedOn());
 		assertEquals("Jane Admin", status.getSignerStatus().get(1).getName());
 		assertEquals(EDucSignerStatusEnum.pending, status.getSignerStatus().get(1).getStatus());
```
jay-hodgson, line 262: You could probably also add the same check for the pending signer, Jane Admin, that the declinedReason and declinedOn are also `null`.   But I see this as a trivial observation that can be ignored
```
		// A pending signer who has not declined has no reason or declined date.
		assertNull(status.getSignerStatus().get(1).getDeclinedReason());
		assertNull(status.getSignerStatus().get(1).getDeclinedOn());
```

jay-hodgson (approved):

PR #5800, opened 2026-08-27
if the issue is due to a clock skew or delayed credential refresh bot…

john-hill:
…h cases should be retried.


SandhraSokhal (approved):
LGTM

PR #5801, opened 2026-08-27
PLFM-9914

brucehoff:
In our eDUC/Docusign envelopes, when routing the Principal Investigator, we used the email address from the PI's Synapse account.  However, the data access request (DAR) creator is asked to enter the PI's "institutional address" into the DAR and the requirement is to route the envelope to that address rather than to the address from the PI's Synapse account.


SandhraSokhal (approved):
LGTM

PR #5802, opened 2026-08-27
PLFM-9663 Json Schema for achema based AR

SandhraSokhal:


lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/dataaccess/schema/GenerateDataAccessSchemaRequestInterface.json
```diff
@@ -0,0 +1,34 @@
+{
+	"title": "Generate Data Access Schema Request",
+	"type": "interface",
+	"description": "Request body to generate the JSON Schema and UI Schema for a form.",
+	"implements": [
+		{
+			"$ref": "org.sagebionetworks.repo.model.asynch.AsynchronousRequestBody"
+		}
+	],
```
nickgros, line 9: When we go to implement this, we may find that it is not actually that costly to build the response. If that is the case, then we should make this a synchronous API.

OK to defer that decision until the PR where this is implemented.

lib/lib-auto-generated/src/main/resources/schema/org/sagebionetworks/repo/model/dataaccess/schema/GeneratedFormStep.json
```diff
@@ -0,0 +1,16 @@
+{
+	"title": "Generated Form Step",
+	"description": "The schema information used to render a single step of a form. The jsonSchema validates the slice of submission data corresponding to that step's fields.",
```
nickgros, line 3: It might make more sense to have the UI split the FormTemplate + JSON Schema into individual steps, but I think for now it makes sense to include it on the backend.

nickgros (approved):
Looks great!

PR #5807, opened 2026-09-10
PLFM-9907

brucehoff:
Updates to the [service](https://rest-docs.synapse.org/rest/POST/dataAccessRequest/list.html) to list data access requests:
- add filter parameters: (1) access requirement id, (2) isEDuc;
- fix the missing signaturesAcquired and signaturesRequested fields

The fix involves calling a different Docusign API:  Instead of calling `listStatus` we now call `listStatusChanges` which includes the recipients for the routed envelopes along with the envelopes' statuses.  Once we have the recipients we can count them up and fill the signaturesAcquired and signaturesRequested fields.

This PR adds a manual test that I ran to verify that the new API returns the recipient information as expected.


SandhraSokhal (approved):
LGTM

PR #5812, opened 2026-09-14
PLFM-9945 Use schema types when importing CSV to grid

afwillia:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/grid/CsvSchemaReconciler.java
```diff
@@ -36,11 +54,80 @@ public static void reconcile(List<ColumnModel> csvSchema, JsonSchema validationS
...
+		Type type = property == null ? null : property.getType();
+		if (type == null) {
+			return;
+		}
+		switch (type) {
+			case string -> applyStringType(column, property.getMaxLength());
+			case integer -> applyIntegerType(column);
+			default -> {
+				// the inferred type is kept as-is
 			}
```
nickgros: Could we reuse `RecordSetSchemaResolver.toColumnModel`? I think these two methods are trying to do the same thing, but now they diverge, and this is missing cases (like `number`)

services/repository-managers/src/test/java/org/sagebionetworks/repo/manager/grid/CsvSchemaReconcilerTest.java
```diff
@@ -135,6 +136,199 @@ public void testReconcileWithMultipleColumns() {
...
+				new ColumnModel().setName("col1").setColumnType(ColumnType.MEDIUMTEXT)
+		);
+		JsonSchema validationSchema = new JsonSchema().setProperties(Collections.singletonMap(
+				"col1", new JsonSchema().setType(Type.array)
+		));
+		// call under test
+		CsvSchemaReconciler.reconcile(csvSchema, validationSchema);
+		assertEquals(new ColumnModel().setName("col1").setColumnType(ColumnType.MEDIUMTEXT), csvSchema.get(0));
+	}
+
```
nickgros: need a test for `{"type":"array","items":{"type":"string"}}` with no maxLength (I think this would fail today as it would be coerced to MEDIUMTEXT which is inappropriate for an array)

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/grid/CsvSchemaReconciler.java
```diff
@@ -2,26 +2,44 @@
...
+ * - A property of {@code "type": "array"} upgrades the column to the list
+ *   equivalent of its element type (e.g. STRING to STRING_LIST).
+ * - A property of {@code "type": "string"} replaces an inferred ENTITYID,
+ *   INTEGER or other non-text type with STRING or MEDIUMTEXT.
+ * - A property of {@code "type": "integer"} replaces an inferred
+ *   ENTITYID with INTEGER.
+ * A column that does not match a top-level schema property keeps its inferred
+ * type.
  */
 public class CsvSchemaReconciler {
```
nickgros, line 33: `CsvSchemaReconciler` and `RecordSetSchemaResolver` have different concerns and I think we need to separate them and make them clear.

`CsvSchemaReconciler` intends to determine how to parse a CSV to import it into the grid. It should not be concerned with limits specific to the `ColumnType` objects used by Synapse Table services (we should not have used `ColumnType` in the first place here IMO, but getting rid of it might be a larger change than we should make here).

`RecordSetSchemaResolver` intends to determine how to parse a CSV to make it queryable via Synapse Table services. The `ColumnType` limits are essential in this case, because if the CSV exceeds the limits, the table cannot be indexed & queried.

**Recommendation**: 
- Do not consider table/index limits inside `CsvSchemaReconciler`. It should not use any `ColumnConstants`/`ColumnTypeListMappings` limits. 
- Move all size/list-length capping into `RecordSetSchemaResolver`, applied as one uniform pass over the whole reconciled schema (CSV-matched columns and schema-only columns alike), reusing the capping/fallback rules `RecordSetSchemaResolver.toColumnModel` already encodes and already has test coverage for. Concretely:

 1. Revert `applyStringType`'s `MAX_ALLOWED_STRING_SIZE` cap (you can keep the `TEXT_COLUMN_TYPES` early-return). Let a capped maxLength pass through uncapped for the grid.
 2. In `RecordSetSchemaResolver.getReconciledSchema`, after `CsvSchemaReconciler.reconcile` + `addJsonSchemaOnlyColumns`, run every column through the existing size/list-limit rules (extract them from  `toColumnModel`'s string/array cases into a small reusable step) so a CSV-matched `STRING(5000)` and a schema-only `STRING(5000)` are clamped identically.

nickgros (changes requested):

nickgros:
Also I think would benefit from a regression test. Please add or extend an existing test in `GridEventBrokerWorkerIntegrationTest` to verify the fix. I'm still not 100% sure if the bug is in `CsvSchemaReconciler`, a regression test would prove it

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/table/RecordSetSchemaResolver.java
```diff
@@ -102,6 +110,9 @@ public ReconciledSchema getReconciledSchema(String entityId, FileHandle fileHand
 		Optional<JsonSchema> validationSchema = getBoundValidationSchema(entityId);
 		validationSchema.ifPresent(vs -> CsvSchemaReconciler.reconcile(schema, vs));
 		validationSchema.ifPresent(vs -> addJsonSchemaOnlyColumns(schema, vs));
+		// The grid reconciles types from the data and the schema alone, so every column is
+		// capped here to keep the schema within what the table index can store.
+		schema.forEach(RecordSetSchemaResolver::applyIndexLimits);
```
nickgros: Sorry, I was partially wrong in my earlier review. The class that handles adding the RecordSet's CSV data to the table index does _NOT_ use `getReconciledSchema`.

For that reason, you should remove the `applyIndexLimits` call in `getReconciledSchema` here (but still keep it in `toColumnModel`, that _is_ used for the table index).

A regression in this PR that would catches this is to have inferred entity ID data (`"syn123"`) against a string array schema with no max length (`{type: array, items: {type: string}}`) - the expected data should be a STRING_LIST `["syn123"]`, not be coerced to MEDIUMTEXT.

services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/grid/CsvSchemaReconciler.java
```diff
@@ -36,11 +57,76 @@ public static void reconcile(List<ColumnModel> csvSchema, JsonSchema validationS
...
+	/**
+	 * An integer property only replaces an inferred ENTITYID: the other inferred
+	 * types either already hold integers or hold values that are not integers at
+	 * all (e.g. a DATE inferred from date strings).
+	 */
+	private static void applyIntegerType(ColumnModel column) {
+		if (ColumnType.ENTITYID.equals(column.getColumnType())) {
+			column.setColumnType(ColumnType.INTEGER).setMaximumSize(null);
+		}
+	}
```
nickgros: Can you add an integration test for this case where it's inferred as an entity ID but should be set as integer?

I think this will fail downstream -- if the type was coerced to entity ID, then the values are like "syn123". The LongTranslator will not know how to handle these values, and will fail with an exception (instead of importing the wrong value, which is better since a user can fix manually or with the grid agent).

It doesn't look like this `ENTITY_ID` -> `integer` issue came up in the ticket (only `ENTITY_ID` -> `string`), so IMO, it's probably easiest to just drop this part

nickgros (changes requested):

PR #5814, opened 2026-09-16
PLFM-9942: Add check for assigneeId

xschildw:
In PLFM-9942, we see that an invalid assignee ID goes all the way to the DB and causes a FK exception (500). In this PR, we add a check on the assignee ID and return a 400 instead.


john-hill (approved):
LGTM!

PR #5816, opened 2026-09-16
PLFM-9665 create update new JsonSchemaAR

SandhraSokhal:


services/repository-managers/src/main/java/org/sagebionetworks/repo/manager/dataaccess/SubmissionManagerImpl.java
nickgros, line 189: Should validate the submitted data, which will be done in PLFM-9667

nickgros (approved):
Looks great! Some outstanding work, but we deferred it to future tickets.

PR #5818, opened 2026-09-18
PLFM-9941: Handle token error

xschildw:


integration-test/src/test/java/org/sagebionetworks/ITJsonSchemaControllerTest.java
```diff
@@ -144,6 +144,17 @@ public void testCreateOrganization() throws SynapseException {
 		assertEquals(synapse.getMyProfile().getOwnerId(), organization.getCreatedBy());
 	}
 
+	@Test
```
john-hill, line 147: This is well covered at the unit test level.

john-hill (approved):

integration-test/src/test/java/org/sagebionetworks/ITJsonSchemaControllerTest.java
xschildw: removed.

